PR #9, opened 2023-05-01
Testing lcm results against analytical solution

MImmesberger:
I implement a test against the analytical solution from [Iskhakov et al (2017)](https://onlinelibrary.wiley.com/doi/abs/10.3982/QE643). The analytical solution was created in the [lcm-dev](https://github.com/OpenSourceEconomics/lcm-dev) repo.

We also:

- Switch to csv testing data
- Add regression tests


MImmesberger:
I created datasets that contain the analytical solution. As discussed with @timmens I skipped the pre-commit hooks for file size as the files are larger than what is currently allowed (approx. 1.8Mb each file).

Also, note that the tests still fail for the initial retired state.

timmens:
Try to use

```pyton
    return wealth >= consumption > 0
```

instead of

```python
def next_wealth_constraint_no_borrowing(next_wealth, labor_income):
    return next_wealth >= labor_income
```

src/lcm/example_models.py
```diff
@@ -125,3 +129,29 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
     },
     "n_periods": 20,
 }
+
+PHELPS_DEATON_NO_BORROWING = {
+    "functions": {
+        "utility": phelps_deaton_utility_with_filter,
+        "next_wealth": next_wealth,
+        "next_wealth_constraint": next_wealth_constraint,
+        "consumption_constraint": consumption_constraint,
```
timmens, line 138: Are we sure we need both constraints here?

tests/test_analytical_solution.py
```diff
@@ -0,0 +1,75 @@
...
+def numerical_solution(params):
+    """Numerical solution."""
+    model = PHELPS_DEATON_NO_BORROWING
+    model["n_periods"] = params["num_periods"]
+    model["choices"]["consumption"]["start"] = 1
+    model["choices"]["consumption"]["stop"] = 100
+    model["choices"]["consumption"]["n_points"] = 10_000
+    model["states"]["wealth"]["start"] = 1
+    model["states"]["wealth"]["stop"] = 100
+    model["states"]["wealth"]["n_points"] = 10_000
```
timmens, line 24: Can you write these parameters in a dict and then update the model like:

```python
config_update = {...}
model = {**PHELPS_DEATON_NO_BORROWING, **config_update}
```

Can you use `n_periods` instead of `num_periods`?

tests/test_analytical_solution.py
```diff
@@ -0,0 +1,75 @@
...
+
+    solve_model, params_template = get_lcm_function(model=model)
+
+    params_template["beta"] = params["beta"]
+    params_template["next_wealth"]["wage"] = params["wage"]
+    params_template["next_wealth"]["interest_rate"] = params["r"]
+    params_template["utility"]["delta"] = params["delta"]
```
timmens, line 31: Same here, just update the template like

```python
params = {**params_template, **_params}
```

tests/test_create_params.py
```diff
@@ -6,14 +6,14 @@
...
-        "utility$delta",
-        "next_wealth$interest_rate",
-        "next_wealth$wage",
-        "wage_shock$sd",
-        "additive_utility_shock$scale",
+        "utility__delta",
+        "next_wealth__interest_rate",
+        "next_wealth__wage",
+        "wage_shock__sd",
+        "additive_utility_shock__scale",
```
timmens, line 16: Why did you change the sperator?

timmens (changes requested):
Thanks a lot! Janos and I will take a look at the failing tests. In the meantime you can address my comments.

src/lcm/example_models.py
MImmesberger: No, this is redundant.

tests/test_analytical_solution.py
MImmesberger: Done. I haven't found a way to really only update the values that I'm trying to change. The `retirement` dict is, for instance, part of the `choices` dict and I have to explicitly add it in the `config_update` dict, event though I leave it unchanged compared to the version in `PHELPS_DEATON_NO_BORROWING`.

tests/test_create_params.py
MImmesberger: That was @hmgaudecker's attempt to save his eyes from getting hurt: [268a692](https://github.com/OpenSourceEconomics/lcm/pull/9/commits/268a692041bcfefaab948390d62364151af745c4)

hmgaudecker: :man_shrugging:

tests/test_analytical_solution.py
```diff
@@ -0,0 +1,75 @@
...
+
+    numerical_solution = np.array(solve_model(params=params_template))
```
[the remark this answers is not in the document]
MImmesberger: [a620239](https://github.com/OpenSourceEconomics/lcm/pull/9/commits/a620239801b530ec67e2844c051612822554566f)

PR #13, opened 2023-05-14
Implement simulation function

timmens:
- [x] Implement data state choice space
- [x] Implement argmax and segment_argmax
- [x] Store optimal choices and continuation value in dict
- [x] Update state and iterate
- [x] Fix bugs

### DIFF
--- a/src/lcm/argmax.py
+++ b/src/lcm/argmax.py
@@ -19,9 +19,9 @@ def argmax(a, axis=None, initial=None, where=None):
             for which where is True are considered for the argmax.
 
     Returns:
-        jax.numpy.ndarray: Array with the same shape as a, except for the dimensions
-            specified in axis, which are dropped. The value corresponds to an index
-            that can be translated into a tuple of indices using jnp.unravel_index.
+        - jax.numpy.ndarray: Array with the same shape as a, except for the dimensions
+          specified in axis, which are dropped. The value corresponds to an index
+          that can be translated into a tuple of indices using jnp.unravel_index.
 
     """
     # Preparation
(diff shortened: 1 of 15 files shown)


src/lcm/argmax.py
```diff
@@ -0,0 +1,149 @@
...
+    max_value_mask = a == _max
+    if where is not None:
+        max_value_mask = max_value_mask & where
+    argmax = jnp.argmax(max_value_mask, axis=-1)
+
+    return argmax, _max.reshape(argmax.shape)
+
+
+def _move_axes_to_back(a, axes):
+    front_axes = tuple(set(range(a.ndim)) - set(axes))
```
janosg, line 57: Doesn't the set mess up the order (potentially). If so, the output array could end up transposed and the translation from generalized index to normal indices would get silently produce wrong results. 

It did not happen in a simple test but is definitely dangerous and set orders are not guaranteed. 

We should also add some tests for `_move_axes_to_back` and `_flatten_last_n_axes`

src/lcm/dispatchers.py
```diff
@@ -128,6 +128,25 @@ def _product_map(func, product_axes):
     return vmapped
 
 
+def vmap_1d(func, variables):
```
janosg, line 131: This needs a docstring

src/lcm/entry_point.py
```diff
@@ -117,9 +138,119 @@ def get_lcm_function(model, targets="solve", interpolation_options=None):
...
+        compute_ccv_argmax_functions=compute_ccv_argmax_functions,
+        model=_mod,
+        next_state=next_state,
+    )
+
+    _target = solve_model if targets == "solve" else simulate_model
+    return _target, _mod.params
+
+
+def create_compute_conditional_continuation_value(
```
janosg, line 160: If I see it correctly, this function is only tested via a test of the simulate function. I think we should test it in isolation!

src/lcm/entry_point.py
```diff
@@ -117,9 +138,119 @@ def get_lcm_function(model, targets="solve", interpolation_options=None):
...
+
+    @functools.wraps(u_and_f_mapped_over_cont_choices)
+    def compute_ccv(*args, **kwargs):
+        u, f = u_and_f_mapped_over_cont_choices(*args, **kwargs)
+        return u.max(where=f, initial=-jnp.inf)
+
+    return compute_ccv
+
+
+def create_compute_conditional_continuation_value_argmax(
```
janosg, line 196: If I see it correctly, this function is only tested via a test of the simulate function. I think we should test it in isolation!

src/lcm/entry_point.py
```diff
@@ -117,9 +138,119 @@ def get_lcm_function(model, targets="solve", interpolation_options=None):
...
+
+    return compute_ccv_argmax
+
+
+# ======================================================================================
+# Next state
+# ======================================================================================
+
+
+def get_next_state_function(model):
```
janosg, line 239: If I see it correctly, this function is only tested via a test of the simulate function. I think we should test it in isolation!

src/lcm/simulate.py
```diff
@@ -0,0 +1,457 @@
...
+from lcm.argmax import argmax, segment_argmax
+from lcm.dispatchers import spacemap, vmap_1d
+from lcm.interfaces import Space
+
+# ======================================================================================
+# Simulate
+# ======================================================================================
+
+
+def simulate(
```
janosg, line 16: We should write a test for simulate that does not use `get_lcm_function` so we can see if the problem is in the input generation or simulate itself.

janosg (commented):
I think this looks very good overall. The next step should be to write more tests for individual components and this will also help to find out why the main test for the simulation fails. I made specific comments about a few functions I would like to be tested. 

I also think we should start a few more naming conventions:

- I think we need some well documented abbreviations
    - scs: state choice space
    - ss: state space
    - ccv: conditional continuation values
- Instead of arg_emax we should say policy. E.g. instead of `compute_conditional_continuation_value_argmax` say `compute_continuous_policy`

src/lcm/simulate.py
timmens: In `test_simulate.py` I have created the tests:
- `test_simulate()`
- `test_simulate_debug()`
The second one calls `simulate` using `get_lcm_function` and the first one creates the inputs. The tests should test the same behavior (use the same model and parameters); however, only the first one passes. I will look into the details tomorrow.

tests/test_simulate.py
```diff
@@ -0,0 +1,230 @@
...
+    got = simulate(
+        params=params,
+        vf_arr_list=[None],
+        initial_states={"wealth": jnp.array([1.0, 50.5])},
+        **simulate_inputs,
+    )
+
+    choices = got[0]["choices"]
+
+    assert_array_equal(choices["retirement"], 1)
```
ChristianZimpelmann, line 83: Why exactly do we expect that the agent is not working in the last period? If I am not mistaken she could consume more than her current wealth which might be optimal in some cases.

tests/test_simulate.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # assert that everyone retires in the last period
+    assert_array_equal(res[-1]["choices"]["retirement"], 1)
+
+    # assert that higher wealth leads to higher consumption
+    for period in range(n_periods):
+        assert jnp.all(jnp.diff(res[period]["choices"]["consumption"]) >= 0)
+
+        # The following does not work. I.e. the continuation value in each period is not
+        # weakly increasing in wealth. It is unclear if this needs to hold.
```
ChristianZimpelmann, line 135: Seems plausible to me that it increases weekly in wealth.

ChristianZimpelmann (commented):
Looks nice! 

Some other ideas for test cases:

- if delta is higher for a model, consumption should be weakly lower and retirement should be weakly higher in all periods 
- if beta is higher for a model, utility in first period should be lower and utility in last period should be higher
- for beta = 1 and `interest_rate > 0`, utility should be increasing in periods (I think). For beta < 1 and `interest_rate = 0` should be the opposite.

tests/test_simulate.py
timmens: I rewrote the constraint, so this should be clearer. We have the borrowing constraint:

```python
consumption <= wealth
```

and the wealth transition

```python
next_wealth = (1 + interest_rate) (wealth - consumption) + wage * working
```

If you work in period t, your wealth in period t+1 increases by `wage`. But in your last period, you do not care about the wealth in your next period. One could discuss whether this timing assumption is sensible, but it seems like this is how it is formulated by Iskhakov et al. (2017).

ChristianZimpelmann: Ah, then it makes sense.

I thought we were using this borrowing constraint:

```

def next_wealth_constraint(next_wealth):
    return next_wealth >= 0
    
```

I don't have a good intution on whether we would like to use the `next_wealth_constraint` or `consumption_constraint`. We can discuss it later.

timmens:
Saving this code snippet here because it will be deleted from git when we squash-merge this PR.

```python
def _create_segment_nd_arange(segment_ids, num_segments, shape):
    """Create an nd-arange for each segment.

    Args:
        segment_ids (jax.numpy.ndarray): 1d array with segment identifiers. See
            jax.ops.segment_max.
        num_segments (int): Total number of segments. See jax.ops.segment_max.
        shape (tuple): The shape to which the index map should be broadcasted. The first
            dimension must coincide with len(segment_ids).

    Returns:
        jax.numpy.ndarray: An array of shape 'shape' containing an arange in the first
            dimension for each segment, and repeating this value along the remaining
            shape[1:] dimensions.

    """
    bincount = jnp.bincount(segment_ids, length=num_segments)
    cumsum = jnp.cumsum(bincount)

    # shift the cumulative sum to the right and pad with zero at the beginning
    shifted_cumsum = jnp.pad(cumsum[:-1], (1, 0), constant_values=0)

    # create an array of indices for each segment
    segment_arange = jnp.arange(shape[0]) - shifted_cumsum[segment_ids]

    # reshape the array of indices to be broadcastable to the desired shape
    reshaped = segment_arange.reshape(-1, *((1,) * (len(shape) - 1)))
    return jnp.broadcast_to(reshaped, shape)
```

PR #20, opened 2023-07-13
Some updates to docstrings

hmgaudecker:


hmgaudecker:
@timmens, just a couple of small things that might be good to merge before they grow large...
- I enabeled `nbstripout` again in the hooks and updated all of them in the process.
- Hence, you can ignore all changes to notebooks
- In `function_evaluator`, just a renaming of a loop variable, which had confused me before
- Do check the docstring changes of `simulate` thoroughly to make sure that the docstring matches (future) behaviour. Janoś and I thought that we wrote what it *should* do, if you agree and that is not current behaviour, please open an issue.

timmens (approved):
Looks good, approved; thank you!

The behavior you describe in the result of `simulate` is currently not implemented. I opened issue #25 accordingly.

PR #21, opened 2023-07-13
Fix bugs in space info creation.

timmens:
In creating lcm internal functions in `entry_point.py`, the space info was used for the wrong periods; in particular, in period t, one needs to use the space info of period t+1. We further needed to exclude certain auxiliary variables in the last period.

### DIFF
--- a/src/lcm/__init__.py
+++ b/src/lcm/__init__.py
@@ -1,3 +1,8 @@
 import jax
 
+from lcm import mark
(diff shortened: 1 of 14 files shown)


timmens:
We will increase the test coverage of the new stochastic components in a separate PR.

hmgaudecker (approved):
Excelente!

PR #22, opened 2023-07-16
Implement interface for stochastic transitions of discrete variables 

timmens:
Questions:
- [ ] How do we process stochastic next functions such that they work well with dags


timmens:
Is handled in #34

PR #23, opened 2023-07-17
Call solve in simulate function

timmens:
In this PR we add the target `"solve_and_simulate"` which partials the solve function into the simulate function.

### DIFF
--- a/src/lcm/entry_point.py
+++ b/src/lcm/entry_point.py
@@ -36,7 +36,7 @@ def get_lcm_function(model, targets="solve", interpolation_options=None):
     Args:
         model (dict): User model specification.
         targets (str or iterable): The requested function types. Currently only
-            "solve" is supported.
+            "solve", "simulate" and "solve_and_simulate" are supported.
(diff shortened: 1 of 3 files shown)


src/lcm/entry_point.py
```diff
@@ -48,7 +48,7 @@ def get_lcm_function(model, targets="solve", interpolation_options=None):
     # ==================================================================================
     # preparations
     # ==================================================================================
-    if targets not in {"solve", "simulate"}:
+    if targets not in {"solve", "simulate", "solve_and_simulate"}:
```
hmgaudecker, line 51: Did not look thoroughly, but is this being tested?

timmens: I've added an additional test to check that `solve_and_simulate` produces the same result as first calling `solve` and then calling `simulate`.

hmgaudecker:
Great thanks—pls go ahead & merge!

Tim Mensinger ***@***.***> schrieb am Di., 18. Juli 2023,
18:27:

PR #24, opened 2023-07-19
Split dags call in creation of utility and feasibility

timmens:


src/lcm/model_functions.py
```diff
@@ -53,20 +53,30 @@ def _big_u(utility, continuation_value, params):
             return utility + params["beta"] * continuation_value
 
     feasibility = get_combined_constraint(model)
+    next_state = get_next_state_function(model)
 
-    func_dict = {"__big_u__": _big_u, "feasibility": feasibility, **model.functions}
+    func_dict = {
+        "__big_u__": _big_u,
```
hmgaudecker, line 59:
```suggestion
        "_big_u": _big_u,
```

Or is there a reason to have a key different from the identifier? Would confuse the hell out of me. In general, variables should be different phonetically so one can easily talk about the code.

timmens:
Is handled in #34

PR #33, opened 2023-09-14
Implement allow_args and improve allow_kwargs decorator

timmens:
In this PR we implement the `allow_args` decorator and improve the `allow_kwargs` decorator.

### `allow_kwargs`

We allow for the case where the function to be wrapped has a keyword-only parameter.

### `allow_args`

We implement a decorator in line with `allow_kwargs`, which lets us call a function (that potentially has some keyword-only parameters) with positional args only. Importantly, the resulting function accepts both args and kwargs.

### DIFF
--- a/src/lcm/dispatchers.py
+++ b/src/lcm/dispatchers.py
@@ -193,22 +193,21 @@ def allow_kwargs(func):
     def allow_kwargs_wrapper(*args, **kwargs):
         parameters = inspect.signature(func).parameters
 
-        # Separate keyword-only arguments
+        # Separate keyword-only arguments and remove from kwargs
         keyword_only = {
             k: v
             for k, v in kwargs.items()
             if k in parameters and parameters[k].kind == inspect.Parameter.KEYWORD_ONLY
         }
-        for kwarg_name in keyword_only:
-            kwargs.pop(kwarg_name)
-
-        positional = list(args) if args is not None else []
+        for k in keyword_only:
+            kwargs.pop(k)
 
         # Check if the total number of arguments matches the function signature
-        kwargs = {} if args is None else kwargs
         if len(args) + len(kwargs) + len(keyword_only) != len(parameters):
             raise ValueError("Not enough or too many arguments provided.")
 
+        # Separate positional arguments and convert keyword arguments to positional
+        positional = list(args)
         positional += convert_kwargs_to_args(kwargs, list(parameters))
 
         return func(*positional, **keyword_only)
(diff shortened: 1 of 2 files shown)


src/lcm/dispatchers.py
```diff
@@ -191,21 +191,81 @@ def allow_kwargs(func):
...
-        kwargs = {} if args is None else kwargs
-        if len(args) + len(kwargs) != len(parameters):
+        parameters = inspect.signature(func).parameters
+
+        # Retrieve keyword-only arguments and remove from kwargs
+        keyword_only = {
+            k: v
+            for k, v in kwargs.items()
+            if k in parameters and parameters[k].kind == inspect.Parameter.KEYWORD_ONLY
+        }
```
janosg, line 201: It wasn't done before but I think we should move everything we can out of the wrapper and into the decorator. I.e. reading the signature and creating a list of keyword only arguments. Just to make sure we don't get a runtime cost.

src/lcm/dispatchers.py
```diff
@@ -191,21 +191,81 @@ def allow_kwargs(func):
...
+        for k in keyword_only:
```
janosg, line 202: I would prefer a dict comprehension over pop

src/lcm/dispatchers.py
```diff
@@ -191,21 +191,81 @@ def allow_kwargs(func):
...
-
-        kwargs = {} if args is None else kwargs
-        if len(args) + len(kwargs) != len(parameters):
+        parameters = inspect.signature(func).parameters
+
+        # Retrieve keyword-only arguments and remove from kwargs
+        keyword_only = {
+            k: v
+            for k, v in kwargs.items()
+            if k in parameters and parameters[k].kind == inspect.Parameter.KEYWORD_ONLY
```
janosg, line 200: Instead of converting all but the KEYWORD_ONLY arguments into positional ones, we could just convert the POSITIONAL_ONLY ones

janosg: After discussion we put less focus on style and just try to work around all problem introduced by vmap. Example: 

```python
from jax import vmap
import jax.numpy as jnp
import inspect

def f(*, a, b):
    return a + b


v = vmap(f, in_axes=(0, 0))

parameters = inspect.signature(v).parameters

parameters["a"].kind
```
```txt
<_ParameterKind.KEYWORD_ONLY: 3>
```
I.e., despite being a positional only parameter (due to jax.vmap), the signature still says that we have a keyword only argument here.

janosg (approved):

PR #34, opened 2023-09-20
Refactor model functions and implement basic stochastic transitions

timmens:
In this PR, we refactor the model functions module to have more control over the execution order. This is necessary to implement stochastic transitions, which we implement for the basic fully discrete case that only depends on state variables.


src/lcm/model_functions.py
```diff
@@ -10,28 +14,138 @@ def get_utility_and_feasibility_function(
...
+    # ==================================================================================
+    # Generate dynamic functions
+    # ==================================================================================
+    current_u_and_f = get_current_u_and_f(model)
+
+    if not is_last_period:
+        next_states = get_next_states_function(model)
+        next_weights = get_next_weights_function(model)
+
+        scalar_function_evaluator = get_function_evaluator(
```
janosg, line 34: Rename to `scalar_value_function` = get_function_evaluator

src/lcm/model_functions.py
```diff
@@ -10,28 +14,138 @@ def get_utility_and_feasibility_function(
...
+
+    # ==================================================================================
+    # Update this section
+
+    arg_names = set()
+    for func in relevant_functions:
+        parameters = inspect.signature(func).parameters
+        arg_names.update(parameters.keys())
+    arg_names = list({"vf_arr", *arg_names})
+    arg_names = [arg for arg in arg_names if "next_" not in arg]
```
janosg, line 61: This should read something like:

`arg_names = get_union_of_arguments(relevant_functions). 

`get_union_of_arguments` might be useful somewhere else, so it should probably be a public function in `lcm.signature.py` or something similar.

src/lcm/model_functions.py
```diff
@@ -10,28 +14,138 @@ def get_utility_and_feasibility_function(
...
+        states = {k: v for k, v in kwargs.items() if k in state_variables}
+        choices = {k: v for k, v in kwargs.items() if k in choice_variables}
+
+        u, f = current_u_and_f(**states, **choices, params=kwargs["params"])
+
+        if is_last_period:
+            big_u = u
+
+        else:
+            _next_states = next_states(**states, **choices, params=kwargs["params"])
```
janosg, line 83: I think we should pull out the if conditions from the generated `u_and_f` function and have two def statements instead.

src/lcm/model_functions.py
```diff
@@ -10,28 +14,138 @@ def get_utility_and_feasibility_function(
...
+            if "state_indexer" in kwargs:
+                ccvs_at_nodes = function_evaluator(
+                    **_next_states,
+                    vf_arr=kwargs["vf_arr"],
+                    state_indexer=kwargs["state_indexer"],
+                )
+            else:
+                ccvs_at_nodes = function_evaluator(
+                    **_next_states,
+                    vf_arr=kwargs["vf_arr"],
```
janosg, line 103: `**{k: v for k, v in kwargs if k in function_evaluator_arguments}`

And `function_evaluator_arguments` needs to be determined outside of u_and_f

That's basically what dags does internally.

src/lcm/model_functions.py
```diff
@@ -10,28 +14,138 @@ def get_utility_and_feasibility_function(
...
 ):
     """Create a func to calculate utility and feasibility of a state choice combination.
 
-    Utility here is not just the current period reward but also includes the discounted
-    expected future reward. In the dynamic programming literature this is often denoted
-    as capital U, whereas the period reward is lowercase u.
+    Utility here is not just the current period reward (or felicity function) but also
+    includes the discounted expected future reward. In the dynamic programming
+    literature this is often denoted as capital U or V, whereas the felicity function is
+    lowercase u.
```
janosg, line 133: Delete!

src/lcm/model_functions.py
```diff
@@ -74,6 +188,26 @@ def _big_u(utility, continuation_value, params):
...
+    Args:
+        stochastic_variables (list): TODO
+
+    Returns:
+        callable
+
+    """
+    arg_names = [f"weight_next_{var}" for var in stochastic_variables]
+
+    @allow_kwargs
```
janosg, line 203: Need a comment that `allow_kwargs` is here just a lazy way of not consolidating args and kwargs inside `_outer`

janosg: The clean way would to do this would be to write a `all_as_args` as analogue to the `all_as_kwargs` function. 

We need `lcm.functools`!

janosg (commented):
Very nice and a huge step forward.

src/lcm/process_model.py
```diff
@@ -288,14 +336,99 @@ def _get_function_with_dummy_params(func):
...
+        grids (dict): Dictionary containing all variables of the model. The keys are
+            the names of the variables. The values are the grids.
+
+    Returns:
+        dict: Dictionary that maps variable names to label translators.
+
+    """
+    return {
+        var: _get_label_translator(labels=grids[var], in_name=var) for var in variables
+    }
```
timmens, line 434: @janosg Can you check whether this is the intended usage of the label translator defined in `function_evaluator.py`?

janosg: In principle yes, but I think in `weight_func` I would use `all_as_args` instead of `all_as_kwargs` and then simplify the rest a bit. 

Also: took me quite a while to see the plural-s on `_get_label_translators`. We need a more visible distinction. 

And probably make `_get_label_translator` public then?

PR #36, opened 2023-10-02
Improve simulation output

timmens:
In this PR, we change the simulation output from a list of dicts containing the simulated choices (on dict for each period) , to a data frame with multiindex (periods, initial_states_id) and columns corresponding to choices, states and targets (auxiliary variables that might be of interest).

---

After this simulation call

```python
    res = simulate_model(
        params,
        vf_arr_list=vf_arr_list,
        initial_states={
            "wealth": jnp.array([1.0, 20, 40, 70]),
        },
        targets=["utility"],
    )
```

the output will look like this:

Screenshot 2023-10-04 at 10 23 31


src/lcm/simulate.py
```diff
@@ -19,6 +21,7 @@ def simulate(
     initial_states,
     solve_model=None,
     vf_arr_list=None,
+    targets=None,
```
ChristianZimpelmann, line 24:
```suggestion
    additional_targets=None,
```
?

PR #49, opened 2023-11-01
Allow for models with only discrete states

timmens:


timmens:
Has been solved in #47

PR #53, opened 2024-02-06
Allow for dependence on discrete choices in stochastic transitions

timmens:


src/lcm/create_params.py
```diff
@@ -102,12 +102,12 @@ def _create_shock_params(model, variable_info, grids):
...
         )
 
-        _check_variables_are_all_discrete_states(
+        _check_variables_are_all_discrete_or_period(
             variables=dependencies,
             variable_info=variable_info,
             msg_suffix=(
-                f"The function next_{var} can only depend on discrete state variables "
-                f"or '_period'."
+                f"The function next_{var} can only depend on discrete variables or"
```
hmgaudecker, line 109: Just a clarification because there seems to be a discrepancy to me between this line and the PR title: Does it also include derived variables? 

Say state and choice jointly determine whether a child is born during the period -- but that is neither a (beginning-of-period) state nor a choice. Would that be allowed for?

timmens: Currently, this is not allowed. Let me check whether this is a straightforward extension.

timmens: No, it would not be straightforward. We would have to implement a lot more changes. For example, internally, we only keep track of "pure" model variables. We don't have information on derived variables. There are also some other problems: since we are using the index of the variables to access the correct entry in the transition matrix, the transition matrix for derived variables could look very weird. Think `health` in `{0, 1, 2}`, `partner` in `{0, 1, 2}` and `well-being = health * partner`. Then `well-being` in `{0, 1, 2, 4}`.

ChristianZimpelmann: > However, one could simulate this behavior by making the state transition dependent on both variables.

That's how we currently do it in our project: Whether a child is born depends on the derived variable `n_children` which is derived from the state variables `n_children_lagged` and `age_youngest_child` (indicating whether a child is born this period).

However, this leads to unnecessary large and difficult-to-handle transition matrices.

hmgaudecker: Cool. We should discuss in detail at some point whether this will be worth it or not!

src/lcm/example_models_stochastic.py
```diff
@@ -9,7 +9,7 @@
 
 
 def utility(consumption, working, health, partner, delta, gamma):  # noqa: ARG001
-    return jnp.log(consumption) + (gamma * health - delta) * working
+    return jnp.log(consumption + 1) + (gamma * health - delta) * working
```
hmgaudecker, line 12: Why is this change necessary? The constraint `consumption > 0` should be handled elsewhere, if possible. Not clear to me why `consumption > -0.99999` is the easier thing to handle?

src/lcm/example_models_stochastic.py
```diff
@@ -62,13 +62,18 @@ def consumption_constraint(consumption, wealth):
...
+    "utility": {"delta": 0.5, "gamma": 0.25},
+    "next_wealth": {"interest_rate": 0.05, "wage": 10.0},
     "next_health": {},
     "consumption_constraint": {},
     "shocks": {
         "health": jnp.array([[[0.5, 0.5], [0.5, 0.5]], [[0.5, 0.5], [0.5, 0.5]]]),
-        "partner": jnp.array([[1.0, 0], [0.0, 1]]),
+        "partner": jnp.array(
+            [
+                [[[0, 1.0], [1.0, 0]], [[0, 1.0], [0.0, 1.0]]],
```
hmgaudecker, line 74: Do we rely on these somewhere? If possible, numbers like `[[0.9, 0.1], [0.05, 0.95]]` or so would be even better (=more instructive for people using this as documentation). Same for health; `beta` could be 0.95 ?

hmgaudecker (approved):
Excellent! Just some tiny stuff.

src/lcm/example_models_stochastic.py
timmens: Very good point. We cannot change the degenerate distribution for the `partner` shock because the tests use this, however, I added more comments and made the distribution of `health` more realistic.

src/lcm/example_models_stochastic.py
timmens: Good point. Currently, we only have a proper constraint for `consumption < wealth`. I changed the grids of the model for `consumption` and `wealth` to both start at 1 (in principle, any positive number would work).

src/lcm/create_params.py
timmens: I am adding it to my LCM notes.

PR #56, opened 2024-02-28
Refactor example models

### LINKED ISSUE #46: ENH: Clean up testing and example models
There are two modules `example_models.py` and `example_models_stochastic.py` containing multiple models. Then there is also the `get_models.py` module, defining some other models. This is cluttered and leads to bugs.

I propose we create a `_models` folder in which there are many more modules to reduce clutter. On top we should have a `models.py` module that imports all finished models from `_models` and puts them in `__all__`.

timmens:
In this PR, I refactor the example models.

Previously, the example models were located in `src/example_models` but were not used in any of the code but only in the tests. One example model did not have a meaningful test but existed only for experimenting on the server.

To change this, I created two new folders:

1. `tests/test_models` contains all the models used in the tests.
2. `examples` contain example models that can be used elsewhere or as a reference but do not have a corresponding test.

I also added comments and cleaned up the example models in the process. Moreover, while deleting the module `get_model.py`, I updated the regression tests.

Closes #46


examples/long_running.py
```diff
@@ -0,0 +1,121 @@
...
+
+def age(_period):
+    return _period + 18
+
+
+# --------------------------------------------------------------------------------------
+# State transitions
+# --------------------------------------------------------------------------------------
+def next_wealth(wealth, consumption, working, wage, interest_rate):
+    return (1 + interest_rate) * (wealth - consumption) + wage * working
```
hmgaudecker, line 46:
```suggestion
    return (1 + interest_rate) * (wealth + working * wage - consumption)
```
This would be the usual way to write it down. Also for the constraint below -- I should be able to consume my labor income without having to pay interest on it.

Turn around working / wage: 0/1-indicator upfront eases readability.

examples/long_running.py
```diff
@@ -0,0 +1,121 @@
...
+    wage_shock,
+    interest_rate,
+):
+    return interest_rate * (wealth - consumption) + wage * wage_shock * working
+
+
+# --------------------------------------------------------------------------------------
+# Constraints
+# --------------------------------------------------------------------------------------
+def consumption_constraint(consumption, wealth):
```
hmgaudecker, line 67:
```suggestion
def consumption_constraint(consumption, wealth, labor_income):
```
Else states with 0 wealth cannot support positive consumption. Probably useful to define a function labor_income, in that case stochastics are put into that function as well, no need to carry around shocks.

examples/long_running.py
```diff
@@ -0,0 +1,121 @@
...
+    return consumption <= wealth
```
hmgaudecker, line 68:
```suggestion
    return consumption <= wealth + labor_income
```

examples/README.md
hmgaudecker, line 1: Might be useful to import the test models here, to the extent they are economically useful and don't just serve to illustrate something. But really only a question once we do not only have user-developers.

tests/test_models/phelps_deaton.py
```diff
@@ -1,65 +1,71 @@
...
+
+# --------------------------------------------------------------------------------------
+# Utility functions
+# --------------------------------------------------------------------------------------
+def utility(consumption, working, delta):
     return jnp.log(consumption) - delta * working
 
 
-def phelps_deaton_utility_with_filter(
+def utility_with_filter(
```
hmgaudecker, line 28: We need this function because ATM, we have to put all states into the utility function, right? Maybe add a to-do to schedule it for removal, which is associated with the respective issue.

tests/test_models/phelps_deaton.py
```diff
@@ -1,65 +1,71 @@
...
+
+def age(_period):
+    return _period + 18
 
 
+# --------------------------------------------------------------------------------------
+# State transitions
+# --------------------------------------------------------------------------------------
 def next_wealth(wealth, consumption, working, wage, interest_rate):
     return (1 + interest_rate) * (wealth - consumption) + wage * working
```
hmgaudecker, line 56: See above, this is really weird. Funny that I notice only now. Can we add a reference where we got the model from? Was that the DC-EGM paper?

tests/test_models/phelps_deaton.py
```diff
@@ -1,65 +1,71 @@
...
 
-RETIREMENT_AGE = 65
-N_CHOICE_GRID_POINTS = 500
-N_STATE_GRID_POINTS = 100
+# ======================================================================================
+# Numerical parameters and constants
+# ======================================================================================
 
+N_GRID_POINTS = {
+    "states": 100,
```
hmgaudecker, line 10:
```suggestion
    "wealth": 100,
```
I think it should be more useful to be explicit rather than general here.

tests/test_models/phelps_deaton.py
```diff
@@ -1,65 +1,71 @@
...
+    "choices": 500,
```
hmgaudecker, line 11:
```suggestion
    "consumption": 500,
```

tests/test_models/phelps_deaton.py
```diff
@@ -1,65 +1,71 @@
...
-def wage(age):
-    return 1 + 0.1 * age
-
-
-def age(_period):
-    return _period + 18
-
-
+# --------------------------------------------------------------------------------------
+# Filters
```
hmgaudecker, line 67: Now I finally understand what is confusing me here in terms of terminology.

Typically, when we say "filter", we mean that we *filter out* the big particles that don't pass it.

But we define the states that should *pass through* the filter. 

Not sure there is simple terminology to turn that around, but I would find it helpful.

tests/test_models/stochastic.py
```diff
@@ -1,21 +1,39 @@
...
+# --------------------------------------------------------------------------------------
 def utility(consumption, working, health, partner, delta, gamma):  # noqa: ARG001
     return jnp.log(consumption) + (gamma * health - delta) * working
 
 
+# --------------------------------------------------------------------------------------
+# Deterministic state transitions
+# --------------------------------------------------------------------------------------
 def next_wealth(wealth, consumption, working, wage, interest_rate):
     return (1 + interest_rate) * (wealth - consumption) + wage * working
```
hmgaudecker, line 31: See above.

tests/test_models/stochastic.py
```diff
@@ -1,21 +1,39 @@
...
 
+# ======================================================================================
+# Model functions
+# ======================================================================================
 
+
+# --------------------------------------------------------------------------------------
+# Utility function
+# --------------------------------------------------------------------------------------
 def utility(consumption, working, health, partner, delta, gamma):  # noqa: ARG001
```
hmgaudecker, line 23: Would be neat to know what delta and gamma are supposed to do.

tests/test_models/stochastic.py
```diff
@@ -26,11 +44,18 @@ def next_partner(_period, working, partner):  # noqa: ARG001
     pass
 
 
+# --------------------------------------------------------------------------------------
+# Constraints
+# --------------------------------------------------------------------------------------
 def consumption_constraint(consumption, wealth):
```
hmgaudecker, line 50: See above

tests/test_models/stochastic.py
```diff
@@ -1,21 +1,39 @@
...
 import lcm
 
-N_CHOICE_GRID_POINTS = 500
-N_STATE_GRID_POINTS = 100
+# ======================================================================================
+# Numerical parameters and constants
+# ======================================================================================
+
+N_GRID_POINTS = {
+    "states": 100,
```
hmgaudecker, line 11: See above.

examples/long_running.py
```diff
@@ -0,0 +1,121 @@
+"""Example specification for a consumption-savings model with health and leisure."""
+
+import jax.numpy as jnp
+
+# ======================================================================================
+# Numerical parameters and constants
+# ======================================================================================
+N_GRID_POINTS = {
+    "states": 100,
```
hmgaudecker, line 9: See my comment in PHELPS_DEATON below.

hmgaudecker (commented):
Much clearer, thanks a lot!

I have a bunch of comments, most of which are unrelated to the changes made here, though :innocent:

tests/test_models/phelps_deaton.py
timmens: image

It is from the DC-EGM model. I will add a reference.

tests/test_models/phelps_deaton.py
timmens: Yes, exactly! Very good point, will do that!

examples/long_running.py
timmens: See below. This actually follows the DC-EGM paper.

examples/long_running.py
timmens: See below. This actually follows the DC-EGM paper.

tests/test_models/phelps_deaton.py
timmens: Currently we have:

```python
def mandatory_retirement_filter(retirement, age):
    return jnp.logical_or(retirement == 1, age < RETIREMENT_AGE)
```

Would you prefer for users to define the negation:

```python
def mandatory_retirement_filter(retirement, age):
    return jnp.logical_and(retirement == 0, age >= RETIREMENT_AGE)
```

hmgaudecker: No, we agreed with @janosg that ultimately, this way 'round is much clearer.

It is just a matter of terminology. And maybe even just a change that I need to make in my brain.

timmens: I guess a better term for what we are doing right now would be "mask".

tests/test_models/stochastic.py
timmens: Good point. Renamed it to "disutility_of_work".

tests/test_models/phelps_deaton.py
hmgaudecker: Not sure, masking to me implies that other states are generated and hidden. 

But again, might just be a switch in my head towards filtering=keeping the parts in question.

examples/long_running.py
hmgaudecker: The we should call it dc_egm example and use a sensible Phelps-Deaton for our own examples!

This basically means your starve to death despite working for a good wage if you have not saved anything...

examples/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Running an example
+
+Say you want to solve the [`long_running`](./long_running.py) example locally. In a
+Python shell, execute:
+
+```python
+from lcm.entry_point import get_lcm_function
+
+from long_running import MODEL_CONFIG, PARAMS
```
hmgaudecker, line 17: Why would this work? Don't I need a path unless my shell is located in that very folder?

tests/test_analytical_solution.py
```diff
@@ -2,45 +2,72 @@
...
-from lcm.get_model import get_model
 from numpy.testing import assert_array_almost_equal as aaae
 
+from tests.test_models.deterministic import BASE_MODEL_WITH_FILTERS
+
+# ======================================================================================
+# Model specifications
+# ======================================================================================
+
+ISKHAVOV_2017_PARAMS = {
```
hmgaudecker, line 15:
```suggestion
ISKHAKOV_2017_PARAMS = {
```

examples/long_running.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+# --------------------------------------------------------------------------------------
+# Auxiliary variables
+# --------------------------------------------------------------------------------------
+def labor_income(wage, working):
+    return wage * working
+
+
+def working(leisure):
```
hmgaudecker, line 36:
```suggestion
def leisure(working):
```
`working` just needs to be the choice to fit my brain.

examples/long_running.py
```diff
@@ -0,0 +1,117 @@
...
+        "next_wealth": next_wealth,
+        "next_health": next_health,
+        "consumption_constraint": consumption_constraint,
+        "labor_income": labor_income,
+        "working": working,
+        "wage": wage,
+        "age": age,
+    },
+    "choices": {
+        "leisure": {"options": [0, 1]},
```
hmgaudecker, line 82:
```suggestion
        "working": {"options": [0, 1]},
```

examples/long_running.py
```diff
@@ -0,0 +1,117 @@
...
+# ======================================================================================
+
+MODEL_CONFIG = {
+    "functions": {
+        "utility": utility,
+        "next_wealth": next_wealth,
+        "next_health": next_health,
+        "consumption_constraint": consumption_constraint,
+        "labor_income": labor_income,
+        "working": working,
```
hmgaudecker, line 77:
```suggestion
        "leisure": leisure,
```

examples/long_running.py
```diff
@@ -0,0 +1,117 @@
...
+
+# --------------------------------------------------------------------------------------
+# Auxiliary variables
+# --------------------------------------------------------------------------------------
+def labor_income(wage, working):
+    return wage * working
+
+
+def working(leisure):
+    return 1 - leisure
```
hmgaudecker, line 37:
```suggestion
    return 1 - working
```

tests/test_entry_point.py
```diff
@@ -5,21 +5,22 @@
...
 from pybaum import tree_equal, tree_map
 
+from tests.test_models.deterministic import (
+    BASE_MODEL,
+    BASE_MODEL_FULLY_DISCRETE,
+    BASE_MODEL_WITH_FILTERS,
+    utility,
+)
+
 MODELS = {
```
hmgaudecker, line 20: Unless we'll use this more than once, I find this more confusing than helpful.

tests/test_models/deterministic.py
```diff
@@ -0,0 +1,171 @@
...
+# --------------------------------------------------------------------------------------
+def utility(consumption, working, disutility_of_work):
+    return jnp.log(consumption) - disutility_of_work * working
+
+
+def utility_with_filter(
+    consumption,
+    working,
+    disutility_of_work,
+    # Temporary workaround for bug described in issue #30, which requires us to pass
```
hmgaudecker, line 39: We should enable the todo stuff from ruff, which enforces a format like:

    # TODO(@hmgaudecker): Remove explicit cast when vectorisation is enabled.
    # https://github.com/iza-institute-of-labor-economics/gettsim/issues/515

I find that quite useful, at least.

tests/test_models/deterministic.py
```diff
@@ -0,0 +1,171 @@
...
+
+def age(_period):
+    return _period + 18
+
+
+# --------------------------------------------------------------------------------------
+# State transitions
+# --------------------------------------------------------------------------------------
+def next_wealth(wealth, consumption, working, wage, interest_rate):
+    return (1 + interest_rate) * (wealth - consumption) + wage * working
```
hmgaudecker, line 65: Can you do the `labor_income` thing here, too?

tests/test_models/stochastic.py
```diff
@@ -1,21 +1,56 @@
...
+    partner,  # noqa: ARG001, TODO: Remove unused arguments once #30 is fixed.
+    disutility_of_work,
+):
+    return jnp.log(consumption) - (1 - health / 2) * disutility_of_work * working
+
+
+# --------------------------------------------------------------------------------------
+# Deterministic state transitions
+# --------------------------------------------------------------------------------------
 def next_wealth(wealth, consumption, working, wage, interest_rate):
```
hmgaudecker, line 47: Can you do the `labor_income` thing here, too?

hmgaudecker (approved):
Just small stuff left, thanks a lot! 

If other models enter the tests, we may want to rename `deterministic` to `base_deterministic` etc., but no need to do that preemptively.

examples/README.md
timmens: Correct, it this will only run when you are in the examples folder. This code snippet aims to show from which module to take the model config and params, and what lcm functions to call. This is not supposed to be directly "runnable." Currently, I do not think that I want these examples inside src (which would fix the path issue), because I want them to be standalone. Any suggestions? I could add a warning sign noting that you need to be in the examples folder for this to work?

hmgaudecker: Yeah, should just be part of the instructions.

tests/test_entry_point.py
timmens: Do you mean the `models` dictionary?

examples/README.md
```diff
@@ -8,8 +8,16 @@
 
 ## Running an example
 
-Say you want to solve the [`long_running`](./long_running.py) example locally. In a
-Python shell, execute:
+Say you want to solve the [`long_running`](./long_running.py) example locally. First,
+clone this repository and move into the example folder. In a console, type:
+
+```console
+$ git clone https://github.com/OpenSourceEconomics/lcm.git
```
hmgaudecker, line 15:
```suggestion
$ git clone https://github.com/OpenSourceEconomics/lcm.git
$ mamba env create
```

examples/README.md
```diff
@@ -8,8 +8,16 @@
...
+$ cd lcm/examples
```
hmgaudecker, line 16:
```suggestion
$ cd lcm/examples
$ ipython
```

examples/README.md
```diff
@@ -8,8 +8,16 @@
...
+```
+
+Make sure that you have `lcm` installed in your Python environment. Then, in a Python
```
hmgaudecker, line 19:
```suggestion
```

examples/README.md
```diff
@@ -8,8 +8,16 @@
...
+shell run
```
hmgaudecker, line 20:
```suggestion
In that shell, run the following code:
```

tests/test_analytical_solution.py
```diff
@@ -19,23 +21,38 @@
...
+    # remove age and wage functions, as they are not modelled in Iskhakov et al. (2017)
+    model_config["functions"] = {
+        name: func
+        for name, func in model_config["functions"].items()
+        if name not in ("age", "wage")
+    }
+    return model_config
+
+
+def _get_iskhakov_2017_params(disutility_of_work):
```
hmgaudecker, line 37: Why don't we make all parameters modifiable and leave the defaults at the values found in the main specification of the paper?

tests/test_analytical_solution.py
```diff
@@ -19,23 +21,38 @@
...
-ISKHAKOV_2017_PARAMS = {
-    "beta": 0.98,
-    "utility": {"disutility_of_work": None},
-    "next_wealth": {
-        "interest_rate": 0.0,
-        "wage": 20.0,
-    },
-}
+
+def _get_iskhakov_2017_model(n_periods):
```
hmgaudecker, line 25: Same comment as for the params below, but less important here.

tests/test_models/deterministic.py
```diff
@@ -38,14 +38,20 @@ def utility_with_filter(
     disutility_of_work,
     # Temporary workaround for bug described in issue #30, which requires us to pass
     # all state variables to the utility function.
-    lagged_retirement,  # noqa: ARG001, TODO: Remove unused arguments once #30 is fixed.
+    # TODO(@timmens): Remove unused arguments once #30 is fixed.
```
hmgaudecker, line 41:
```suggestion
    # TODO(@timmens): Remove function once #30 is fixed (just re-use "utility").
```

tests/test_simulate.py
```diff
@@ -116,12 +116,12 @@ def _model_solution(n_periods):
         solve_model, _ = get_lcm_function(model=model)
 
         params = {
```
hmgaudecker, line 118: See main comment.

hmgaudecker (commented):
There seems to be quite some repetetive code setting up the models/params. In `test_analytical_solution`, you are now using  `_get_iskhakov_2017_model` and `_get_iskhakov_2017_params`, which seems very nice. 

Would it make sense to re-factor that into a module that is used to generate fixtures across multiple test files?

PR #59, opened 2024-02-29
Add windows to CI, update testing and improve project config

### LINKED ISSUE #58: ENH: Run CI tests on Windows
It seems that jax has added support for Windows (see [here](https://jax.readthedocs.io/en/latest/installation.html)). In light of this, we should add CI for Windows.

timmens:
In this PR, I

- [x] Add support to run CI tests on windows-latest
- [x] Improve instructions in `examples/README.md`
- [x] Implement `get_model_config` and `get_params` functions for each model specification module (implementing the suggestions made in #56 after the merge) 
- [x] Move contents of `setup.cfg` to `pyproject.toml`
- [x] Add a cuda-related package to environment for GPU computing

> [!NOTE]
> The idea of the new model config interface is that in each model specification module (`deterministic.py`, `stochastic.py`) functions `get_model_config` and `get_params` are implemented that follow the same interface. Perspectively, we could think about a single combined function for all model configs and params with unique model names.

Closes #58


tests/test_models/stochastic.py
```diff
@@ -35,7 +37,7 @@ def utility(
     health,
     # Temporary workaround for bug described in issue #30, which requires us to pass
     # all state variables to the utility function.
-    # TODO(@timmens): Remove unused arguments once #30 is fixed.
+    # TODO(@timmens): Remove function arguments once #30 is fixed (re-use "utility").
```
hmgaudecker, line 40:
```suggestion
    # TODO(@timmens): Remove function once #30 is fixed (re-use "utility").
```

hmgaudecker (approved):
Very nice! Much better setup for CI etc. and the changes that should have gone into the previous PR make things much clearer!

hmgaudecker:
Only thing that occurred to me later: We should add some description to the models in `deterministic.py` etc. In a way, they are all based on the DC-EGM paper. So it is a bit confusing to call one `ISKHAKOV_ET_AL_2017` and the others `BASE` (+variations). But let's best discuss in person!

tests/test_models/stochastic.py
timmens: So in stochastic.py it is the function argument, since there is only one utility function. In deterministic.py it is the entire function. But I will remove the "(re-use 'utility')" from stochastic.py

tests/test_models/deterministic.py
```diff
@@ -93,15 +95,15 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
 # Model specifications
 # ======================================================================================
 
-BASE_MODEL = {
+ISKHAKOV_ET_AL_2017 = {
     "functions": {
```
hmgaudecker, line 99: Add description, just for completeness. Briefly explain why wage/age not need.

tests/test_models/deterministic.py
```diff
@@ -119,12 +121,17 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
...
+        "lagged_retirement": {"options": [0, 1]},
     },
     "n_periods": 3,
 }
 
 
-BASE_MODEL_FULLY_DISCRETE = {
+ISKHAKOV_ET_AL_2017_STRIPPED_DOWN = {
+    "description": (
+        "Starts from Iskhakov et al. (2017), removes filters, the lagged_retirement"
```
hmgaudecker, line 132:
```suggestion
        "Starts from Iskhakov et al. (2017), removes filters and the lagged_retirement"
```

tests/test_models/deterministic.py
```diff
@@ -150,26 +162,21 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
 }
 
 
-BASE_MODEL_WITH_FILTERS = {
+ISKHAKOV_ET_AL_2017_FULLY_DISCRETE = {
+    "description": (
+        "Starts from Iskhakov et al. (2017), removes filters, the lagged_retirement "
```
hmgaudecker, line 167:
```suggestion
        "Starts from Iskhakov et al. (2017), removes filters and the lagged_retirement "
```

tests/test_models/stochastic.py
```diff
@@ -113,69 +115,109 @@ def consumption_constraint(consumption, wealth):
...
-        # 1st dimension: Current health state
-        # 2nd dimension: Current Partner state
-        # 3rd dimension: Probability distribution over next period's health state
-        "health": jnp.array(
+# ======================================================================================
+# Get models and params
+# ======================================================================================
+
+IMPLEMENTED_MODELS = {
+    "base": MODEL_CONFIG,
```
hmgaudecker, line 123:
```suggestion
    "only_discrete_vars_stochastic": MODEL_CONFIG,
```

hmgaudecker (approved):
:+1:

PR #62, opened 2024-03-21
Update argmax.py

timmens:
In this PR, we update everything related to `argmax.py`.

- [x] Add type hints (See [JAX Typing Best Practices](https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices))
- [x] Run mypy and add it to CI
- [x] Add an explanation of what `argmax` and `segment_argmax` do to the developer notes
  - [x] Briefly explain what the functions do on a programming level
  - [x] Draw a connection between the math notation of the dynamic programming model and functions


src/lcm/argmax.py
```diff
@@ -1,31 +1,42 @@
 import jax
```
hmgaudecker, line 1:
```suggestion
from jax.ops import segment_max
```
I think we'll generally want to use that as a convention, there are only very few instances where `import jax` is used. 

As a convention, I'd suggest:
- leave `jnp`
- `jax.random` becomes `jrnd`
- everything else via `from jax.x import y`

Just a note: In `test_process_model`, we are using both `jnp` and `jax.numpy` explicitly -- any subtle reason for that or just local optimization?

hmgaudecker: > For the `jax.random` module I believe its best to follow the examples from the `jax` documentation and do `from jax import random`. 

To the extent that this will never ever conflict with anything a user might do with `numpy.random`, fine with me!

> We do not need it that often currently anyways.

Other than `jax.numpy`, I think that was the most frequently used thing!

src/lcm/argmax.py
```diff
@@ -113,13 +146,13 @@ def segment_argmax(a, segment_ids, num_segments):
 
     # Check where the array attains its maximum
     # ==================================================================================
-    max_value_mask = a == segment_max_expanded
+    max_value_mask = data == segment_max_expanded
 
     # Create index array of argmax indices for each segment (has same shape as a)
```
hmgaudecker, line 151:
```suggestion
    # Create index array of argmax indices for each segment (has same shape as data)
```

Are we sure that a unique value is returned in case of ties for maxima? *(assuming we rely on that implicitly in interpolations etc)* Do we have a test for that?

src/lcm/argmax.py
```diff
@@ -1,31 +1,43 @@
 import jax
 import jax.numpy as jnp
+from jax import Array
```
hmgaudecker, line 3: Would look nicer IMO if we imported both from `jax.typing`, assuming that has the same effect.

src/lcm/argmax.py
```diff
@@ -87,24 +100,44 @@ def _flatten_last_n_axes(a, n):
...
+        - Array: Array with shape (num_segments, *a.shape[1:]). The maximum
+            value for the k-th segment.
+
     """
+    # Preparation
+    # ==================================================================================
+    data = jnp.asarray(data)
+    segment_ids = jnp.asarray(segment_ids)
+
     # Compute segment maximum and bring to the same shape as a
```
hmgaudecker, line 137:
```suggestion
    # Compute segment maximum and bring to the same shape as data
```

src/lcm/argmax.py
timmens: `Array` is not available in the `jax.typing` module because it is not a type per se. The way I do it here is copied from the [JAX Typing Best Practices](https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices) documentation section.

hmgaudecker: Sorry for the noise!

src/lcm/argmax.py
timmens: I've added a test for `argmax` and `segment_argmax` and a corresponding note in the code.

hmgaudecker (commented):
Thanks, those two tests are very helpful! *(if only to explain obvious behaviour to half-knowledgeable people like me)*

Just leaving this note here so I can focus on other stuff in an actual review.

src/lcm/argmax.py
```diff
@@ -1,31 +1,45 @@
...
+            the dimensions specified in axis, which are dropped. The value corresponds
+            to an index that can be translated into a tuple of indices using
+            jnp.unravel_index.
+
+        - Array: The corresponding maximum values.
 
     """
     # Preparation
     # ==================================================================================
+    a = jnp.asarray(a)
```
hmgaudecker, line 41: I am sure you do this and the typing to ArrayLike instead of Array for a reason, but my naive mind thought we'd always have jnp.arrays only. Maybe add a hint somewhere why we can't assume that?

src/lcm/argmax.py
```diff
@@ -39,25 +53,28 @@ def argmax(a, axis=None, initial=None, where=None):
     # Do same transformation for where
     # ==================================================================================
     if where is not None:
+        where = jnp.asarray(where)
```
hmgaudecker, line 56: Same same.

src/lcm/argmax.py
```diff
@@ -87,24 +104,45 @@ def _flatten_last_n_axes(a, n):
...
             for the k-th segment will be in jnp.arange(segment_ids[k]).
 
+        - Array: Array with shape (num_segments, *data.shape[1:]). The maximum
+            value for the k-th segment.
+
     """
-    # Compute segment maximum and bring to the same shape as a
+    # Preparation
+    # ==================================================================================
+    data = jnp.asarray(data)
```
hmgaudecker, line 139: See comments above in `argmax`.

src/lcm/argmax.py
```diff
@@ -1,31 +1,45 @@
...
 
 # ======================================================================================
 # argmax
 # ======================================================================================
 
 
-def argmax(a, axis=None, initial=None, where=None):
-    """Compute the argmax of a n-dim array along axis.
+def argmax(
+    a: ArrayLike,
```
hmgaudecker, line 12: Why do we use `a` here and `data` in `segment_argmax`?

hmgaudecker (approved):
Excellent! Some questions, but no show-stopper. Thanks a lot!

src/lcm/argmax.py
timmens: We follow the naming convention used in JAX and Tensorflow. I believe they use `data` for these segment operations to highlight that the operations are not directly performed on the input array.

src/lcm/argmax.py
timmens: > I am sure you do this and the typing to ArrayLike instead of Array for a reason

This does not mean that the reason is a good one!

After reading a lot about best practices with JAX typing, I got the impression that a standard approach is weak input types and strict output types (see [here](https://jax.readthedocs.io/en/latest/jax.typing.html#jax-typing-best-practices)).

However, as you are implicitly pointing out if we never use the weaker input types internally, then there is also no reason at all not to use the stricter input type.

I've updated this, thank you!

PR #66, opened 2024-04-10
Update create_params.py

timmens:


src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
-        user_model (dict): A model specification. Has keys
+        model_spec (dict): A model specification provided by the user. Has keys
             - "functions": A dictionary of functions used in the model.
             - "choices": A dictionary of choice variables.
             - "states": A dictionary of state variables.
             - "n_periods": Number of periods in the model (int).
             - "shocks": A dictionary of shock variables (optional).
         variable_info (pd.DataFrame): A dataframe with information about the variables.
-        grids (dict): A dictionary of grids.
+        grids (dict): A dictionary of grids retrieved from model_spec.
```
hmgaudecker, line 27:
```suggestion
        grids (dict): A dictionary of grids consistent with model_spec.
```
Open to other suggestions, but the original version triggered me to think about whether we need it at all.

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
+        default_params (dict): A dictionary of default parameters. Default is None. If
+            None, the default {"beta": np.nan} is used. For other lifetime reward
+            objectives, additional parameters may be required, for example {"beta":
+            np.nan, "delta": np.nan} for beta-delta discounting.
 
     Returns:
-        dict: A dictionary of model parameters.
+        dict: A (nested) dictionary of model parameters.
```
hmgaudecker, line 34:
```suggestion
        dict: A nested dictionary of model parameters.
```
Unless it can actually be non-nested outside of test cases.

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
+    return params_template
 
 
-def _create_function_params(model):
+def _create_function_params(model_spec: dict[str, Any]) -> dict[str, dict[str, Any]]:
     """Get function parameters from a model specification.
 
-    The argument '_period' is handled separately. It is not treated as a parameter, but
-    as the period of the model. Functions like 'age' can depend on it.
+    Note: The function argument '_period' is handled separately. It is not treated as a
```
hmgaudecker, line 57: Too prominent, I think. Could go between Args and Returns, maybe? But maybe irrelevant once previous point is tackled.

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
             grids=grids,
         )
 
-    return params
+    return params_template
 
 
-def _create_function_params(model):
+def _create_function_params(model_spec: dict[str, Any]) -> dict[str, dict[str, Any]]:
     """Get function parameters from a model specification.
```
hmgaudecker, line 55: Would be good to explain a little bit what exactly is being done. IIUC:
- Look at grand DAG
- Remove all arguments that are states, choices, user-provided functions, etc (i.e., things LCM knows from the model spec)
- Assume everything else is a parameter that will need to be provided by the user.

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
+    for var in stochastic_variables:
 
+        # Retrieve corresponding next function and its arguments
+        next_var = model_spec["functions"][f"next_{var}"]
+        dependencies = list(inspect.signature(next_var).parameters)
+
+        invalid = set(dependencies) - valid_vars
+        # If there are invalid dependencies, store them in a dictionary and continue
+        # with the next variable to collect as many invalid arguments as possible.
+        if invalid:
```
hmgaudecker, line 152:
```suggestion
        if invalid := set(dependencies) - valid_vars:
```
Walrus ftw!

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
 
-def _create_standard_params():
-    return {"beta": np.nan}
+    for var in stochastic_variables:
 
+        # Retrieve corresponding next function and its arguments
+        next_var = model_spec["functions"][f"next_{var}"]
+        dependencies = list(inspect.signature(next_var).parameters)
+
+        invalid = set(dependencies) - valid_vars
```
hmgaudecker, line 149:
```suggestion
```

src/lcm/create_params.py
```diff
@@ -1,138 +1,176 @@
...
+        # If there are invalid dependencies, store them in a dictionary and continue
+        # with the next variable to collect as many invalid arguments as possible.
+        if invalid:
+            invalid_dependencies[var] = invalid
+            continue
```
hmgaudecker, line 154: I kinda prefer if/else and adding the comment inside the if-statement instead of this style, but I won't argue over it.

hmgaudecker (approved):
Just a bit of docs and style. Very nice!

PR #67, opened 2024-04-11
Update discrete_emax.py

timmens:
In this PR, I update the `discrete_argmax.py` module by

- [x] Consistent section comments in `create_params.py` (forgotten in #66)
- [x] Rename `discrete_emax.py` -> `discrete_problem.py`
  This is a first step towards the upcoming refactoring.
- [x] Adding type hints
- [x] Custom type for `segment_info` in new `lcm.typing` module
  Perspectively, this object will likely become a NamedTuple once we refactor `state_space.py`, but for now, a `TypedDict` should suffice.
- [x] Write more about choice axes, the value function array's dimensions, etc.
- [x] Raise error if any shock other than `None` is requested
  This is because, currently, there is no infrastructure for utility shocks.
- [x] Improve and update docstrings
- [x] Explaining what `solve_discrete_problem` solves in the math background notes

> [!NOTE]
> I do not update the aggregation with extreme value shocks in this PR. This is noted and will be tackled once we add the infrastructure for utility (taste) shocks.


tests/test_discrete_problem.py
```diff
@@ -90,17 +97,15 @@ def test_aggregation_with_extreme_value_shocks(
...
     )
 
-    calculator = get_emax_calculator(
-        shock_type="extreme_value",
-        variable_info=var_info,
-        is_last_period=False,
+    choice_axes = _determine_dense_discrete_choice_axes(var_info)
+    solve_discrete_problem = partial(
+        _calculate_emax_extreme_value_shocks,
+        choice_axes=choice_axes,
```
timmens, line 103: We do this manually until support for utility/taste shocks is added. Then we will change this back.

src/lcm/create_params.py
hmgaudecker, line 1: Should we change the name of the module, too?

src/lcm/create_params.py
```diff
@@ -119,9 +119,8 @@ def _create_stochastic_transition_params(
     """
     stochastic_variables = variable_info.query("is_stochastic").index.tolist()
 
-    # ----------------------------------------------------------------------------------
     # Assert that all stochastic variables are discrete state variables
-    # ----------------------------------------------------------------------------------
+    # ==================================================================================
```
hmgaudecker, line 123: I am not a huge fan of these things, but fine with them if they won't ever change again... Can you do a bulk search & replace, please? *(maybe this is exactly what happened here, then just disregard)*

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+import jax.numpy as jnp
+import pandas as pd
+from jax import Array
+
+from lcm.typing import SegmentInfo
+
+
+def get_solve_discrete_problem(
+    *,
+    shock_type: Literal["extreme_value"] | None,
```
hmgaudecker, line 34:
```suggestion
    random_utility_shock_type: Literal["extreme_value"] | None,
```
Better be explicit here, there might be other types of randomness within period, as just discussed.

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+        is_last_period (bool): Whether the function is created for the last period.
+        choice_segments (SegmentInfo): Dictionary with the entries "segment_ids" and
+            "num_segments". segment_ids are a 1d integer array that partitions the first
+            dimension of values into choice sets over which we need to aggregate.
+            "num_segments" is the number of choice sets. The segment_ids are assumed to
+            be sorted.
+        params (dict): Dictionary with model parameters.
+
+    Returns:
+        callable: Function that calculates the expected maximum of the conditional
```
hmgaudecker, line 57: I think I'd like to see the full interface here, I find the description a bit confusing. Maybe it's also the type hint `Callable[[Array], Array]`, which I don't quite get. Is the first `[Array]` the input and the second the output?

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+def _solve_discrete_problem_no_shocks(
+    cc_values: Array,
+    choice_axes: tuple[int, ...] | None,
+    choice_segments: SegmentInfo | None,
+    params: dict[str, Any],  # noqa: ARG001
+) -> Array:
+    """Aggregate conditional continuation values over discrete choices.
+
+    Args:
+        cc_values (jax.Array): Array with conditional continuation values. Has one
```
hmgaudecker, line 96:
```suggestion
        cc_values (jax.Array): Array with conditional continuation values. For each
```

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+            dimensions per state and discrete choice variable.
```
hmgaudecker, line 97:
```suggestion
            state and discrete choice variable, it has one axis.
```
Not beautiful, but a little clearer this way around.

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+        choice_axes (tuple[int, ...]): A tuple of indices representing the axes in the
+            value function that correspond to discrete choices. Returns None if there
+            are no discrete choice axes.
+        choice_segments (SegmentInfo): Dictionary with the entries "segment_ids" and
```
hmgaudecker, line 101: I don't think we should repeat all of this, seems to make maintenance harder than necessary. It is in the docstring just above, right?

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+    """Get function that computes the expected max. of conditional continuation values.
+
+    The maximum is taken over the discrete and sparse choice variables in each state.
+
+    Args:
+        shock_type (Literal["extreme_value"] | None): Type of choice shock. Currently
+            only None is supported. Work for "extreme_value" is in progress.
+        variable_info (pd.DataFrame): DataFrame with information about the variables.
+        is_last_period (bool): Whether the function is created for the last period.
+        choice_segments (SegmentInfo): Dictionary with the entries "segment_ids" and
```
hmgaudecker, line 49: I think more important than the precise entries is the description of the content: This is how sparse choices are stored, right?

src/lcm/discrete_problem.py
```diff
@@ -0,0 +1,280 @@
...
+            sorted.
+
+    Returns:
+        jax.Array: An array with shape (num_segments,) + data.shape[1:] representing the
+            segment maximums.
+
+    """
+    return jax.ops.segment_max(
+        data=data,
+        indices_are_sorted=True,
```
hmgaudecker, line 146: Wouldn't it make sense to make this part of `segment_info`? You see it that it is always True at the point where it is defined then.

This function would then be gone, I guess.

hmgaudecker (approved):
I like the changes a lot! Just a bit more big pic documentation would be good. Maybe there's a little room for simplification. And I only skimmed the tests, but they look good!

src/lcm/discrete_problem.py
timmens: Yes, Callable type hints work like this: `Callable[[Input Types], Output Type]`.

src/lcm/discrete_problem.py
timmens: That is a good idea. I cannot add it to `SegmentInfo` right now because doing so would require me to change a large chunk of the code base that I want to refactor at some point. But I have noted it, and I will already delete the wrapper function.

PR #68, opened 2024-04-13
Update dispatchers.py

timmens:
In this PR, I update the `dispatchers.py` module.

- [x] Improve docstrings
- [x] Add type hints
- [x] Check whether I need to add additional tests
- [x] Create an example notebook
  I've added a folder, `explanations`, in which I collect explanation notebooks for the `lcm` internals. This is not ideal yet, but we can easily change it. 
- [x] Update pre-commit hooks
- [x] Delete `distributions.py`
  This is outdated and not used anywhere.

**Notes:**

In line with JAX's [`vmap function`](https://jax.readthedocs.io/en/latest/_modules/jax/_src/api.html#vmap) we use `F = TypeVar("F", bound=Callable)` for the type hint of the function that is dispatched. This signals that the object returned is the same as the input, except that it is now vectorized.


src/lcm/dispatchers.py
```diff
@@ -40,42 +50,53 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
             above but there might be additional dimensions.
 
     """
-    func = allow_args(func)  # vmap cannot deal with keyword-only arguments
-
+    # Check inputs
+    # ==================================================================================
     if not set(dense_vars).isdisjoint(sparse_vars):
```
hmgaudecker, line 55:
```suggestion
    if overlap := set(dense_vars).intersection(set(sparse_vars)):
```
*(untested)* Then print the problematic stuff in the error message. Same for the other two checks. I know it's pedantic, but this will help tremendously for newcomers.

src/lcm/dispatchers.py
```diff
@@ -40,42 +50,53 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
...
         raise ValueError("dense_vars and sparse_vars overlap.")
 
-    _all_variables = dense_vars + sparse_vars
-    if len(_all_variables) != len(set(_all_variables)):
-        raise ValueError("Same argument provided more than once.")
+    if len(set(dense_vars)) < len(dense_vars):
```
hmgaudecker, line 58:
```suggestion
    if duplicates := {v for v in dense_vars if dense_vars.count(v) > 1}:
```
*(untested)* Same below.

src/lcm/dispatchers.py
```diff
@@ -133,45 +157,52 @@ def vmap_1d(func, variables):
             pytree are as described above but there might be additional dimensions.
 
     """
-    if len(variables) != len(set(variables)):
+    if len(set(variables)) < len(variables):
```
hmgaudecker, line 160: As above, more meaningful error.

hmgaudecker (commented):
Just had a glance rather than deleting the e-mail right away :-)

src/lcm/dispatchers.py
```diff
@@ -40,42 +50,53 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
...
     parameters = list(signature.parameters)
 
-    positions = [parameters.index(cv) for cv in sparse_vars]
-    in_axes = []
-    for i in range(len(parameters)):
-        if i in positions:
-            in_axes.append(0)
+    sparse_positions = [parameters.index(cv) for cv in sparse_vars]
+
+    sparse_in_axes = []  # type: list[int | None]
```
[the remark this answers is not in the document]
timmens: `in_axes_for_sparse_vars`? With the name change, I wanted to make clear that these `in_axes` are used by a `vmap` call vectorizing the function over the sparse variables.

src/lcm/dispatchers.py
```diff
@@ -1,24 +1,34 @@
...
+F = TypeVar("F", bound=Callable)
 
-def spacemap(func, dense_vars, sparse_vars, *, dense_first):
+
+def spacemap(
+    func: F,
+    dense_vars: list[str],
+    sparse_vars: list[str],
+    *,
+    dense_first: bool,
```
hmgaudecker, line 17:
```suggestion
    put_dense_first: bool,
```
Makes it clear that this is about outputs.

src/lcm/dispatchers.py
```diff
@@ -1,24 +1,34 @@
...
     In contrast to vmap, spacemap preserves the function signature and allows the
     function to be called with keyword arguments.
 
     Args:
         func (callable): The function to be dispatched.
-        dense_vars (list): Names of the dense variables, i.e. those that are simply
-            stored as array of possible values in the grid because the possible values
-            of those variable does not depend on the value of other variables.
+        dense_vars (list): Names of the dense variables, i.e. those that are stored
+            as arrays of possible values in the grid, because the possible values
```
hmgaudecker, line 30:
```suggestion
            as arrays of possible values in the grid.
```
Remove rest of sentence, too. Nobody should look here to understand the concept. Could add link to dev docs or whatever the right place is, though.

src/lcm/dispatchers.py
```diff
@@ -1,24 +1,34 @@
...
+            of those variables do not depend on the value of other variables.
         sparse_vars (list): Names of the sparse variables, i.e. those that are stored
             as arrays of possible combinations of variables in the grid because the
```
hmgaudecker, line 33:
```suggestion
            as arrays of possible combinations of variables in the grid.
```
Same same.

src/lcm/dispatchers.py
```diff
@@ -40,42 +50,59 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
...
     parameters = list(signature.parameters)
 
-    positions = [parameters.index(cv) for cv in sparse_vars]
-    in_axes = []
-    for i in range(len(parameters)):
-        if i in positions:
-            in_axes.append(0)
+    positions_of_sparse_vars = [parameters.index(cv) for cv in sparse_vars]
+
+    in_axes_for_sparse_vars = []  # type: list[int | None]
```
hmgaudecker, line 79: Can you add a little explainer what this does? A list with zeros and Nones... I don't have an intuition for that.

src/lcm/dispatchers.py
```diff
@@ -40,42 +50,59 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
...
+        if p in positions_of_sparse_vars:
+            in_axes_for_sparse_vars.append(0)
         else:
-            in_axes.append(None)
+            in_axes_for_sparse_vars.append(None)
 
+    # Apply vmap for sparse and _product_map for dense variables
+    # ==================================================================================
     if not sparse_vars:
         vmapped = _product_map(func, dense_vars)
```
hmgaudecker, line 89: Very confusing to have `productmap` and `_product_map` in the same module. Can we add something to the private function?

src/lcm/dispatchers.py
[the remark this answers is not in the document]
hmgaudecker: I think I'd prefer `in_axes_for_vmap`, makes it clear where it will be used and there's a 1:1 mapping to `sparse_vars`, which is very clear once looking at the code description there.

src/lcm/dispatchers.py
```diff
@@ -40,42 +50,59 @@ def spacemap(func, dense_vars, sparse_vars, *, dense_first):
...
-
     signature = inspect.signature(func)
     parameters = list(signature.parameters)
 
-    positions = [parameters.index(cv) for cv in sparse_vars]
-    in_axes = []
-    for i in range(len(parameters)):
-        if i in positions:
-            in_axes.append(0)
+    positions_of_sparse_vars = [parameters.index(cv) for cv in sparse_vars]
```
hmgaudecker, line 77:
```suggestion
    positions_of_sparse_vars = [parameters.index(sv) for sv in sparse_vars]
```
I don't have an intuition for cv, at least.

src/lcm/dispatchers.py
timmens: `_base_productmap` with more info in the respective docstrings?

src/lcm/dispatchers.py
timmens: Very good point!

src/lcm/dispatchers.py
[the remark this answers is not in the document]
timmens: Not entirely convinced, but I also do not have such a strong opinion on this. My last suggestion would be `in_axes_sparse_vmap`. If you don't like that we can stick with `in_axes_for_vmap`.

src/lcm/dispatchers.py
hmgaudecker: Excellent, thanks!

src/lcm/dispatchers.py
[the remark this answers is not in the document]
hmgaudecker: Either `in_axes_for_sparse_vars` or `in_axes_for_vmap`, the mixture makes it more confusing :laughing:

In the end it depends on how familiar you expect readers of that snippet to be with `vmap`'s interface. I am not, hence I would prefer the latter. But if it is clear to most readers where the name `in_axes` comes from, the former is better. Leaving this up to you!

explanations/dispatchers.ipynb
```diff
@@ -0,0 +1,688 @@
+{
```
hmgaudecker, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    a = jnp.linspace(0, 1, 11)
</pre><p> </p><p>Would we see just the decimals then? I find the many digits a bit distracting...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/lcm/pull/68/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

hmgaudecker, line 1: <p>I am not sure about the "awkward" terminology. It is just sparse / some combinations of states and choices are invalid. </p><p> </p><p>Isn't it that the spacemap function combines productmap and vmap_1d? Then we should say so without much ado first. Second, a sentence why the need arises (dense / sparse). Third, the example.</p><p> </p><p>Introduce the example with this sentence and remove the last paragraph:</p><p> </p><p>Consider a simplified version of our deterministic test model. Curly brackets {} denote discrete variables; square brackets [] stand for continuous variables.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/lcm/pull/68/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

hmgaudecker, line 1: <p>Add a sentence in between the two paragraphs talking about memory usage and computational efficiency?</p><p> </p><ul><li>for dense variables, need only to store a row vector for each variable, then take the product: Less storage now, more computations, more storage for results </li><li>for continuous variables, need to store one row for each valid state x choice combination: More storage now (6 elements instead of 4), less computations , lower storage space for results (both 3 instead of 4)</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/lcm/pull/68/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

hmgaudecker, line 1: <p>A short explainer why we expect these particular results would be great. You don't want to go through the functions to understand what is happening when reading this.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/lcm/pull/68/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

hmgaudecker, line 1: <p>See previous comment -- first explain what is happening, then say that consumption will of course take on different values and we'll want to vectorize over it.</p><p> </p><p>It is not fully obvious to me why we have spacemap applying sparse choices but need to do a manual productmap for continuous choices. At least seems to merit a brief explanation!</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenSourceEconomics/lcm/pull/68/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

src/lcm/dispatchers.py
hmgaudecker, line 1: Did anything change since I looked at it last Saturday?

hmgaudecker (commented):
Very nice! 

Just terminology, I think -- and let me know whether I should take another good look at dispatchers.

src/lcm/dispatchers.py
timmens: Not really, just integrated your comments.

timmens:
@hmgaudecker please re-review the PR. I made the following additional changes:

- I tried to integrate your comments on the dispatchers explanation notebook
- I added a boolean flag to `vmap_1d` to choose whether the `allow_kwargs` decorator should be applied before returning the dispatched function, which allowed me to rewrite `spacemap` using only `productmap` and `vmap_1d`.

src/lcm/dispatchers.py
```diff
@@ -99,19 +110,24 @@ def productmap(func, variables):
...
+    vmapped.__signature__ = signature  # type: ignore[attr-defined]
 
     return allow_kwargs(vmapped)
 
 
-def vmap_1d(func, variables):
+def vmap_1d(func: F, variables: list[str], *, apply_allow_kwargs: bool = True) -> F:
     """Apply vmap such that func is mapped over the specified variables.
 
     In contrast to vmap, vmap_1d preserves the function signature and allows the
```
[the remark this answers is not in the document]
timmens: Yes. It is less flexible than `jax.vmap` in terms of vectorization because it is designed to only be called when we have sparse variables where we call the dispatched function on array inputs, and we do not want a Cartesian product but a row-wise execution.

explanations/dispatchers.ipynb
```diff
@@ -0,0 +1,825 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "\n",
+    "### Explanation of Results\n",
+    "\n",
+    "The outputs align with the utility function: The rows represent sparse combinations of _lagged_retirement_ and _retirement_, while the columns represent values of _wealth_. For column $j$, the value of _wealth_ corresponds to $j$.\n",
+    "\n",
+    "Consider the first row, corresponding to _lagged_retirement_ being 0 and _retirement_ also being 0. In this scenario, the agent is working, incurring a cost of -0.5. Since the agent is not retired, there is no bequest utility, resulting in a utility of $log(1) - 0.5 = -0.5$ for all _wealth_ values.\n",
```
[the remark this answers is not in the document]
timmens: Yes, you are completely right; this is nonsense.

Because it works very nicely for the example, I want a term in the utility function that is 0 if  _lagged_retirement_ is 0 and equal to _wealth_ if _lagged_retirement_ is 1.

My current idea would be to call it _retirement utility_, or do you have a different suggestion? The idea would be that once you are retired for at least one period, you get utility corresponding to your wealth level.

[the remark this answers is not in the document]
hmgaudecker: `retirement_habit` ? Seems closest to a habit formation thing. 

Not worth spending much time on it, though, just bequests are too misleading imo.

PR #70, opened 2024-04-24
Remove unused interpolation module

timmens:
In this PR, I remove the interpolation module and the sandbox notebooks that depend on it.

The module `interpolation.py` is not used in lcm. The `function_evaluator.py` module implements its own interpolator. Some sandbox notebooks also use functions from `interpolation.py` but no other functions from lcm.

@janosg: Can this be safely deleted, or were there plans to extend this at some point?

Otherwise, I would delete them here and potentially add them to the lcm-dev repo for historical illustration purposes.


janosg:
Yes, just delete. No need to keep it.

PR #71, opened 2024-04-24
Update functools.py

timmens:
In this PR, I

- [x] Add type hints to `functools.py`
- [x] Replace the `allow_kwargs` decorator with the `allow_only_kwargs` decorator
- [x] Open issue for further improvements of decorator return types (#80) 

The more general decorator `allow_kwargs` is replaced with the more specific `allow_only_kwargs` to reduce complexity when developing (i.e., I do not have to think about any mixed-case scenarios) and to throw errors more quickly and with more informative error messages. When working with functions internally, we either want to call them with keyword-only arguments (in which case I do not need the possibility to call them without keyword arguments) or with positional arguments. In the latter case we do not gain anything by implementing an `allow_only_args` decorator, which is why we stick with the more general one here.


src/lcm/functools.py
```diff
@@ -67,46 +27,66 @@ def allow_kwargs(func):
...
     ]
     new_signature = signature.replace(parameters=new_parameters)
 
     @functools.wraps(func)
-    def allow_kwargs_wrapper(*args, **kwargs):
+    def func_with_only_kwargs(*args, **kwargs):
+        if args:
+            raise ValueError(
+                (
+                    "This function was decorated with allow_only_kwargs, but was "
```
hmgaudecker, line 39:
```suggestion
                    "This function has been decorated so that it allows only kwargs, but was "
```
Decorator has been renamed, better use language.

src/lcm/simulate.py
```diff
@@ -445,7 +445,7 @@ def retrieve_non_sparse_choices(indices, grids, grid_shape):
     return out
 
 
-@partial(vmap_1d, variables=["index"])
+@partial(vmap_1d, variables=["index"], apply_allow_kwargs=False)
```
hmgaudecker, line 448:
```suggestion
@partial(vmap_1d, variables=["index"], callable_with="kwargs")
```
Argument will definitely have to be renamed in `vmap_1d`. This is just a suggestion -- instead of Boolean with an awkward name, use Literal that could be `args` or `kwargs`. But there could be better ideas, I did not think about it for very long.

tests/test_dispatchers.py
```diff
@@ -96,21 +96,25 @@ def test_productmap_with_all_arguments_mapped(func, args, grids, expected, reque
...
 
     decorated = productmap(func, args)
-    calculated_args = decorated(*grids.values())
-    calculated_kwargs = decorated(**grids)
+    calculated = decorated(**grids)
+    aaae(calculated, expected)
 
-    aaae(calculated_args, expected)
-    aaae(calculated_kwargs, expected)
+    match = (
```
hmgaudecker, line 102: Shouldn't we add another test for this? I find it a bit confusing to have two in one. If it runs for very long and that is not an option, make a note?

tests/test_functools.py
```diff
@@ -119,47 +120,49 @@ def test_convert_kwargs_to_args():
...
 
     with pytest.raises(TypeError):
         f(a=1, b=2)
 
-    assert allow_kwargs(f)(a=1, b=2) == 3
+    assert allow_only_kwargs(f)(a=1, b=2) == 3
 
 
-def test_allow_kwargs_incorrect_number_of_args():
+def test_allow_only_kwargs_incorrect_number_of_args():
```
hmgaudecker, line 144: Also for the rest, I would prefer "one test, one assert statement". But those are obvious enough that it is not worth to put more than 20min into it.

tests/test_simulate.py
```diff
@@ -358,8 +358,8 @@ def test_process_simulated_data():
 
 
 def test_retrieve_non_sparse_choices():
-    got = _retrieve_non_sparse_choices(
-        index=jnp.array([0, 3, 7]),
+    got = retrieve_non_sparse_choices(
+        indices=jnp.array([0, 3, 7]),
```
hmgaudecker, line 362: In simulate.py, is there a reason for having `indices` and `index` differently in the public and private version of `retrieve_non_sparse_choices`? If not, this might be a good occasion to harmonise them.

hmgaudecker (approved):
Looks great! Just a few renaming suggestions and general style of tests.

tests/test_simulate.py
timmens: The reason was that the private function was written to work on a single index and then vmapped. Noticing that I only have to vmap `jnp.unravel_index` I rewrote this part for clarity.

PR #73, opened 2024-05-08
Precompile CCV Functions

mj023:
- [X] Let Jax use the gpu backend if available
- [x] Remove inefficient Jit Operations
- [x] Concurrently precompile all functions so that they are ready when needed in each period


mj023:
It turns out, that just jitting the function that get_lcm_function() returns increases the speed for one iteration, while simultaneously drastically increasing the speed if you want to solve the model multiple times with different parameters e.g. the parameter estimation use case. While precompiling is still faster for the first iteration, it is not possible to both precompile the functions inside the returned function and then jit everything. Therefore I will close this pull request.

mj023:
It turns out, that just jitting the function that get_lcm_function() returns increases the speed for one iteration, while simultaneously drastically increasing the speed if you want to solve the model multiple times with different parameters e.g. the parameter estimation use case. While precompiling is still faster for the first iteration, it is not possible to both precompile the functions inside the returned function and then jit everything. Therefore I will close this pull request.

timmens:
Regarding the previous comment, see #77

PR #77, opened 2024-05-29
Jit the returned lcm function

mj023:
In this PR we want to jit the function that is returned when we tell lcm to solve a model. This speeds up the execution time when solving a model for the first time and greatly reduces the execution time when solving the same model with different parameters.

- [X] Let Jax use the gpu backend if available
- [X] Jit the returned lcm function


timmens (approved):
Looks great, thank you. This is a much nicer solution compared to #73.

PR #81, opened 2024-06-14
Update function evaluator

timmens:
In this PR, I update the code corresponding to the function evaluator.

- [x] Add typing to `function_evaluator.py` module
- [x] Replace all occurrences of non-numerical discrete labels. (We will not support them.)
  For example, we test the lookup with characters ["good", "bad"], which we do not want to support.
- [x] Replace all occurrences of numerical labels that are not themselves indices (i.e. [0, 1, ..., endpoint] is okay, but [1, 2] is not)
- [x] Rewrite label translator to only return identity function (works on the assumption that all discrete labels that are processed internally are their own indices)
- [x] Throw error when processing a model with a discrete grid that does not correspond to indices
- [x] Add explanation notebook on function evaluator
  - [x] Write motivation presenting the high-level function
  - [x] Write technical section on behavior of small functions
  - [x] Add task to run explanation notebooks on CI (prevent misaligned of LCM codebase and these notebooks)
- [x] Compare `map_coordinates` to other interpolation tools, e.g., `RegularGridInterpolator`
  - It turns out that the `RegularGridInterpolator` implements a linear extrapolation feature. However, it is prohibitively slower compared to `map_coordinates.` Luckily, we can adjust `map_coordinates` to also perform linear extrapolation. See notes below.

> [!NOTE]
> 1. This PR introduces a new error that will be thrown during the model processing. It was already discouraged to use discrete grids that are not indices (since other grids were not implemented and would throw an uninformative error). Now, an informative error is thrown if the user tries to use a grid that does not coincide with indices.
> 2. During work on this PR, Janos and I discussed options to implement arbitrary discrete grids. We settled on a solution that is described in #82. We do not implement it here, because it will be easier to implement once the model processing has been refactored.
> 3. While I found a very nice solution to add extrapolation to our function evaluator, we will not tackle any extrapolation issues in this PR since we have to wait for a response from a JAX maintainer; see issue #83 for details.


explanations/function_evaluator.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "\n",
+    "wealth_grid = processed_model.grids[\"wealth\"]\n",
```
hmgaudecker, line 317: Graph should not connect the dots at this point.

explanations/function_evaluator.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Visualizing the results\n",
+    "\n",
+    "Notice that the line that connects the points\n",
```
hmgaudecker, line 393: Change on two accounts: New graph with interpolation, no mention of matplotlib.

explanations/function_evaluator.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from functools import partial\n",
+    "\n",
+    "from lcm.function_evaluator import get_function_evaluator\n",
+    "\n",
+    "_value_function = get_function_evaluator(\n",
```
hmgaudecker, line 375: Let's try to brainstorm for whether we can find more expressive names here. I find it hard to wrap my head around this cell.

explanations/function_evaluator.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   "metadata": {},
+   "source": [
+    "## Technical Details\n",
+    "\n",
+    "In the following, we will discuss the underlying functions that are used to implement\n",
+    "the function evaluator.\n",
+    "\n",
+    "### Label Translator\n",
+    "\n",
+    "The label translator is used to map the labels of dense discrete grids to their\n",
```
hmgaudecker, line 463: But this will stay, no? So why discuss like this? Shouldn't it be:

> *(Once issue [#82](https://github.com/OpenSourceEconomics/lcm/issues/82) is tackled, )* LCM converts discrete grids into an internal grid that is directly usable as an index. Thus, the label translator simply is the identity function.

explanations/function_evaluator.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Re-implementation of the function evaluator given the example model\n",
+    "\n",
+    "### The Steps\n",
+    "\n",
+    "In this section, we will re-implement the function evaluator for the example model\n",
```
hmgaudecker, line 809: Would be good to explain the purpose!

src/lcm/function_evaluator.py
```diff
@@ -28,21 +29,21 @@ def get_function_evaluator(
...
-    pre-calculated values of the function.
-    2. The original arguments of the function that was pre-calculated on the grid.
+       1. Helper arguments such as information about the grid, indexer arrays and the
+          pre-calculated values of the function.
+       2. The original arguments of the function that was pre-calculated on the grid.
 
-    After partialling in all helper arguments, the resulting function will completely
-    hide that it is not analytical and feel like a normal function. In particular, it
-    can be jitted, differentiated and vmapped with jax.
+    After partialling in all helper arguments, the resulting function hides that it is
```
hmgaudecker, line 36:
```suggestion
    After partialling in all helper arguments, the resulting function behaves like
```

src/lcm/function_evaluator.py
```diff
@@ -28,21 +29,21 @@ def get_function_evaluator(
...
+    not an analytical function. In particular, it can be jitted, differentiated and
```
hmgaudecker, line 37:
```suggestion
    an analytical function. In particular, it can be jitted, differentiated and
```

src/lcm/function_evaluator.py
```diff
@@ -133,70 +131,48 @@ def get_function_evaluator(
...
     """Create a function that translates a label into a position in a list of labels.
 
+    Currently, only labels are supported that are themselves indices. The label
+    translator in this case is thus just the identity function.
+
     Args:
-        labels (list, np.ndarray, jnp.array): List of allowed labels.
-        in_name (str): Name of the variable that provides the label in the signature
-            of the resulting function.
+        in_name: Name of the variable that provides the label in the signature of trche
```
hmgaudecker, line 155:
```suggestion
        in_name: Name of the variable that provides the label in the signature of the
```

hmgaudecker (commented):
Looks great! There may be room for language improvement, but mainly about docs. I already changed a little bit.

explanations/function_evaluator.ipynb
timmens: How about `scalar_function_evaluator`?

src/lcm/function_representation.py
```diff
@@ -253,24 +229,33 @@ def find_coordinate(*args, **kwargs):
     return find_coordinate
 
 
-def _get_interpolator(data_name, axis_names, map_coordinates_kwargs=None):
+DefaultMapCoordinatesOptions: MapCoordinatesOptions = {
```
hmgaudecker, line 232: It always scares me to set default arguments deep in the code base. Especially the case here, since it is not innocuous at all (concerns what happens outside the grid, right?). Is there any reasonable place to do it close to the user-facing interface? Else, I would actually be in favour of requiring it as user input.

hmgaudecker (approved):
Excellent, thanks!

PR #85, opened 2024-06-30
Update model processing

### LINKED ISSUE #82: ENH: Support arbitrary discrete integer grids
#### Problem

Discrete grids need to coincide with their indices, i.e., `valid_grid = [0, 1, 2, 3]` and `invalid_grid = [2, 3]`.

#### What we want

Support arbitrary **integer** grids. I.e., the following grids should be supported:
- `[2, 3]`
- `[-1, 0, 1]`
- `[1, 10, 100]`
- `[1, 2, 1_000_000]`

#### Before

Before #81, we tried to solve this in `lcm.function_evaluator.get_label_translator()`, using: 
```python
val_to_pos = dict(zip(_grid, range(len(labels)), strict=True))

@with_signature(args=[in_name])
def translate_label(*args, **kwargs):
    kwargs = all_as_kwargs(args, kwargs, arg_names=[in_name])
    return val_to_pos[kwargs[in_name]]
```

The problem was, however, that the lookup `val_to_pos[x]` is not vmappable/jittable in JAX (this might have been different in an earlier JAX version). The underlying problem here is that (1) JAX does not know scalar values but only 0-dimensional arrays, but `val_to_pos` has int keys (i.e. `0` vs. `jax.array(0)`); and (2) we cannot use `jax.array(x)` as a dictionary key, because JAX arrays are not hashable.

#### Bad solution

Add a new label translator that is JAX compatible:

```python
labels_array = jnp.array(labels)
@with_signature(args=[in_name])
def translate_label(*args, **kwargs):
    kwargs = all_as_kwargs(args, kwargs, arg_names=[in_name])
    return jnp.nonzero(labels_array == kwargs[in_name], size=1, fill_value=None)[0][0]
```

The problem here is that this function is much slower (10-100 times) for arbitrary grids compared to index grids. The computational cost also grows linearly with the size of the label arrays.

#### Better solution

The idea is to solve this problem during the model processing and not in the function evaluator. As an example, consider a health state variables that attains values `[-1, 0, 1]`. In the model specification the user will write

```python
...
"states": {
   "health": {"options": [-1, 0, 1]},
   ...
}
...
```

In the model processing, we would not add `health` as a state variable, but `__health_index__`, which would live on the grid `[0, 1, 2]`. Since the user wrote functions that depend on `health` and not on the index, we also need to add a dynamically defined function

```python
def health(__health_index__):
   return health_options[__health_index__]  # health_options must be a jax array
```

Given this, arbitrary discrete integer grids should be supported.

timmens:
In this PR, I update the model processing. I mainly focus on the following things:

- [x] Adding type hints to `process_model.py`
- [x] Improving the user interface through user `Model` dataclass which adds autocompletion and error handling at the creation of the user model
- [x] Improving the user interface through grid dataclass that add autocompletion and error handling at the grid creation time.
   - [x] Implement `DiscreteGrid`, `LinspaceGrid`, `LogspaceGrid`, which inherit from baseclass `Grid`, and `ContinuousGrid`. 
- [x] Update `create_params_template.py`
- [x] Remove `black` formatter and replace by `ruff-formatter` (this also means we can remove nbqa)


src/lcm/exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+    """Raised when there is an error in the grid initialization."""
+
+    def __init__(self, error_message) -> None:
+        """Initialize the exception with a list of errors or a single error message."""
+        if isinstance(error_message, list):
+            error_message = format_errors(error_message)
+        super().__init__(error_message)
+
+
+def format_errors(errors: list[str]) -> str:
```
timmens, line 21: Should also work with `str`.

src/lcm/create_params_template.py
```diff
@@ -1,34 +1,31 @@
...
 
 def create_params_template(
-    model_spec: dict[str, Any],
+    user_model: Model,
     variable_info: pd.DataFrame,
     grids: dict[str, Array],
-    default_params: dict[str, float] | None = None,
-) -> dict[str, dict[str, Any] | float]:
+    default_params: dict[str, int | float] | None = None,
+) -> Params:
```
janosg, line 18: This makes me think that some params Object comes out here; I would prefer to see `PyTree` even if that is less precise.

src/lcm/grid_helpers.py
```diff
@@ -0,0 +1,111 @@
...
+
+    """
+    return jnp.linspace(start, stop, n_points)
+
+
+def get_linspace_coordinate(
+    value: Scalar,
+    start: Scalar,
+    stop: Scalar,
+    n_points: int,
```
janosg, line 47: I think it would make sense to have `get_..._coordinate` and the creation function as methods of the grid object. 

If not, I would at least replace these three arguments by a `grid` or `gridspec` argument. 

(Implementation can of course stay in these functions but they would become private)

src/lcm/grids.py
```diff
@@ -1,111 +1,259 @@
...
+
+        if errors := _validate_discrete_grid(self.options):
+            msg = format_messages(errors)
+            raise GridInitializationError(msg)
+
+    def to_jax(self) -> jnp.ndarray:
+        """Convert the grid to a Jax array."""
+        return jnp.array(list(self.options))
+
+    def replace(self, options: Collection[int | float]) -> "DiscreteGrid":
```
janosg, line 57: Why is the `replace` method ever needed? You can just as easily create a new one.

src/lcm/grids.py
```diff
@@ -1,111 +1,259 @@
...
 
-import jax.numpy as jnp
-from jax import Array
 
-from lcm.typing import Scalar
+@dataclass(frozen=True)
+class ContinuousGrid(Grid):
+    """LCM Continuous Grid base class."""
 
+    kind: ContinuousGridType = field(init=False, default=None)  # type: ignore[arg-type]
```
janosg, line 74: Not sure if we discussed this but in optimagic constraints we got rid of a `kind/type` field by having several subclasses of `Constraint`, i.e. by encoding the `kind` in the type of the constraint.

src/lcm/model.py
```diff
@@ -0,0 +1,120 @@
...
+
+    """
+
+    description: str | None = None
+    _: KW_ONLY
+    n_periods: int
+    functions: dict[str, Callable] = field(default_factory=dict)
+    choices: dict[str, Grid] = field(default_factory=dict)
+    states: dict[str, Grid] = field(default_factory=dict)
+    _skip_checks: InitVar[bool] = False
```
janosg, line 32: In which situation would you want to skip checks?

src/lcm/model.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __post_init__(self, _skip_checks: bool) -> None:
+        if _skip_checks:
+            return
+
+        if type_errors := _validate_attribute_types(self):
```
janosg, line 38: Not a fan of the walrus operator here. Sure, it saves a line by it is already unexpected that a `_valiadate_...` function returns anything so I would rather learn about that fact in a separate line and then read the if condition. https://hackernoon.com/why-senior-devs-write-dumb-code-and-how-to-spot-a-junior-from-a-mile-away-27fa263b101a

src/lcm/create_params_template.py
timmens: `ParamsDict`

src/lcm/model.py
timmens: Write custom testing model

src/lcm/grid_helpers.py
janosg: With this we could get to a point where a user could add a new grid type without changing LCM source code.

src/lcm/grids.py
timmens: Not needed anymore with new `get_coordinate` method

src/lcm/model.py
hmgaudecker, line 1: Why is this not in `lcm.interfaces`? If anything, this seems more of a user-facing interface than `InternalModel`... Also, would not `UserModel` be a more appropriate name?

src/lcm/model.py
```diff
@@ -0,0 +1,122 @@
...
+            **kwargs: Keyword arguments to replace the attributes of the model.
+
+        Returns:
+            A new model with the replaced attributes.
+
+        """
+        return dc.replace(self, **kwargs)
+
+
+def _validate_attribute_types(model: Model) -> list[str]:
```
hmgaudecker, line 59: Isn't the code from here on mainly duplicating pydantic functionality? So even if we can't use that everywhere, could we do so for a subset and reduce own code for a stable dependency?

src/lcm/simulate.py
hmgaudecker, line 1: I would suggest making `initial_states` the second argument because the user will typically need to supply that, right?

hmgaudecker (commented):
Thanks a lot, that is much nicer to read now! 

This clarity reveals some parts that could use high-level improvements -- ofc I did not follow instructions in that respect and noted them. Feel free to push off to some issue wherever you think that is more useful, though! Or, of course, tell me that this or that comment is nonsense.

src/lcm/process_model.py
```diff
@@ -24,99 +29,123 @@ def process_model(user_model):
...
+        user_model: The model as provided by the user.
 
     Returns:
-        lcm.interfaces.Model: The processed model.
+        The processed model.
 
     """
-    _function_info = _get_function_info(user_model)
-    _variable_info = _get_variable_info(
+    function_info = _get_function_info(user_model)
```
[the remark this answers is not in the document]
timmens: That is a good point, but it requires changes to multiple files. I've added it to a to-do list, but will not incorporate it in this PR.

src/lcm/entry_point.py
```diff
@@ -21,7 +22,7 @@
 
 
 def get_lcm_function(
```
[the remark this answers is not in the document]
timmens: I made the type hint changes and updated the docstring. I won't go much further, because the entry point will be completely rewritten in the next 1-2 months.

src/lcm/model.py
timmens: In `lcm.interfaces` we try to collect all abstractions that are required for LCM _internally_, except for the user-facing abstractions, as they require input validation. If you can think of a better name for that module, we can rename it.

I think we should stick to `Model` instead of `UserModel`. For the user, it will be clear that this is a custom model. And I think the following is much nicer to write:

```python
model = Model(
   ...
   choices = {
      "exercise": DiscreteGrid([0, 1])
   },
   ...
)
```

than

```python
model = UserModel(
   ...
   choices = {
      "exercise": DiscreteGrid([0, 1])
   },
   ...
)
```

On top, we would need the same discussion for the grids, right?

src/lcm/model.py
timmens: Yes, we could do that, good point. I've added it to the to-do list.

src/lcm/interfaces.py
```diff
@@ -127,17 +93,17 @@ class Model(NamedTuple):
...
     variable_info: pd.DataFrame
-    functions: dict
+    functions: dict[str, Callable]
     function_info: pd.DataFrame
-    params: dict
-    # not really processed yet
-    shocks: dict
+    params: ParamsDict
     n_periods: int
+    # Not properly processed yet
```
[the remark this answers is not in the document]
timmens: I agree that we should rename to `random_utility_shocks`. For the other points, I would wait with fully deleting it. This is just the internal interface, and we are trying to add the Extreme Value utility shocks soon-ish. The user-facing interface, is another story, and I'd suggest we tackle that once we have integrated the feature.

src/lcm/model.py
hmgaudecker: For `Model`, I would argue that since the keyword is `user_model` everywhere (also user-facing, right?), we should be consistent in order to minimise confusion :wink:.

For grids, I am not sure why we should need a distinction at all. Parse whatever the user gives us right way, no need to make a distinction between internal and user-facing versions? I can see that for `InternalModel` since function signatures are different etc.

timmens: Currently, the only user-facing function is `get_lcm_function`, where the argument is called `model`. Every function that uses `user_model` is, at least for now, internal. Unless you have a strong opinion, I'd leave it as it is for now and revisit the discussion once we have refactored the entry point.

Valid point for the grids, though!

hmgaudecker: Fine to wait. But I'd be pretty sure that whatever the outcome of the refactoring, we should not have 3 names for 2 things! Users will be able to handle `UserModel` / `user_model`.

janosg: Then the class should probably be called `Model` instead of `UserModel`.

hmgaudecker: Yeah, I guess if a user comes across `InternalModel` somewhere he will back off quickly.  And with type annotations, the distinction will also be clear enough for developers!

src/lcm/function_representation.py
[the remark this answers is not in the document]
timmens: Will most likely create a subfolder for state-choice-space-related functions; we will move `function_representation` there.

hmgaudecker (approved):
Excellent work, many thanks!

PR #90, opened 2024-09-15
Allow general discrete grids

timmens:
In this PR, I add the functionality to allow general discrete grids.

Previously, only index grids were allowed. Here we add an updating step, which takes a model containing an arbitrary discrete grid, overwrites the grid with a corresponding index grid, and adds a function to the model that transforms the index values to the correct label. This means that the user can now use arbitrary grids, while internally, we can still use the convenient indexer grids.

### ToDO

- [x] Implement model updating and add first tests
- [ ] Refactor code in `process_model.py` into single model-updating call


src/lcm/input_processing/process_model.py
```diff
@@ -36,18 +39,21 @@ def process_model(model: Model) -> InternalModel:
         The processed model.
 
     """
-    params = create_params_template(model)
+    new_model, converter = convert_discrete_options_to_indices(model)
```
hmgaudecker, line 42:
```suggestion
    tmp_model, converter = convert_discrete_options_to_indices(model)
```

src/lcm/simulate.py
```diff
@@ -65,6 +69,8 @@ def simulate(
             )
         vf_arr_list = solve_model(params)
 
+    params = converter.params_to_internal(params)
```
hmgaudecker, line 72:
```suggestion
    internal_params = converter.params_to_internal(params)
```
Let's not overwrite inputs / have variables contain different but similar things at different points in a function.

src/lcm/solve_brute.py
```diff
@@ -42,12 +43,16 @@ def solve(
...
             expected maximum over all discrete choices of a given state.
+        converter (Converter): Converter for states and parameters between external and
+            internal representation.
         logger (logging.Logger): Logger that logs to stdout.
 
     Returns:
         list: List with one value function array per period.
 
     """
+    params = converter.params_to_internal(params)
```
hmgaudecker, line 54: See above.

tests/input_processing/test_discrete_state_conversion.py
hmgaudecker, line 1: Add a test with a grid `[1, 0]`.

hmgaudecker (commented):
Great start, path forward as discussed in person!

timmens:
Should provide only the following option to create discrete grids:

```python
@dataclass
class RetirementStatus:
    working: int = 0
    retired: int = 1
    
DiscreteGrid(RetirementStatus)
```

src/lcm/grids.py
```diff
@@ -30,13 +30,13 @@ class DiscreteGrid(Grid):
 
     """
 
-    options: Collection[int | float]
+    options: type
```
hmgaudecker, line 33: Just a note that this should be documented really well, if just for my own benefit. I realised only now and after a couple of loops in my brain that we are talking about passing the type and not an instance!

src/lcm/grids.py
```diff
@@ -30,13 +30,13 @@ class DiscreteGrid(Grid):
...
 
     def __post_init__(self) -> None:
-        if not isinstance(self.options, Collection):
+        if not is_dataclass(self.options):
             raise GridInitializationError(
-                "options must be a collection of scalar int or float values, e.g., a ",
-                "list or tuple",
+                "options must be a dataclass with scalar int or float fields, but is "
```
hmgaudecker, line 38: We can be more precise in the description here now. However, I am not sure why we want to distinguish between this error and what is thrown by `_validate_discrete_grid` ?

src/lcm/grids.py
```diff
@@ -46,7 +46,33 @@ def __post_init__(self) -> None:
 
     def to_jax(self) -> Array:
         """Convert the grid to a Jax array."""
-        return jnp.array(list(self.options))
+        return jnp.array(_get_fields(self.options))
```
hmgaudecker, line 49: This seems too late to throw an error, why not do that during validation / use the same code?

src/lcm/grids.py
timmens: In `_validate_discrete_grid`, I want to assume that I am working with a dataclass. Makes collection of error messages much easier. Be sure to checkout the updated version.

examples/long_running.py
```diff
@@ -9,6 +11,15 @@
 # ======================================================================================
 
 
+# --------------------------------------------------------------------------------------
+# Categorical variables
+# --------------------------------------------------------------------------------------
+@dataclass
+class WorkingState:
```
hmgaudecker, line 18:
```suggestion
class Working:
```
Both a state and a choice, so avoid `State` for sure.

explanations/function_representation.ipynb
```diff
@@ -76,11 +76,19 @@
...
    "source": [
+    "from dataclasses import dataclass\n",
+    "\n",
     "import jax.numpy as jnp\n",
     "\n",
     "from lcm import DiscreteGrid, LinspaceGrid, Model\n",
     "\n",
     "\n",
+    "@dataclass\n",
+    "class RetirementStatus:\n",
```
hmgaudecker, line 87:
```suggestion
    "class Working:\n",
```
Just to be consistent with other example

src/lcm/entry_point.py
```diff
@@ -170,6 +170,7 @@ def get_lcm_function(
         continuous_choice_grids=continuous_choice_grids,
         compute_ccv_functions=compute_ccv_functions,
         emax_calculators=emax_calculators,
+        converter=_mod.converter,
```
hmgaudecker, line 173:
```suggestion
        discrete_grid_converter=_mod.discrete_grid_converter,
```

src/lcm/grids.py
```diff
@@ -16,37 +16,65 @@ class Grid(ABC):
...
 
-        errors = _validate_discrete_grid(self.options)
+        names_and_values = _get_field_names_and_values(category_class)
+
+        errors = _validate_discrete_grid(names_and_values)
         if errors:
             msg = format_messages(errors)
             raise GridInitializationError(msg)
 
+        self.__categories = list(names_and_values.keys())
```
hmgaudecker, line 62:
```suggestion
        self.__categories = tuple(names_and_values.keys())
```

src/lcm/grids.py
```diff
@@ -16,37 +16,65 @@ class Grid(ABC):
...
+        self.__codes = list(names_and_values.values())
```
hmgaudecker, line 63:
```suggestion
        self.__codes = tuple(names_and_values.values())
```

src/lcm/grids.py
```diff
@@ -16,37 +16,65 @@ class Grid(ABC):
...
+
+    @property
+    def categories(self) -> list[str]:
```
hmgaudecker, line 66:
```suggestion
    def categories(self) -> tuple[str]:
```

src/lcm/grids.py
```diff
@@ -16,37 +16,65 @@ class Grid(ABC):
...
+        """Get the list of category names."""
+        return self.__categories
+
+    @property
+    def codes(self) -> list[int | float]:
```
hmgaudecker, line 71:
```suggestion
    def codes(self) -> tuple[int | float]:
```

src/lcm/grids.py
```diff
@@ -131,35 +159,62 @@ def get_coordinate(self, value: Scalar) -> Scalar:
...
 
     """
     error_messages = []
 
-    if not len(options) > 0:
-        error_messages.append("options must contain at least one element")
-
-    if not all(isinstance(option, int | float) for option in options):
-        error_messages.append("options must contain only scalar int or float values")
+    if not len(names_and_values) > 0:
```
hmgaudecker, line 175:
```suggestion
    if len(names_and_values) == 0:
```

src/lcm/input_processing/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .discrete_state_conversion import DiscreteStateConverter
 from .process_model import process_model
 
-__all__ = ["process_model"]
+__all__ = ["process_model", "DiscreteStateConverter"]
```
hmgaudecker, line 4:
```suggestion
__all__ = ["process_model", "DiscreteGridConverter"]
```

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+
+        """
+        out = states.copy()
+        for var in self.converted_states:
+            out.pop(var)
+            out[f"__{var}_index__"] = self.code_to_index[var](states[var])
+        return out
+
+
+def convert_discrete_codes_to_indices(
```
hmgaudecker, line 93:
```suggestion
def convert_arbitrary_codes_to_array_indices(
```

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+from lcm.grids import DiscreteGrid
+from lcm.input_processing.util import (
+    get_gridspecs,
+)
+from lcm.typing import ParamsDict
+from lcm.user_model import Model
+
+
+@dataclass
+class DiscreteStateConverter:
```
hmgaudecker, line 18:
```suggestion
class DiscreteGridConverter:
```

- Adjust docstring, making clearer what kinds of stuff is affected here. 
- Add test case with simulation results, these should have the user-supplied codes also for choices.

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+    Attributes:
+        converted_states: The names of the states that have been converted.
+        index_to_code: A dictionary of functions mapping from the internal index to the
+            code for each converted state.
+        code_to_index: A dictionary of functions mapping from the code to the internal
+            index for each converted state.
+
+    """
+
+    converted_states: list[str] = field(default_factory=list)
```
hmgaudecker, line 34:
```suggestion
    converted_grid_names: list[str] = field(default_factory=list)
```

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+@dataclass
+class DiscreteStateConverter:
+    """Converts between representations of discrete states and their parameters.
+
+    While LCM supports general discrete grids, internally, these are converted to
+    indices. This class provides functionality for converting between the internal
+    representation and the external representation.
+
+    Attributes:
+        converted_states: The names of the states that have been converted.
```
hmgaudecker, line 26:
```suggestion
        converted_grid_names: The names of the states that have been converted.
```
In the subsequent two descriptions, make clear that the dict keys are these names. Do we need `converted_grid_names` at all? Seems like there is just scope for confusion.

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+    def states_to_internal(self, states: dict[str, Array]) -> dict[str, Array]:
+        """Convert states from external to internal representation.
+
+        If a state has been converted, the name of its corresponding index function must
+        be changed from `{var}` to `___{var}_index__`, and the values of the state must
+        be converted from codes to indices.
+
+        """
+        out = states.copy()
+        for var in self.converted_states:
```
hmgaudecker, line 87:
```suggestion
        for var, conv in self.code_to_index.items():
```

src/lcm/input_processing/discrete_state_conversion.py
```diff
@@ -0,0 +1,244 @@
...
+            out.pop(var)
+            out[f"__{var}_index__"] = self.code_to_index[var](states[var])
+        return out
+
+
+def convert_discrete_codes_to_indices(
+    model: Model,
+) -> tuple[Model, DiscreteStateConverter]:
+    """Update the user model to ensure that discrete variables have index codes.
```
hmgaudecker, line 96: Adjust language a little.

hmgaudecker (commented):
As discussed, really nice!

hmgaudecker (commented):
Just had a brief look -- very nice! Need to stop now.

Apologies for yet another round of renaming suggestions, but looking at it I wasn't certain that a newcomer would immediately understand what is expected where.

src/lcm/input_processing/discrete_grid_conversion.py
```diff
@@ -14,27 +16,34 @@
...
     index_to_code: dict[str, Callable[[Array], Array]] = field(default_factory=dict)
     code_to_index: dict[str, Callable[[Array], Array]] = field(default_factory=dict)
 
+    def __post_init__(self) -> None:
+        if set(self.index_to_code.keys()) != set(self.code_to_index.keys()):
+            raise ValueError(
+                "The keys of index_to_code and code_to_index must be the same."
+            )
+
     def internal_to_params(self, params: ParamsDict) -> ParamsDict:
```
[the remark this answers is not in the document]
timmens: Because of the minor changes to params we do not want to introduce another concept here; as discussed.

src/lcm/input_processing/discrete_grid_conversion.py
```diff
@@ -56,51 +66,57 @@ def params_to_internal(self, params: ParamsDict) -> ParamsDict:
...
-            out.pop(f"next_{var}")
-            out[f"next___{var}_index__"] = params[f"next_{var}"]
+        for var in self.index_to_code:
+            old_name = f"next_{var}"
+            if old_name in out:
+                out[f"next___{var}_index__"] = out.pop(old_name)
         return out
 
-    def internal_to_states(self, states: dict[str, Array]) -> dict[str, Array]:
+    def internal_to_discrete_vars(
```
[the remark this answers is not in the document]
timmens: Would prefer `internal_vars_to_vars` and `vars_to_internal_vars`, because with `internal_vars_to_discrete_vars` I get the impression that all internal variables are converted to discrete variables.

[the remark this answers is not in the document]
hmgaudecker: Great.

src/lcm/grids.py
```diff
@@ -16,37 +16,56 @@ class Grid(ABC):
...
+        codes: The list of category codes.
+
+    Raises:
+        GridInitializationError: If the `category_class` is not a dataclass with scalar
+            int or float fields.
 
     """
 
-    options: Collection[int | float]
+    def __init__(self, category_class: type) -> None:
```
ChristianZimpelmann, line 40: why `type`? I would have expected `dataclass`.

ChristianZimpelmann (approved):
Very nice! I really see how the OOP features are beneficial. 

Just a few questions or comments.

src/lcm/grids.py
```diff
@@ -16,37 +16,56 @@ class Grid(ABC):
...
-        if errors:
-            msg = format_messages(errors)
-            raise GridInitializationError(msg)
+        self.__categories = tuple(names_and_values.keys())
+        self.__codes = tuple(names_and_values.values())
+
+    @property
+    def categories(self) -> tuple[str, ...]:
+        """Get the list of category names."""
+        return self.__categories
```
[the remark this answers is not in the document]
timmens: I want to hide the attribute holding the actual data from the user. This can be achieved, at least partly, by using `__categories`.

src/lcm/input_processing/process_model.py
```diff
@@ -36,18 +39,21 @@ def process_model(model: Model) -> InternalModel:
         The processed model.
 
     """
-    params = create_params_template(model)
+    tmp_model, converter = convert_arbitrary_codes_to_array_indices(model)
```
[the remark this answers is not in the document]
timmens: `tmp_model` is not ideal, but that is okay for now.

src/lcm/grids.py
timmens: Unfortunately `dataclass` is not a type. And as we input the class itself and not an instance of the class, we cannot assume more than `type`.

```python
@dataclass
class Health:
   pass

DiscreteGrid(Health)  # here Health is the class and an instance
```

tests/conftest.py
```diff
@@ -1,5 +1,23 @@
...
 def pytest_sessionstart(session):  # noqa: ARG001
     config.update("jax_enable_x64", val=True)
+
+
+def _category_class_factory(values: list[int]):
+    init = [(f"cat{i}", int, value) for i, value in enumerate(values)]
+    return make_dataclass("CategoryClass", init)
+
+
+@pytest.fixture(scope="session")
```
hmgaudecker, line 16: Just to check whether you really need the two functions here? I.e., with and without underscore? Can't we just decorate the function itself or does that lead to issues because it takes an argument?

hmgaudecker (approved):
Excellent, thanks!

tests/conftest.py
timmens: Yes, the argument becomes a problem. While pytest supports fixtures that have arguments, it seems like you have to parametrize it via the parametrize decorator, and that would not work in our case, because we need to use different versions of it in the same test.

PR #92, opened 2024-09-23
Restrict discrete grid codes to array indices

timmens:
In this PR, 

**I remove the ability to specify general discrete grid codes.** While it may be beneficial in the future, we currently do not have a use case, while the feature makes maintaining the library more complex. We have, therefore, decided to drop the feature for now. If it turns out to be needed in the future, we can easily add it by following this PR.

Additionally, I renamed the discrete deterministic test model from `ISKHAKOV_ET_AL_2017_FULLY_DISCRETE` to `ISKHAKOV_ET_AL_2017_DISCRETE` since the state is no longer discrete, only the choices.

### What changed?

This was allowed before merging of this PR:

```python
@dataclass
class Health:
   bad: int = -1
   neutral: int = 0
   good: int = 1
   
DiscreteGrid(Health)
```

Now, this would raise an error, and you have to specify the grid like so:

```python
@dataclass
class Health:
   bad: int = 0
   neutral: int = 1
   good: int = 2
   
DiscreteGrid(Health)
```


tests/test_entry_point.py
```diff
@@ -20,7 +20,7 @@
 
 STRIPPED_DOWN_AND_FULLY_DISCRETE_MODELS = [
```
hmgaudecker, line 21:
```suggestion
STRIPPED_DOWN_AND_DISCRETE_MODELS = [
```

But see below before changing anything.

tests/test_models/deterministic.py
```diff
@@ -171,35 +171,34 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
...
 @dataclass
-class DiscreteConsumptionStatus:
-    low: int = 1
-    high: int = 2
-
-
-DiscreteWealthStatus = make_dataclass(
-    "DiscreteWealthStatus", [(f"level_{w}", int, w) for w in range(1, 401)]
-)
+class ConsumptionStatus:
```
hmgaudecker, line 174:
```suggestion
class ConsumptionChoice:
```
Status sounds really weird here. Apologies for not noting before. I think `Working` is a special case in the sense that a choice turning directly into a state can be better seen as a `Status`. 

More importantly: Define this at the top of the module, it is already being used in line 54.

tests/test_models/deterministic.py
```diff
@@ -171,35 +171,34 @@ def absorbing_retirement_filter(retirement, lagged_retirement):
 
 
 @dataclass
-class DiscreteConsumptionStatus:
-    low: int = 1
-    high: int = 2
-
-
-DiscreteWealthStatus = make_dataclass(
```
hmgaudecker, line 179: What do we gain and loose by using a continuous grid? That is, an alternative would be to just use, say, five discrete grid points and keep the example *fully* discrete?

hmgaudecker (commented):
Very nice -- much easier to read!!!

tests/test_models/deterministic.py
timmens: Went _fully_ discrete.

tests/test_models/deterministic.py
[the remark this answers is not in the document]
timmens: I think I am against this. The `deterministic.py` test module contains the Iskhakov et al. (2017) model and derivatives of that model. These models can share functions and a namespace. Moving only the discrete model to a separate module would confuse me more than having it in there. On top, I can have one `get_model_config` and `get_params` function, that I use in the tests.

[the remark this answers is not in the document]
hmgaudecker: I just think the fully discrete version is enough (and different enough from the original) to not mix them too much. Also, the definition of the discrete stuff and the place where it is used now are 90-150 lines apart from each other.

How about having:
```
tests/test_models/deterministic.py
tests/test_models/discrete_deterministic.py
```
and the latter importing almost all relevant stuff from the former?

hmgaudecker (approved):

tests/input_processing/test_process_model.py
```diff
@@ -18,7 +18,7 @@
     process_model,
 )
 from lcm.mark import StochasticInfo
-from tests.test_models.deterministic import (
+from tests.test_models import (
```
hmgaudecker, line 21:
```suggestion
from tests.test_models import get_model_config
```

tests/test_entry_point.py
```diff
@@ -10,17 +10,21 @@
 from lcm.input_processing import process_model
 from lcm.model_functions import get_utility_and_feasibility_function
 from lcm.state_space import create_state_choice_space
-from tests.test_models.deterministic import get_model_config
+from tests.test_models import get_model_config
+from tests.test_models.deterministic import (
```
hmgaudecker, line 14:
```suggestion
from tests.test_models.deterministic import RetirementStatus
```

hmgaudecker (approved):
Even  nicer!

PR #93, opened 2024-09-24
Switch to Apache 2.0 License

timmens:
We are planning to include Apache 2.0-licensed code from JAX in LCM. To avoid having two licenses, we will switch from the MIT license to the Apache 2.0 license.

Different licenses are compared [here](https://choosealicense.com/licenses/). The MIT and Apache 2.0 licenses have similar implications for entities that want to re-use LCM's code.


hmgaudecker (approved):
Perfect, thanks!

PR #94, opened 2024-09-24
Implement linear extrapolation

### LINKED ISSUE #83: ENH: Add linear extrapolation to `map_coordinates`
### Why

Linear extrapolation around the boundaries is useful for various reasons. For example, (1) it can robustify the solution if, in a state, transition values outside the original grid are requested, or (2) it helps when simulating stochastic where node points are outside the grid.

### When

I have asked a question in [JAX's discussion forum](https://github.com/google/jax/discussions/22139). If they decide not to add this feature, we can simply implement a custom `map_coordinates` function for `LCM.`

### Implementation

When using `map_coordinates` to linearly interpolate between values, internally the following function is called:

```python
def _linear_indices_and_weights(coordinate: Array) -> list[tuple[Array, ArrayLike]]:
  lower = jnp.floor(coordinate)
  upper_weight = coordinate - lower
  lower_weight = 1 - upper_weight
  index = lower.astype(jnp.int32)
  return [(index, lower_weight), (index + 1, upper_weight)]
```

To allow for linear extrapolation for coordinate values that are either below `0`, or larger than the original grid size `input_size`, one only has to change the first line to:

```python
...
lower = jnp.clip(jnp.floor(coordinate), min=0, max=input_size - 2)
...
```

timmens:
In this PR, I add capabilities to perform linear extrapolation.

In particular, I

- Implement a custom `map_coordinates` function that allows for linear extrapolation. (A [discussion](https://github.com/jax-ml/jax/discussions/22139#discussion-6869226) on the JAX repo implied that they would not integrate this feature)
- Remove interpolation options. Since we only want to perform linear inter- and extrapolation, there is no need for added complexity.
- Update the regression tests since the results are different. Notably, the results now look better (more correct) than before. This also shows that the linear extrapolation helps with the solution and simulation.

Closes #83

### How it works

To extrapolate linearly, one has to adjust the function that computes weights and indices of the coordinate points. For the linear interpolation, the JAX code can be found [here](https://github.com/jax-ml/jax/blob/1fe0c5dad51cdb7acb8a8420df3d758be3a01bad/jax/_src/scipy/ndimage.py#L61C1-L66C60).

This implementation implicitly assumes that `coordinate` is in the domain (interpolation). For values outside of the domain (extrapolation), one has to adjust the first line to

```python
lower = jnp.clip(jnp.floor(coordinate), 0, input_size - 2)
```

where `input_size` corresponds to a dimension of the input shape. This leads to the same behavior for interpolation but assigns the correct indices and weights needed for linear extrapolation for coordinates outside the domain.


hmgaudecker (commented):
Very cool. Had a brief look and leaving the note so I can see down the road what has changed from this point onwards.

tests/test_ndimage.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@pytest.mark.parametrize("map_coordinates", JAX_IMPLEMENTATIONS)
+@pytest.mark.parametrize("shape", TEST_SHAPES)
+@pytest.mark.parametrize("coordinates_shape", TEST_COORDINATES_SHAPES)
+@pytest.mark.parametrize("dtype", [np.int64, np.float64])
+def test_map_coordinates_against_scipy(
+    map_coordinates, shape, coordinates_shape, dtype
+):
+    """Test that all libraries implement same behavior with integer input."""
```
timmens, line 61: Update docstring

tests/test_ndimage.py
```diff
@@ -0,0 +1,126 @@
...
+    # Gradient of f(step) is 2 * step
+    assert_allclose(jax.grad(f)(0.5), 1.0)
+    assert_allclose(jax.grad(f)(1.0), 2.0)
+
+
+def test_extrapolation():
+    x = jnp.arange(3.0)
+    c = [jnp.array([-2.0, -1.0, 5.0, 10.0])]
+
+    got = lcm.ndimage.map_coordinates(x, c)
```
timmens, line 123: `lcm_map_coordinates`

tests/test_ndimage.py
```diff
@@ -0,0 +1,126 @@
...
+    c_jax = [jnp.asarray(c_i) for c_i in c]
+
+    expected = scipy_map_coordinates(x, c)
+    got = map_coordinates(x_jax, c_jax)
+
+    assert_array_almost_equal(got, expected, decimal=14)
+
+
+@pytest.mark.parametrize("map_coordinates", JAX_IMPLEMENTATIONS)
+def test_map_coordinates_round_half_integer_input(map_coordinates):
```
timmens, line 74: Just combine with below test and add another parametrization. Usually we try to avoid this, but the test body is complex, so repeating it is worse than additionally looping over dtypes.

src/lcm/ndimage.py
timmens, line 1: Add docstrings and some helpful comments to auxiliary functions.

src/lcm/ndimage.py
```diff
@@ -0,0 +1,94 @@
...
+    Args:
+      input: N-dimensional input array from which values are interpolated.
+      coordinates: length-N sequence of arrays specifying the coordinates
+        at which to evaluate the interpolated values
+
+    Returns:
+      The interpolated (extrapolated) values at the specified coordinates.
+
+    """
+    if len(coordinates) != input.ndim:
```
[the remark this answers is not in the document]
timmens: I don't know if this will be optimized away, but the computation is tiny compared to the rest of the function. Also, this condition is JAX-friendly since it does not rely on specific values of the inputs, i.e., the JAX tracers can pass through it.

When experimenting with this function, I got this error multiple times, meaning it's easy to pass the coordinates in the wrong format. Therefore, I believe it's best to keep it.

PR #98, opened 2024-10-23
Bump the Version of Jax and Jaxlib to 0.4.34

mj023:
This PR changes the dependencies for the pixi environments to require JAX>= 0.4.34 and jaxlib >= 0.4.34. The reason for this is, that #77 created memory allocation issues when using older JAX and jaxlib versions. For larger models this made them impossible to solve with limited GPU memory.

**Problem description**
With #77 the solve_model() function returned by LCM is jitted by default. After this change, one would run into problems when solving a model on the GPU. Jax would throw an error, because the program tried to save huge arrays, with the same dimensions as the the whole state-choice-space, into the GPU memory. For larger models these arrays could be multiple TB big. In the past high memory usage has never been a problem and considering the algorithm used for LCM, there should be no reason to save these arrays.

**Reasons for Memory Allocation Issues**
When jitting a function with JAX a computation graph will be created, that graph will then be passed to a compiler for further optimization. The Memory Allocation Issues probably stem from this optimization step. It is possible to visualize the computation graph before and after the compiler optimization. Below you can see the optimized computation graph with older Jax and Jaxlib versions. For some reason, the compiler splits the fusion into two parts, instead of creating one big fusion, with a reduce operator as the root. The arrays get passed as parameters from one fusion to the other and therefore are saved in the GPU memory.

for_loop_comp

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,7 +87,7 @@ repos:
       - id: mypy
         files: src|tests
         additional_dependencies:
-          - jax>=0.4.20
+          - jax>=0.4.34
           - numpy
           - packaging
           - pandas-stubs
--- a/README.md
+++ b/README.md
@@ -5,17 +5,23 @@
 [![image](https://codecov.io/gh/OpenSourceEconomics/lcm/branch/main/graph/badge.svg)](https://codecov.io/gh/OpenSourceEconomics/lcm)
 
 This package aims to generalize and facilitate the specification, solution, and
-estimation of dynamic choice models.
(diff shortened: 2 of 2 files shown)


hmgaudecker:
Great explanation! From my perspective, you won't have to go into such detail -- it is not that we are talking about a library that is in use across thousands of weird environments here. Everyone can update their environments without problems.

timmens (approved):
Thanks a lot!

I like the detailed explanation, I believe it can help us/others to find similar problems in the future. For now, I am relieved that the memory issues are gone now! :tada: 

I've quickly updated the README and pre-commit hooks myself.

PR #101, opened 2025-02-03
Remove filters

timmens:
In this PR, we remove filters from the code-base.

We remove filters and related code from the code-base without starting a major refactoring. This will be done in a separate PR.

> [!NOTE]
> Sparse variables are still required in the simulation; however, no filters are needed and only state variables can be sparse not choices. I will think about a better name for "sparse" variables in this context in the next PRs.

#### Refactoring and Renaming

- Rename sparse and dense variables!
- Extensive refactoring given the simpler structure
- Update dispatchers explanation notebook
- [Update return type of `_determine_dense_discrete_choice_axes`](https://github.com/OpenSourceEconomics/lcm/pull/101#discussion_r1941023586)
- Replace double backticks with single backticks
- Currently we have `discrete_problem._determine_dense_discrete_choice_axes` and `simulate.determine_discrete_dense_choice_axes`: They look similar, and do similar things, but not exactly -> Combine these with an argument like `purpose: Literal["solution", "simulation"]`.


src/lcm/simulate.py
hmgaudecker, line 1: Again, rename `dense` to `discrete`, unless bigger changes are to come or I am missing something.

src/lcm/simulate.py
```diff
@@ -504,83 +490,20 @@ def create_data_scs(
...
-        name: grid
-        for name, grid in model.grids.items()
-        if name in vi.query("is_sparse & is_choice").index.tolist()
-    }
-
     dense_choices = {
         name: grid
         for name, grid in model.grids.items()
-        if name in vi.query("is_dense & is_choice & ~is_continuous").index.tolist()
+        if name in vi.query("is_choice & ~is_continuous").index.tolist()
```
hmgaudecker, line 496:
```suggestion
        if name in vi.query("is_choice & is_discrete").index.tolist()
```
Much easier to read.

src/lcm/simulate.py
```diff
@@ -655,32 +575,6 @@ def dict_product(d):
...
-        repeats=n_choice_combinations,
-    )
-    segments = state_ids[mask]
-    return {
-        "segment_ids": jnp.array(segments),
-        "num_segments": len(jnp.unique(segments)),
-    }
-
-
 def determine_discrete_dense_choice_axes(variable_info):
```
hmgaudecker, line 578: Seems the same as `discrete_problem._determine_dense_discrete_choice_axes`

src/lcm/simulate.py
```diff
@@ -693,7 +587,7 @@ def determine_discrete_dense_choice_axes(variable_info):
 
     """
     discrete_dense_choice_vars = variable_info.query(
-        "~is_continuous & is_dense & is_choice",
+        "~is_continuous & is_choice",
```
hmgaudecker, line 590:
```suggestion
        "is_choice & is_discrete",
```

Same order as above.

src/lcm/discrete_problem.py
```diff
@@ -213,31 +140,19 @@ def _segment_logsumexp(a, segment_info):
...
     """
-    has_sparse = variable_info["is_sparse"].any()
-
     # List of dense variables excluding continuous choice variables.
-    dense_vars = variable_info.query(
-        "is_dense & ~(is_choice & is_continuous)",
-    ).index.tolist()
-
-    axes = ["__sparse__", *dense_vars] if has_sparse else dense_vars
+    axes = variable_info.query("~(is_choice & is_continuous)").index.tolist()
```
hmgaudecker, line 155:
```suggestion
    axes = variable_info.query("is_state | is_discrete").index.tolist()
```

Combine the three lines starting with this one into one? Or will that be gone, anyhow?

tests/test_state_space.py
hmgaudecker, line 1: Just a little surprised so much is gone from this module. What is left is a check that we can call `create_state_choice_space` without error, right?

I thought that we now express filters by constraints, so we could re-use the tests and check for the correct masks being applied?

explanations/dispatchers.ipynb
```diff
@@ -49,6 +56,13 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [
+    {
```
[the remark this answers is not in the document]
timmens: Fully agree once we have docs; added it to the list of todos here: #91

src/lcm/discrete_problem.py
```diff
@@ -213,31 +140,19 @@ def _segment_logsumexp(a, segment_info):
 def _determine_dense_discrete_choice_axes(
```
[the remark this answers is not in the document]
timmens: I would say it is out-of-scope. The problem is that we still need --what we currently call sparse variables-- for the simulation. I would like to avoid a state of the code-base where we have "variables" and "sparse variables" to avoid confusion. In the upcoming renaming/refactoring PR I am planning to give these two concepts better fitting names.

src/lcm/discrete_problem.py
```diff
@@ -213,31 +140,19 @@ def _segment_logsumexp(a, segment_info):
...
-        variable_info (pd.DataFrame): DataFrame with information about the variables.
+        variable_info: DataFrame with information about the variables.
 
     Returns:
-        tuple[int, ...] | None: A tuple of indices representing the axes in the value
-            function that correspond to discrete choices. Returns None if there are no
-            discrete choice axes.
+        tuple[int, ...] | None: A tuple of indices representing the axes positions in
+            the value function that correspond to discrete choices. Returns None if
+            there are no discrete choice axes.
```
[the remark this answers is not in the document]
timmens: No, very good catch. We can just return the empty tuple here. I have added this to the list of todo's for the refactoring.

src/lcm/interfaces.py
```diff
@@ -80,18 +80,16 @@ class InternalModel:
...
         functions: Dictionary that maps names of functions to functions. The functions
-            differ from the user functions in that they all except the filter functions
-            take ``params`` as keyword argument. If the original function depended on
-            model parameters, those are automatically extracted from ``params`` and
-            passed to the original function. Otherwise, the ``params`` argument is
-            simply ignored.
+            differ from the user functions in that they take ``params`` as keyword
+            argument. If the original function depended on model parameters, those are
+            automatically extracted from ``params`` and passed to the original function.
+            Otherwise, the ``params`` argument is simply ignored.
```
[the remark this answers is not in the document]
timmens: I don't know, but they should be replaced with single backticks. I have added it as a todo for the renaming PR.

src/lcm/simulate.py
timmens: No they are not the same, but it is not clear at all from the code why; and that needs to change!

The condition in the first query is different and the choice indices in the function in the simulation module need to be offset by 1, to account for the initial states (sparse state variables).

I have added it to the list of ToDo's for the refactoring.

tests/test_state_space.py
timmens: Correct, this test file is a bit empty now. But I would add any new tests before the refactoring, since that will likely change the state-choice-space.

Yes, that is correct. However, the constraints are applied at a different place in the code-base. Constraints are much easier to work with because they do not mess with the state-space.

After this PR and the refactoring, the state-choice-space object becomes much simpler!

src/lcm/dispatchers.py
```diff
@@ -12,9 +12,7 @@
 def spacemap(
     func: F,
     dense_vars: list[str],
-    sparse_vars: list[str],
-    *,
-    put_dense_first: bool,
+    sparse_vars: list[str] | None = None,
```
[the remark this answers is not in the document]
timmens: Sparse variables are still required in the simulation; however, no filters are needed and only state variables can be sparse not choices. I will think about a better name for "sparse" variables in this context in the next PRs.

src/lcm/simulate.py
timmens: I would wait with this for the refactoring and renaming PRs. Additionally, there are also dense continuous variables.

hmgaudecker (approved):
As discussed, looks great with the possible exception of recycling the code from `test_state_space` !

PR #103, opened 2025-02-10
Remove sparse variables and refactor

timmens:


src/lcm/simulation/simulate.py
```diff
@@ -575,26 +537,25 @@ def dict_product(d):
...
 
     choice_vars = set(variable_info.query("is_choice").index.tolist())
 
-    # We add 1 because the first dimension corresponds to the sparse state variables
-    choice_indices = [
-        i + 1 for i, ax in enumerate(discrete_dense_choice_vars) if ax in choice_vars
-    ]
-
-    return None if not choice_indices else tuple(choice_indices)
+    # We must add 1 because the first dimension corresponds to the state variables,
```
hmgaudecker, line 557:
```suggestion
    # The first dimension corresponds to the simulated states, so add 1.
```

src/lcm/simulation/simulate.py
```diff
@@ -575,26 +537,25 @@ def dict_product(d):
...
+    # which are treated as combination variables during the simulation.
```
hmgaudecker, line 558:
```suggestion
```

src/lcm/solution/state_space.py
hmgaudecker, line 1: Rename file to `state_choice_space.py`

src/lcm/solution/state_space.py
```diff
@@ -0,0 +1,56 @@
...
+    discrete_states = {sn: model.gridspecs[sn] for sn in discrete_states_names}
+    continuous_states = {sn: model.gridspecs[sn] for sn in continuous_states_names}
+
+    state_grids = {sn: model.grids[sn] for sn in vi.query("is_state").index.tolist()}
+    choice_grids = {
+        sn: model.grids[sn] for sn in vi.query("is_choice & is_discrete").index.tolist()
+    }
+    ordered_var_names = vi.query("is_state | is_discrete").index.tolist()
+
+    state_space_info = SpaceInfo(
```
hmgaudecker, line 44: Can we make clear somewhere why `state_space_info` is helpful, but `state_choice_space_info` does not seem to be? If we are sure that is the case, maybe even rename the class to `StateSpaceInfo` ?

tests/simulation/test_simulate.py
```diff
@@ -40,7 +40,7 @@ def simulate_inputs():
     model_config = get_model_config("iskhakov_et_al_2017_stripped_down", n_periods=1)
     model = process_model(model_config)
 
-    _, space_info = create_state_choice_space(
+    _, sc_space_info = create_state_choice_space(
```
hmgaudecker, line 43:
```suggestion
    sc_space_info = create_state_choice_space( ... )[1]
```

Won't flag it anymore, if you agree, maybe use a regex to search for that pattern.

tests/test_model_functions.py
```diff
@@ -30,14 +30,14 @@ def test_get_utility_and_feasibility_function():
         },
     }
 
-    _, space_info = create_state_choice_space(
+    _, sc_space_info = create_state_choice_space(
```
hmgaudecker, line 33:
```suggestion
    _, s_space_info = create_state_choice_space(
```
Right?

hmgaudecker (commented):
Very neat! 

I think type annotations throughout would really help. Especially getting rid of `None` instances where empty containers will just do the trick and allow simplifying the code. No need to include here, but probably a good one to tackle soon.

src/lcm/discrete_problem.py
```diff
@@ -108,18 +104,18 @@ def _calculate_emax_extreme_value_shocks(values, choice_axes, params):
     """Aggregate conditional continuation values over discrete choices.
 
     Args:
-        values (jax.numpy.ndarray): Multidimensional jax array with conditional
+        values (jax.Array): Multidimensional jax array with conditional
```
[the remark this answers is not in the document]
timmens: Fully agree! Will do that in a separate PR for the entire code-base (#104).

src/lcm/dispatchers.py
```diff
@@ -11,71 +11,78 @@
 
 def spacemap(
     func: F,
-    dense_vars: list[str],
-    sparse_vars: list[str] | None = None,
+    product_vars: list[str],
+    combination_vars: list[str] | None = None,
```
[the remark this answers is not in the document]
timmens: Agree!

1. Will remove the default and switch to tuples (no reason for mutables here).
2. Changed it to `product_var_names` and `sim_states_var_names`.

src/lcm/discrete_problem.py
```diff
@@ -137,26 +133,20 @@ def _calculate_emax_extreme_value_shocks(values, choice_axes, params):
...
-        tuple[int, ...] | None: A tuple of indices representing the axes' positions in
-            the value function that correspond to discrete choices. Returns None if
-            there are no discrete choice axes.
+        tuple[int, ...]: A tuple of indices representing the axes' positions in
+            the value function that correspond to discrete choices.
 
     """
-    # List of dense variables excluding continuous choice variables.
+    # List of all model variables excluding the continuous choice variables.
     axes = variable_info.query("is_state | is_discrete").index.tolist()
```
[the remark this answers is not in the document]
hmgaudecker: Sure, that was nonsense. But wouldn't this do the same and be clearer?

```py
    discrete_choice_vars = set(variable_info.query("is_choice & is_discrete").index.tolist())
    return tuple(i for i, ax in enumerate(variable_info.index) if ax in discrete_choice_vars)
```

I guess it depends on whether we have a guarantee that continuous choices come last, which I thought would be the case?

src/lcm/dispatchers.py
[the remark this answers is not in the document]
timmens: The more I think about it, I’m starting to wonder if we might not actually want a single `spacemap` function for both the solution and simulation cases. Instead, it could be better to have two separate `spacemap` functions—one for the solution case and one for the simulation case. This would result in very little code duplication, but it would make the namespace much clearer.

What do you think?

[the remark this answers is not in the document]
hmgaudecker: Fully agree!

src/lcm/solution/state_space.py
timmens: The general `get_function_representation` function works with a `SpaceInfo` object. However, we **only** use this for the value function. Since the value function is defined on the state space (and not on the state-choice space), we exclusively need `state_space_info`.  

To align with our goal of reducing abstraction in the codebase, we could rename it to `StateSpaceInfo` and clarify that `get_function_representation` is intended specifically for the value function—unless you foresee a potential use case for it in the future.

hmgaudecker: Yeah, let's do that. Renaming the other way is always easy.

timmens:
As discussed, we will further split the code-base into parts that are relevant only for simulation or solution. For this we will also split `spacemap` into two functions; therefore we won't resolve the comments on `dispatchers.py` here, but in a separate PR.

hmgaudecker (approved):
:+1:

PR #105, opened 2025-02-14
Add preliminary typing to the entire code-base

### LINKED ISSUE #104: ENH: Add typing everywhere
- [x]  Add type hints to all functions
- [x]  Successfully run mypy on all files
- [x]  Remove type hints from docstrings (also return-types)

timmens:
- [x] Use stricter mypy configuration
- [x] Add typing to all functions
- [x] Successfully run mypy in strict mode on src and tests

Fixes #104


src/lcm/typing.py
```diff
@@ -15,3 +22,10 @@ class ShockType(Enum):
 
     EXTREME_VALUE = "extreme_value"
     NONE = None
+
+
+class Target(Enum):
+    """Target of the function."""
+
+    SOLVE = "solve"
```
hmgaudecker, line 30: Missed this before -- why do we use uppercase?

hmgaudecker (commented):
Couldn't resist to have a look -- impressive!

tests/test_discrete_problem.py
```diff
@@ -41,7 +41,7 @@ def test_get_solve_discrete_problem_illustrative():
         ],
     )
 
-    got = solve_discrete_problem(cc_values, params=None)
+    got = solve_discrete_problem(cc_values, params={"ignore": "this"})
```
hmgaudecker, line 44: Use an empty dict or explain. 
```suggestion
    got = solve_discrete_problem(cc_values, params={"ignore": "because ..."})
```

tests/test_discrete_problem.py
```diff
@@ -69,7 +69,7 @@ def test_solve_discrete_problem_no_shocks_illustrative():
     got = _solve_discrete_problem_no_shocks(
```
hmgaudecker, line 69: Make two tests out of this -- same below. Second check won't run if the first one fails.

src/lcm/entry_point.py
```diff
@@ -28,7 +30,7 @@ def get_lcm_function(
     *,
```
hmgaudecker, line 30: Can't we use the enum here? Also, I'd prefer getting rid of the default value.

src/lcm/function_representation.py
```diff
@@ -129,7 +131,7 @@ def get_value_function_representation(
 
 def _get_label_translator(
     in_name: str,
-) -> Callable[..., Array]:
+) -> Callable[..., Scalar]:
```
hmgaudecker, line 134:
```suggestion
) -> Callable[..., Int]:
```
? 

Same for the remainder here, some of that would be `Float`.

If we need the inclusion of 0-d Jax Arrays, we could define `IntScalar` and `FloatScalar`.

src/lcm/dispatchers.py
```diff
@@ -1,13 +1,13 @@
...
 from collections.abc import Callable
-from typing import Literal, TypeVar
+from typing import Literal, TypeVar, cast
 
 from jax import Array, vmap
 
 from lcm.functools import allow_args, allow_only_kwargs
 from lcm.utils import find_duplicates
 
-F = TypeVar("F", bound=Callable[..., Array])
```
hmgaudecker, line 10: In functools, you are using `ReturnType` for such a construct. I think I prefer the more verbose version.

src/lcm/input_processing/process_model.py
```diff
@@ -94,12 +94,13 @@ def _get_internal_functions(
...
+
     for name, func in raw_functions.items():
         is_weight_next_function = name.startswith("weight_next_")
 
         if is_weight_next_function:
-            processed_func = func
+            processed_func = cast(InternalUserFunction, func)
 
         else:
             # params[name] contains the dictionary of parameters for the function, which
```
hmgaudecker, line 106: lines 108-110 should just be:

```py
    if params[name]:
```

src/lcm/input_processing/process_model.py
```diff
@@ -94,12 +94,13 @@ def _get_internal_functions(
     # the dynamically generated weighting functions for stochastic next functions, since
     # they are constructed to accept the 'params' argument by default.
 
-    functions = {}
+    functions: dict[str, InternalUserFunction] = {}
+
     for name, func in raw_functions.items():
```
hmgaudecker, line 99:
```suggestion
    for func_name, func in raw_functions.items():
```
Makes it a bit clearer.

src/lcm/next_state.py
```diff
@@ -7,31 +7,48 @@
...
+        return _get_next_state_function_solution(model)
 
-def get_next_state_function(model: InternalModel, target):
-    if target == "solve":
-        out = _get_next_state_function_solution(model)
-    elif target == "simulate":
-        out = _get_next_state_function_simulation(model)
-    else:
-        raise ValueError(f"Target must be 'solution' or 'simulation'. Got {target}.")
+    if target == Target.SIMULATE:
```
hmgaudecker, line 38:
```suggestion
    elif target == Target.SIMULATE:
```

The old version seems more readable to me...

src/lcm/next_state.py
```diff
@@ -7,31 +7,48 @@
...
+        return _get_next_state_function_simulation(model)
 
-    return out
+    raise ValueError(f"Invalid target: {target}")
```
hmgaudecker, line 41: Can we ever reach this point?

src/lcm/next_state.py
```diff
@@ -7,31 +7,48 @@
...
 
 
 # ======================================================================================
 # Solution
 # ======================================================================================
 
 
-def _get_next_state_function_solution(model: InternalModel):
+def _get_next_state_function_solution(
```
hmgaudecker, line 49:
```suggestion
def _get_next_state_function_for_solution(
```
Else it is a bit ambiguous, could be the function's solution...

src/lcm/next_state.py
```diff
@@ -57,7 +74,9 @@ def _get_next_state_function_solution(model: InternalModel):
 # ======================================================================================
 
 
-def _get_next_state_function_simulation(model: InternalModel):
+def _get_next_state_function_simulation(
```
hmgaudecker, line 77:
```suggestion
def _get_next_state_function_for_simulation(
```

src/lcm/random_choice.py
```diff
@@ -3,23 +3,29 @@
 import jax
 
 
-def random_choice(key, probs, labels):
+def random_choice(
+    key: jax.Array,
+    probs: jax.Array,
+    labels: jax.Array,
+) -> jax.Array:
```
hmgaudecker, line 10: Add dtype info? Should be ints, right?

src/lcm/simulation/simulate.py
```diff
@@ -10,21 +13,22 @@
...
-    next_state,
-    logger,
-    solve_model=None,
-    vf_arr_list=None,
-    additional_targets=None,
-    seed=12345,
-):
+    next_state: Callable[..., dict[str, Array]],
+    logger: logging.Logger,
+    solve_model: Callable[..., list[Array]] | None = None,
```
hmgaudecker, line 27: For the next refactoring: I have a fairly strong opinion that we should have another function `solve_and_simulate`, which
- removes this keyword argument
- gets rid of the `None` option in `pre_computed_vf_arr_list` (then we can change to `value_function_arrays` and make that a tuple).

Also, we should get rid of the default for `seed`.

src/lcm/simulation/simulate.py
```diff
@@ -74,7 +80,7 @@ def simulate(
     # value function array is not required, as the agents only consider the current
```
hmgaudecker, line 80: Stupid q: Why do we compute/store this in the first place?

Just call this `next_vf_arr`. Also, store as a dict with periods as keys. Not for this PR, ofc.

src/lcm/simulation/simulate.py
```diff
@@ -347,16 +358,18 @@ def _process_simulated_data(results):
 # ======================================================================================
 
 
-def _generate_simulation_keys(key, ids):
+def _generate_simulation_keys(
+    key: Array, ids: list[str]
+) -> tuple[Array, dict[str, Array]]:
     """Generate PRNG keys for simulation.
```
hmgaudecker, line 364: Explain PRNG once? I can only guess / look up on the web.

src/lcm/simulation/simulate.py
```diff
@@ -444,7 +457,7 @@ def create_data_scs(
 
     Returns:
         - Space: Data state choice space.
-        - dict: Dict with choice segments.
+        - None
```
hmgaudecker, line 460:
```suggestion
```

src/lcm/simulation/simulate.py
```diff
@@ -444,7 +457,7 @@ def create_data_scs(
 
     Returns:
         - Space: Data state choice space.
```
hmgaudecker, line 459:
```suggestion
        Data state choice space.
```

src/lcm/simulation/simulate.py
```diff
@@ -519,11 +530,11 @@ def _calculate_discrete_argmax(values, choice_axes):
 # ======================================================================================
 
 
-def dict_product(d):
+def dict_product(d: dict[str, Array]) -> tuple[dict[str, Array], int]:
     """Create a product of the entries of a dictionary.
 
     Args:
-        d (dict): Dictionary where all values are arrays.
+        d: Dictionary where all values are arrays, and keys are strings.
```
hmgaudecker, line 537:
```suggestion
        d: Dictionary.
```
Just repeats the typing info.

src/lcm/simulation/simulate.py
```diff
@@ -519,11 +530,11 @@ def _calculate_discrete_argmax(values, choice_axes):
...
 
     Returns:
         - dict: Dictionary with same keys but values correspond to rows of product.
```
hmgaudecker, line 540: Unclear to me...

src/lcm/simulation/simulate.py
```diff
@@ -519,11 +530,11 @@ def _calculate_discrete_argmax(values, choice_axes):
 # ======================================================================================
 
 
-def dict_product(d):
+def dict_product(d: dict[str, Array]) -> tuple[dict[str, Array], int]:
```
hmgaudecker, line 533: Maybe include in the func name that the len is returned, too? Also, a short explainer of what exactly is returned would be great. I honestly have no idea from the code. Instead of `entries`, use `values`.

src/lcm/solution/solve_brute.py
```diff
@@ -75,31 +78,31 @@ def solve(
...
-        params (dict): Dict of model parameters.
+        vf_arr: Value function array.
+        params: Dict of model parameters.
 
     Returns:
-        jnp.ndarray: Jax array with continuation values for each combination of a
-            state and a discrete choice. The number and order of dimensions is defined
-            by the `gridmap` function.
+        Jax array with continuation values for each combination of a state and a
+            discrete choice. The number and order of dimensions is defined by the
```
hmgaudecker, line 104:
```suggestion
            set of all discrete choices. The number and order of dimensions is defined by the
```
?

Not perfect, but "a discrete choice" makes me think we are just looking at one of the potentially many discrete choices.

hmgaudecker (approved):
Much, much clearer! 

I do have a number of remarks, some of which are not related to typing. No show-stoppers, though!

src/lcm/typing.py
```diff
@@ -1,17 +1,45 @@
 from enum import Enum
-from typing import Any
+from typing import Any, Protocol
 
 from jax import Array
 
 # Many JAX functions are designed to work with scalar numerical values. This also
 # includes zero dimensional jax arrays.
 Scalar = int | float | Array
```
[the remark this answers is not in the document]
timmens: I think that is a very valid point, and I am not really happy with the current solution myself. I do think, however, this is out-of-scope for this PR.

I've opened an issue (#107).

pyproject.toml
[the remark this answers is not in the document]
hmgaudecker: Sure. I was just not sure about second-order type checking errors, but even better if that is not an issue!

src/lcm/utils.py
```diff
@@ -10,3 +10,11 @@ def find_duplicates(*containers: Iterable[T]) -> set[T]:
     combined = chain.from_iterable(containers)
     counts = Counter(combined)
     return {v for v, count in counts.items() if count > 1}
+
+
+class ReplaceMixin:
+    """Mixin for replacing init-attributes of a class."""
```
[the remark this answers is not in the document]
timmens: After reflecting on this, I realized I leaned too far into the OOP paradigm with mixins. I initially created `ReplaceMixin` to avoid duplicating the `replace()` method, but it ended up adding unnecessary complexity without significant line savings. Thanks for pointing that out—I’ve reworked the implementation.

src/lcm/entry_point.py
timmens: We can get rid of the default value, but we cannot use the `Enum` here for two reasons:

1. The `Enum` does not work for "solve_and_simulate". This is something we should discuss anyways: Should there even be a "solve_and_simulate" option? Or should there only be "solve" and "simulate", and if someone wants a simulation function that does not automatically do the solving step we add a boolean argument?
2. I would not want the user to import that `Enum`, but rather try to convert the user input into an `Enum` during the input processing.

hmgaudecker: Cool, makes sense.

I do think there should be a one-stop-shop, but via a separate function that calls `solve` and `simulate` in order. See my comment below in the interface to `simulate`.

src/lcm/dispatchers.py
timmens: Here the return type is known to be Array or tuple[Array, Array]. The construct `F` would map to the construct `Callable[..., ReturnType]` in functools.

I think there is room for improvement in the typing, but this would require me to look up advanced typing features. FYI, we are doing slightly more than JAX is doing in their code-base (see [here](https://github.com/jax-ml/jax/blob/44872296a139204ef9d4ca102859deb7305f8707/jax/_src/api.py#L94)).

src/lcm/next_state.py
timmens: Currently no. But if for any reason someone adds a target category to the `Target` enum, the function will raise an Error instead of returning `None`.

src/lcm/random_choice.py
timmens: There is currently no way to specify dtypes of jax arrays without additional packages like jaxtyping.

src/lcm/dispatchers.py
hmgaudecker: I just meant the string "F" (one character, could be anything) vs "ReturnType" (10 characters, allows to guess what is meant)!

src/lcm/random_choice.py
hmgaudecker: Cool, just add to #107 and adjust the description there.

src/lcm/dispatchers.py
timmens: Personally, I like `F`, as it is also what the JAX source-code is using for `vmap`. If you want something more declarative, I'd propose `FunctionWithArrayReturn`?

PR #108, opened 2025-02-26
Refactor solution and simulation

timmens:
In this PR, I try to make clear where simulation and solution code differs, and where it does not. I also perform some additional refactoring along the way.

Specifically, I

- [x] Refactored the next state module, to showcase the difference between simulation and solution functionality
- [x] Rename the `model_functions` module to `utility_and_feasibility` and rewrite for clarity
- [x] Use `productmap` instead of `spacemap` in the solution
- [x] Rename `spacemap` and adjust it so that it's usage during the simulation is clear
- [x] Integrate comments from the previous PRs
  - [x] I remove the default random seed
- [x] Create the `conditional_continuation` module
- [x] Start refactoring the entry point.
- [x] Fix stochastic test `test_get_lcm_function_with_simulate_target` (test did not account for stochastics...)
- [x] Implement `solve_and_simulate` function
- [x] Refactor simulation
- [x] Add continuous choice grids to state-choice-space
- [x] Combine both state-choice-space modules


src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
+        solve_and_simulate,
+        compute_ccv_policy_functions=compute_ccp_functions,
+        model=internal_model,
+        next_state=next_state_simulate,  # type: ignore[arg-type]
         logger=logger,
+        solve_model=solve_model,
     )
 
-    target_func: Callable[..., list[Array] | pd.DataFrame]
+    target_func: Callable[..., dict[int, Array] | pd.DataFrame]
```
hmgaudecker, line 162: Maybe add a comment that this is just repeating the signature to get rid of mypy complaints? _(I suppose)_

src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
-        model=_mod,
-        next_state=jax.jit(_next_state_simulate),
+        compute_ccv_policy_functions=compute_ccp_functions,
+        model=internal_model,
+        next_state=next_state_simulate,  # type: ignore[arg-type]
+        logger=logger,
+    )
+
+    solve_and_simulate_model = partial(
+        solve_and_simulate,
```
hmgaudecker, line 154: I would define the function `solve_and_simulate` here, I think. Or in a separate module of the same name. But I would not look for it in the `simulation.simulate` module...

If there was a clever way of telling `simulate_model` that the partialled argument to `simulate(vf_arr_dict)` will be the output of `solve_model` (2nd order partialling, I guess), that would be even better...

src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
         solve,
         state_choice_spaces=state_choice_spaces,
-        continuous_choice_grids=continuous_choice_grids,
         compute_ccv_functions=compute_ccv_functions,
-        emax_calculators=emax_calculators,
+        emax_calculators=solve_discrete_problem_functions,
         logger=logger,
     )
 
+    _next_state_simulate = get_next_state_function(
```
hmgaudecker, line 138: Move below `solve_model = ...`. I appreciate having both jit calls next to each other but having a "solve" and a "simulate" block would be even better.

src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
-            model=_mod,
-            state_space_info=state_space_infos[period],
+            model=internal_model,
+            next_state_space_info=next_state_space_info,
             period=period,
             is_last_period=is_last_period,
         )
 
-        compute_ccv = create_compute_conditional_continuation_value(
+        compute_ccv = get_compute_conditional_continuation_value(
```
hmgaudecker, line 105:
```suggestion
        compute_cc_val = get_compute_conditional_continuation_value(
```
Symmetry because I'd like `ccp` to be `cc_pol`

src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
             utility_and_feasibility=u_and_f,
-            continuous_choice_variables=list(_choice_grids),
+            continuous_choice_variables=tuple(state_choice_space.continuous_choices),
         )
-        compute_ccv_functions.append(compute_ccv)
 
-        compute_ccv_argmax = create_compute_conditional_continuation_policy(
+        compute_ccp = get_compute_conditional_continuation_policy(
```
hmgaudecker, line 110:
```suggestion
        compute_cc_pol = get_compute_conditional_continuation_policy(
```
Since we have CCP estimators and there `P` stands for `probability`, this is too confusing IMO.

src/lcm/entry_point.py
```diff
@@ -59,196 +63,116 @@ def get_lcm_function(
...
     # ==================================================================================
     # select requested solver and partial arguments into it
     # ==================================================================================
     _solve_model = partial(
         solve,
         state_choice_spaces=state_choice_spaces,
-        continuous_choice_grids=continuous_choice_grids,
         compute_ccv_functions=compute_ccv_functions,
-        emax_calculators=emax_calculators,
+        emax_calculators=solve_discrete_problem_functions,
```
hmgaudecker, line 134: Would there be a value in renaming the `emax_calculators` argument, too?

src/lcm/interfaces.py
```diff
@@ -6,6 +7,7 @@
 
 from lcm.grids import ContinuousGrid, DiscreteGrid, Grid
 from lcm.typing import InternalUserFunction, ParamsDict, ShockType
+from lcm.utils import first_non_none
 
 
 @dataclass(frozen=True)
```
hmgaudecker, line 13: Does this make the mutable attributes frozen, too? Else, would there be a value in using `FrozenDicts` or the like?

src/lcm/state_choice_space.py
```diff
@@ -0,0 +1,117 @@
...
+        ordered_var_names=ordered_var_names,
+    )
+
+
+def create_state_space_info(
+    model: InternalModel,
+    *,
+    is_last_period: bool,
+) -> StateSpaceInfo:
+    """Create a state-space information for the model solution.
```
hmgaudecker, line 66:
```suggestion
    """Collect information on the state space for the model solution.
```

src/lcm/state_choice_space.py
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    return StateSpaceInfo(
+        states_names=tuple(state_names),
+        discrete_states=discrete_states,
+        continuous_states=continuous_states,
+    )
+
+
+def _validate_initial_states(
```
hmgaudecker, line 103:
```suggestion
def _validate_initial_state_names(
```
This does not do validation of values, right? With the earlier name I'd expect that kind of checks.

hmgaudecker (commented):
Very nice! Only managed to look at the files I have comments in so far, will try to do more on the weekend but no guarantees.

src/lcm/simulation/simulate.py
```diff
@@ -1,218 +1,220 @@
...
+        logger=logger,
+        vf_arr_dict=vf_arr_dict,
+        additional_targets=additional_targets,
+        seed=seed,
+    )
 
 
 def simulate(
     params: ParamsDict,
     initial_states: dict[str, Array],
```
hmgaudecker, line 56: I did not try to understand in all detail, but at least `as_data_frame` makes a balanced panel out of this. We must eventually allow starting from different ages, though. Maybe make an issue out of this? *(while looking for whether we have one already, I added the fix for #96, example needed updating anyhow)*

src/lcm/dispatchers.py
```diff
@@ -12,64 +12,59 @@
...
+        scalar, the dispatched function returns an Array with k + 1 dimensions, where k
+        is the length of `choices_var_names` and the additional dimension corresponds to
+        the `states_var_names`. The order of the dimensions is determined by the order
+        of `choices_var_names`. If the output of `func` is a jax pytree, the usual jax
         behavior applies, i.e. the leading dimensions of all arrays in the pytree are as
         described above but there might be additional dimensions.
 
     """
-    if duplicates := find_duplicates(product_vars, combination_vars):
+    if duplicates := find_duplicates(choices_var_names, states_var_names):
```
hmgaudecker, line 53: IIUC, when running `solve_and_simulate`, this check would only be made after the first model solution. Can we do that upfront somehow and assume here that we get correct input? Not necessarily for this PR, ofc!

hmgaudecker (approved):
Very nice! With the renaming tbd for solving #109, I do not have any show-stoppers!

src/lcm/interfaces.py
timmens: I does not!

I am unsure whether `FrozenDicts` have a net-positive value. I'd wait until we have more usage data before adding another dependency.

src/lcm/dispatchers.py
timmens: This is checked when a user creates a `Model` specification instance. I like having this check here, so that it is easier for me to debug when I mess up something in between the user model and calling the dispatchers.

timmens:
Leaving the open remarks as unresolved as they are either about the entry point module, or about renaming; both of which will be tackled in separate PRs.

src/lcm/interfaces.py
hmgaudecker: I guess it would add safety. From searching and comparing a bit, [immutables.Map](https://github.com/MagicStack/immutables) seems like they would be the way to go in case we decide for that.

src/lcm/dispatchers.py
hmgaudecker: Cool. No need to leave this open, then, just add a little comment in the code.

PR #123, opened 2025-05-30
Enable dag annotations

timmens:


timmens:
@hmgaudecker I am starting to doubt the usefulness of this endeavor, at least in the current design.

- As pylcm models are build using functions, some of these functions can be used in several models. In most cases, it is irrelevant for the function whether an input is discrete or continuous in the pylcm sense. Hence, in one model an action can be continuous while in another model an action could be discrete -- in both cases the function would be the same. I think the fact that this is possible is something great about pylcm, making it easy to change components of a model. If we add type annotations that fix the discrete or continuous nature, then this is not possible anymore, and a "quick change" to see how a model changes if a discrete variable is used instead of a continuous one (or vice versa) would require major changes in possibly multiple modules. (This is appearing in the pylcm test suite where we re-use certain functions for a discrete and a continuous model.)
- Given the above point one could argue that we should then only focus on annotating that an input is a State or an Action variable (or an AuxiliaryVariable). Again, I would argue that we restrict pylcm's dynamic nature. If a user plays around and decides to fix an action (by removing it from the action space and adding it to the parameters), she would again have to change annotations everywhere.

I see three routes to take:
1. **We implement the feature as mandatory.** To fix the tests here I simply duplicate some code in the pylcm test suite such that I have the correct annotations for the respective model. IMO this could make using pylcm very annoying during development.
2. **We implement the feature as optional.** This way one could ignore the type annotations during the development stage of a model and check consistency afterwards (or iteratively). IMO this would be much better than (1), but I would still have doubts, because of the added complexity (checking the diff, a user needs to use at least 5 new types for annotations and 7 if we want to distinguish between discrete and continuous).
3. **We leave out the feature completely.** We could still show examples of adding `jax.Array` type annotations to user-functions. I would prefer this the most for various reasons: (a) one can use pylcm in a very dynamic way, (b) when looking at examples and docs you see the `jax.Array` annotation, making clear that at runtime this is a `jax.Array` (this would also help with code-completion).


What are your thoughts? Should we think of a re-design?

hmgaudecker:
Thanks! Hard to think for me of cases where you just want to switch between discrete and continuous and functions tension the same — could you point me to the cases in the test suite, so I get an idea of what you mean?

timmens:
In the test suite you can find two model modules: `tests/test_models/deterministic.py` and `tests/test_models/discrete_deterministic.py`. The discrete-deterministic model imports several model-functions from the deterministic one. If you run
```console
pixi run tests tests/simulation/test_simulate.py::test_simulate_with_only_discrete_actions
```
you get the `dags.exceptions.AnnotationsMismatchError`.

Whether this is a scenario that can happen in the "wild", I cannot tell. Judging from your reaction, this might occur only in test suites. In this case, we can also just move forward as planned!!

hmgaudecker:
Thanks! 

Indeed; switching some state from discrete to continuous would be a huge change in the model, for which something like the borrowing constraint (changed the name to match what is the set term) would definitely need adjustment. Even in this case, we can just use the union of types and the test runs through:

```python
def borrowing_constraint(
    consumption: ContinuousAction | DiscreteAction,
    wealth: ContinuousState
) -> ConstraintMask:
    return consumption <= wealth
```

*(somewhat scarily, the annotation for `wealth` does not require adjustment...)*

In the end, it is just the common trade-off between duck-typing and strong type hints. I am pretty confident the latter is helpful.

src/lcm/typing.py
```diff
@@ -1,8 +1,27 @@
...
+
+    def get(self) -> T:
+        return self.value
+
+
+type ContinuousState = Array
+type ContinuousAction = Array
+type DiscreteState = Array
+type DiscreteAction = Array
+type AuxiliaryVariable = Array
```
hmgaudecker, line 21:
```suggestion
type DerivedBool = Bool[Array, "..."]
```
using `jaxtyping`. And add `Int` and `Float` versions. Without the type info, the hint is not quite as helpful. `ContinuousState` should become `Float[Array, "..."]`  and `DiscreteState` probably `Int[Array, "..."]` (though I am not perfectly sure about the dataclasses here). 

We might be able to do better than the `"..."` for the shape info ([link](https://docs.kidger.site/jaxtyping/api/array/#shape)), but I am too far removed of lcm's internals right now to tell.

src/lcm/grids.py
```diff
@@ -93,7 +95,7 @@ def to_jax(self) -> Array:
     def get_coordinate(self, value: Scalar) -> Scalar:
         """Get the generalized coordinate of a value in the grid."""
 
-    def replace(self, **kwargs: Any) -> "ContinuousGrid":  # noqa: ANN401
+    def replace(self, **kwargs: Any) -> ContinuousGrid:  # noqa: ANN401
```
hmgaudecker, line 98:
```suggestion
    def replace(self, **kwargs: int | float) -> ContinuousGrid:   
```

? (not sure whether that works, but we know the types)


Also, please just use `import dataclasses` above (same in `user_model.py`), we only use it once and I had to look up what was meant by `dc`.

src/lcm/typing.py
```diff
@@ -1,8 +1,27 @@
...
+type ConstraintMask = Array
+type Utility = Array
```
hmgaudecker, line 23:
```suggestion
type Utility = Array
```
We can just use `DerivedFloat` if we stick with the above strategy.

tests/test_Q_and_F.py
```diff
@@ -68,18 +70,24 @@ def test_get_Q_and_F_function():
 
 @pytest.fixture
 def internal_model_illustrative():
-    def age(period):
+    def age(period: int) -> int:
         return period + 18
 
-    def mandatory_retirement_constraint(retirement, age, params):  # noqa: ARG001
+    def mandatory_retirement_constraint(
+        retirement: Array, age: int, params: ParamsDict
```
hmgaudecker, line 77:
```suggestion
        retirement: DiscreteAction, age: int, params: ParamsDict
```

right?

tests/test_Q_and_F.py
```diff
@@ -68,18 +70,24 @@ def test_get_Q_and_F_function():
...
+    ):
         # Individuals must be retired from age 65 onwards
         return jnp.logical_or(retirement == 1, age < 65)
 
-    def mandatory_lagged_retirement_constraint(lagged_retirement, age, params):  # noqa: ARG001
+    def mandatory_lagged_retirement_constraint(
+        lagged_retirement: Array, age: int, params: ParamsDict
```
hmgaudecker, line 83:
```suggestion
        lagged_retirement: DiscreteState, age: int, params: ParamsDict
```
?

src/lcm/typing.py
hmgaudecker: Note I kind of like the `Derived` better than `Auxiliary` because it makes clear that the stuff is somehow based on states and actions. But no need to spend time on changing the entire naming conventions if it takes too much time / puts knots in your head.

src/lcm/typing.py
```diff
@@ -2,14 +2,23 @@
...
+type DiscreteAction = Int[Array, "..."]
+type DerivedFloat = Float[Array, "..."]
+type DerivedInt = Int[Array, "..."]
+type DerivedBool = Bool[Array, "..."]
 
 ParamsDict = dict[str, Any]
 
+# Many JAX functions are designed to work with scalar numerical values. This also
+# includes zero dimensional jax arrays.
+Scalar = int | float | ScalarArray
```
hmgaudecker, line 20: Ouch, apologies for not seeing this earlier. Same comment as above for the arrays -- I am pretty sure we want the distinction between ints and floats here. So something like:

```py
ScalarInt = int | Int[Scalar]
```

? 

*(I did not check jaxtyping syntax)*

tests/test_Q_and_F.py
```diff
@@ -68,18 +77,30 @@ def test_get_Q_and_F_function():
 
 @pytest.fixture
 def internal_model_illustrative():
-    def age(period):
+    def age(period: DerivedInt) -> DerivedInt:
```
hmgaudecker, line 80: Is this a scalar or an array?

tests/test_entry_point.py
```diff
@@ -207,8 +214,8 @@ def test_get_max_Q_over_c():
         next_V_arr=jnp.empty(0),
     )
     assert val == iskhakov_et_al_2017_utility(
-        consumption=30.0,
-        working=RetirementStatus.working,
+        consumption=30.0,  # type: ignore[arg-type]
```
hmgaudecker, line 217: Just ignore arg-type for this file (or the tests in general)?

src/lcm/state_action_space.py
```diff
@@ -1,11 +1,16 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from lcm.grids import ContinuousGrid, DiscreteGrid
 from lcm.interfaces import InternalModel, StateActionSpace, StateSpaceInfo
 
+if TYPE_CHECKING:
+    import pandas as pd
+    from jax import Array
```
hmgaudecker, line 12:
```suggestion
    from lcm.typing import ContinuousState | DiscreteState
```

*(from memory)*

src/lcm/grids.py
```diff
@@ -93,7 +97,7 @@ def to_jax(self) -> Array:
     def get_coordinate(self, value: Scalar) -> Scalar:
         """Get the generalized coordinate of a value in the grid."""
 
-    def replace(self, **kwargs: Any) -> "ContinuousGrid":  # noqa: ANN401
+    def replace(self, **kwargs: float) -> ContinuousGrid:
```
hmgaudecker, line 100:
```suggestion
    def replace(self, **kwargs: float | int) -> ContinuousGrid:
```
I could replace `n_points`, right?

hmgaudecker (commented):
Looks great in general! However, while still fresh, I think it's worth investing the afternoon to be stricter about our own types rather than the generic `jax.Array`.

Maybe do a regex search of:

```: (Array|Scalar)```

and see what can be replaced quickly? No need to be super-pedantic, if that is the best we can do somewhere without a major effort, no problem. But many cases will clearly be `DiscreteState | ContinuousState` or the like.

File: tests/test_Q_and_F.py (hunk 7 above)
Comment (timmens) [reply tests/test_Q_and_F.py:80]: In the solution scalar, in the simulation an array.

**Comment by timmens** [comment]
I tried to follow your suggestion and updated the type hints. I've also added some to represent arbitrary 1-dimensional arrays that are not directly linked to a choice or state. If you dislike this, I can remove it again.

File: src/lcm/function_representation.py [hunk 11]
```diff
@@ -172,11 +181,11 @@ def _get_lookup_function(
...
 
-    @with_signature(args=arg_names)
-    def lookup_wrapper(*args: Scalar, **kwargs: Scalar) -> Scalar:
+    @with_signature(args=dict.fromkeys(arg_names, "Array"), return_annotation="Array")
+    def lookup_wrapper(*args: Array, **kwargs: Array) -> Array:
         kwargs = all_as_kwargs(args, kwargs, arg_names=arg_names)
         positions = tuple(kwargs[var] for var in axis_names)
-        arr = cast("Array", kwargs[array_name])
+        arr = kwargs[array_name]
         return arr[positions]
```
hmgaudecker, line 189:
```suggestion
        return kwargs[array_name][positions]
```

Now that there is no `cast()` any more, this is easier to read.

timmens:
Thanks! I had to make a few additional changes:

1. The typing module contains only types that are used for type checking now
2. I added the jaxtyping library to the list of dependencies, since users need to be able to import from it

timmens:
Do you mean the `return kwargs[array_name][positions]`? I already implemented your suggestion in the last commit. Or is there something else?

hmgaudecker:
> Do you mean the `return kwargs[array_name][positions]`? I already implemented your suggestion in the last commit. Or is there something else?

Yes! For whatever reason, I did not see that one in what I had looked at...

PR #138, opened 2025-08-19
Fix #30

### LINKED ISSUE #30: BUG: vmap is applied on variables which are not arguments of function
LCM Version: af6661ee732fa3bc3f8fa385433915b8a809d76e

### Example

A state variable that only influences the next period. In this case, the last period utility `compute_ccv` function does not depend on this state variable, since the `next_`-functions are not included. But, since it is a state variable, LCM tries to vmap over it, resulting in an error.

### Potential Solution

For each period compare variables over which we want to vmap, and variables that are in the signature of the functions we wish to vmap over. Then update the state space and space info accordingly.

### Current Fix

Add all state variables as arguments to the utility function, no matter whether they are actually used or not.

timmens:
Fixes #30 by preventing vmap from being applied to variables that aren't function arguments.

Key changes:
- Modified `get_next_state_function()` to accept `next_states` parameter for selective computation
- Updated simulation to only compute relevant states per period based on newly introduced `enters_concurrent_valuation` and `enters_transition` flags for states and actions variables.

---
Closes #30

### DIFF
--- a/src/lcm/input_processing/util.py
+++ b/src/lcm/input_processing/util.py
@@ -70,21 +70,17 @@ def get_variable_info(model: Model) -> pd.DataFrame:
         for var in variables
     ]
 
-    enter_concurrent_valuation = _get_variables_that_enter_concurrent_valuation(
+    info["enters_concurrent_valuation"] = _indicator_enters_concurrent_valuation(
         states_and_actions_names=list(variables),
         function_info=function_info,
         user_functions=model.functions,
     )
-    info["enters_concurrent_valuation"] = [
-        var in enter_concurrent_valuation for var in variables
-    ]
 
-    enter_transition = _get_variables_that_enter_transition(
+    info["enters_transition"] = _indicator_enters_transition(
         states_and_actions_names=list(variables),
         function_info=function_info,
         user_functions=model.functions,
     )
-    info["enters_transition"] = [var in enter_transition for var in variables]
 
     order = info.query("is_discrete & is_state").index.tolist()
     order += info.query("is_discrete & is_action").index.tolist()
@@ -97,16 +93,16 @@ def get_variable_info(model: Model) -> pd.DataFrame:
     return info.loc[order]
 
 
-def _get_variables_that_enter_concurrent_valuation(
+def _indicator_enters_concurrent_valuation(
     states_and_actions_names: list[str],
     function_info: pd.DataFrame,
     user_functions: dict[str, UserFunction],
-) -> list[str]:
-    """Get variables that enter the concurrent valuation.
+) -> pd.Series[bool]:
+    """Determine which states and actions enter the concurrent valuation.
 
     The concurrent valuation is the evaluation of the Q_and_F function. Hence, all
-    variables that influence the "utility" (Q), as well as the constraints (F), count
-    as relevant for the concurrent valuation.
+    variables that (directly or indirectly) influence the "utility" (Q) or the
+    constraints (F), count as relevant for the concurrent valuation.
 
     Special variables such as the "_period" or parameters will be ignored.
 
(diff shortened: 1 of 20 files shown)


hmgaudecker (commented):
Thanks! Just a couple of comments while trying to get an overview...

src/lcm/input_processing/util.py
```diff
@@ -70,12 +70,21 @@ def get_variable_info(model: Model) -> pd.DataFrame:
...
-        state_variables=info.query("is_state").index.tolist(),
+    enter_concurrent_valuation = _get_variables_that_enter_concurrent_valuation(
+        states_and_actions_names=list(variables),
         function_info=function_info,
         user_functions=model.functions,
     )
-    info["is_auxiliary"] = [var in auxiliary_variables for var in variables]
+    info["enters_concurrent_valuation"] = [
+        var in enter_concurrent_valuation for var in variables
+    ]
```
hmgaudecker, line 80:
```suggestion
    info["enters_concurrent_valuation"] = _indicator_enters_concurrent_valuation(
        states_and_actions_names=list(variables),
        function_info=function_info,
        user_functions=model.functions,
    )
```
No strong opinion about the naming, but the function should directly return what we need.

Same below.

src/lcm/input_processing/util.py
```diff
@@ -88,33 +97,52 @@ def get_variable_info(model: Model) -> pd.DataFrame:
...
 
-def _get_auxiliary_variables(
-    state_variables: list[str],
+def _get_variables_that_enter_concurrent_valuation(
+    states_and_actions_names: list[str],
     function_info: pd.DataFrame,
     user_functions: dict[str, UserFunction],
 ) -> list[str]:
-    """Get state variables that only occur in next functions.
+    """Get variables that enter the concurrent valuation.
```
hmgaudecker, line 105:
```suggestion
    """Get states and action variables that enter the concurrent valuation.
```

Should be symmetric to function below.

src/lcm/next_state.py
```diff
@@ -63,20 +71,20 @@ def get_next_state_function(
 
 def get_next_stochastic_weights_function(
     model: InternalModel,
+    next_stochastic_variables: tuple[str, ...],
```
hmgaudecker, line 74:
```suggestion
    next_stochastic_states: tuple[str, ...],
```

I think there is value in being specific here, unless it clashes with terminology used throughout.

src/lcm/simulation/simulate.py
```diff
@@ -89,8 +85,29 @@ def simulate(
...
+
+        relevant_states = model.variable_info.query(
+            "is_state and enters_concurrent_valuation"
+        ).index.tolist()
+        if not is_last_period:
+            relevant_states += model.variable_info.query(
+                "is_state and enters_transition"
+            ).index.tolist()
+
+        states_for_state_action_space = {n: states[n] for n in set(relevant_states)}
```
hmgaudecker, line 103:
```suggestion
        if is_last_period:
            query = "is_state and enters_concurrent_valuation"
        else:
            query = "is_state and (enters_concurrent_valuation | enters_transition)"
        states_for_state_action_space = {
            n: states[n] for n in model.variable_info.query(query).index
        }
```

Seems clearer I hope?

src/lcm/simulation/simulate.py
```diff
@@ -89,8 +85,29 @@ def simulate(
...
+
+        state_action_space = create_state_action_space(
+            model=model,
+            initial_states=states_for_state_action_space,
```
hmgaudecker, line 107:
```suggestion
            current_states=states_for_state_action_space,
```
?

Looks a little like the naming might be a leftover of this being done once for all periods? More a question/reminder than request for implementation, though.

src/lcm/simulation/simulate.py
```diff
@@ -89,8 +85,29 @@ def simulate(
     # ----------------------------------------------------------------------------------
     simulation_results = {}
 
+    last_period = n_periods - 1
```
hmgaudecker, line 88:
```suggestion
```
Indicator below is enough.

src/lcm/simulation/simulate.py
```diff
@@ -89,8 +85,29 @@ def simulate(
...
+
     for period in range(n_periods):
-        state_action_space = state_action_space.replace(states)
+        logger.info("Period: %s", period)
+
+        is_last_period = period == last_period
```
hmgaudecker, line 93:
```suggestion
        is_last_period = period == n_periods - 1
```

src/lcm/simulation/simulate.py
```diff
@@ -145,25 +162,30 @@ def simulate(
...
+
+            states_with_next_prefix = next_state(
+                **states,
+                **optimal_actions,
+                _period=jnp.repeat(period, n_initial_states),
+                params=params,
+                keys=stochastic_variables_keys,
+            )
+            # 'next_' prefix is added by the next_state function, but needs to be
+            # removed because in the next period, next states will be current states.
```
hmgaudecker, line 185:
```suggestion
            # removed for the next iteration of the loop, where these will be the current states.
```

hmgaudecker (approved):
Very nice, thanks! Just a few small suggestions for renamings/code clarity.

PR #139, opened 2025-08-19
Fix stochastics

### LINKED ISSUE #39: BUG: Stochastic next functions need to depend on at least one argument
## This works
```python
@lcm.mark.stochastic
def next_health(health):  # noqa: ARG001
    pass
```
## This does not
```python
@lcm.mark.stochastic
def next_partner():
    pass
```
 
The second version should also work. The problem currently is in the weighting function:

```
    new_kwargs = [*function_parameters, "params"]

    @with_signature(args=new_kwargs)
    def weight_func(*args, **kwargs):
        args = all_as_args(args, kwargs, arg_names=new_kwargs)
        
        # by definition of new_kwargs the last argument is params
        params = args[-1]

        indices = [
            label_translators[arg_name](arg)
            for arg_name, arg in zip(function_parameters, args, strict=False)
        ]

        return params["shocks"][name][*indices]
```

In particular, the `indices` are empty in the second case because there is no argument, although we want an index for each initial state.

timmens:
There seems to be something wrong about the stochastics in PyLCM.

In this PR, I want to validate and potentially fix the parts in PyLCM that work with stochastics. This will most likely close #39.

In particular, we have to answer the following questions:

- [ ] Mathematically, what do the state transition probabilities represent (joint probabilities or conditional probabilities)
- [ ] Given that we assume that stochastic state transitions can depend on other stochastic variables, does this interfere with our assumption that in a given period, the stochastic states are independent?

I believe here we must start form the mathematics again, and then work ourselves down to the implementation.

### DIFF
--- a/src/lcm/input_processing/process_model.py
+++ b/src/lcm/input_processing/process_model.py
@@ -223,14 +223,18 @@ def _get_stochastic_weight_function(
             f"but {name} depends on {invalid}.",
         )
 
-    annotations = get_annotations(raw_func) | {"params": "ParamsDict"}
+    annotations = get_annotations(raw_func)
     annotations.pop("return")
 
-    @with_signature(args=annotations, return_annotation="FloatND")
+    @with_signature(
+        args=annotations | {"params": "ParamsDict"}, return_annotation="FloatND"
+    )
     def weight_func(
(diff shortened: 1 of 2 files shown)


hmgaudecker:
Can be closed, I believe?

PR #140, opened 2025-08-19
Validate value function array

### LINKED ISSUE #55: ENH: Throw an error when utility is not defined on a grid value
## Current behavior

If the utility is evaluated on a grid value that is undefined we either introduce NAs, which lead to an error, or we introduce negative infinity, which does not result in an error.

## Expected behavior

In any of the above two cases, throw an informative error, so that either the calculated value function contains only proper numbers or an error is thrown.

timmens:
In this PR, I add functionality to validate the value function array and raise errors / log data in case the array contains invalid values.

Closes #55

### ToDo

- [ ] Check whether CSV-files already exist and potentially change filename
- [ ] Add logged file names to error message
- [ ] Add test case for when an invalid value function only occurs in the simulation

### DIFF
--- a/src/lcm/_config.py
+++ b/src/lcm/_config.py
@@ -3,3 +3,18 @@ from __future__ import annotations
 from pathlib import Path
 
 TEST_DATA = Path(__file__).parent.parent.parent.resolve().joinpath("tests", "data")
+
+LOG_DIRECTORY = ".pylcm"
+
+
+def get_log_dir() -> Path:
+    """Returns the absolute path to the package's log directory.
+
+    If the directory does not exist, it will be created.
+
+    """
+    # Find project root: assume current working dir if not configured
+    project_root = Path.cwd()
+    log_dir = project_root / LOG_DIRECTORY
+    log_dir.mkdir(parents=True, exist_ok=True)
+    return log_dir
--- a/src/lcm/error_handling.py
+++ b/src/lcm/error_handling.py
@@ -1,0 +1,77 @@
+from collections.abc import Callable
+
+import jax.numpy as jnp
+import pandas as pd
+from jax import Array
+
+from lcm._config import get_log_dir
+from lcm.exceptions import InvalidValueFunctionError
+from lcm.interfaces import StateActionSpace
+
+
+def validate_value_function_array_integrity(
+    V_arr: Array,
+    state_action_space: StateActionSpace,
+) -> None:
+    """Validate the value function array for NaN and infinity values.
+
+    This function checks the value function array for any NaN or infinity values. If any
+    such values are found, we log the corresponding state-combinations to
+
+    Args:
+        V_arr: The value function array to validate.
+        state_action_space: The state-action space to check against.
+
+    Raises:
+        InvalidValueFunctionError: If the value function array contains NaN or infinity
+            values. The invalid states are logged to CSV files in the '.pylcm' directory
+            for further analysis.
+
+    """
+    # Early return if the value function array contains only valid values
+    if not jnp.any(jnp.isnan(V_arr) | jnp.isinf(V_arr)):
+        return
+
+    nan_states_df = _get_dataframe_of_invalid_states(
+        V_arr=V_arr,
+        check_fn=jnp.isnan,
+        ordered_state_names=list(state_action_space.states),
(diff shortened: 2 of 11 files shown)


src/lcm/_config.py
```diff
@@ -3,3 +3,18 @@
...
+LOG_DIRECTORY = ".pylcm"
+
+
+def get_log_dir() -> Path:
+    """Returns the absolute path to the package's log directory.
+
+    If the directory does not exist, it will be created.
+
+    """
+    # Find project root: assume current working dir if not configured
```
timmens, line 16: Delete this line.

src/lcm/error_handling.py
```diff
@@ -0,0 +1,77 @@
...
+        ordered_state_names=list(state_action_space.states),
+        states=state_action_space.states,
+    )
+
+    inf_states_df = _get_dataframe_of_invalid_states(
+        V_arr=V_arr,
+        check_fn=jnp.isinf,
+        ordered_state_names=list(state_action_space.states),
+        states=state_action_space.states,
+    )
```
timmens, line 47: Lets combine this into a single check, single `invalid_states_df` and single CSV file.

tests/test_entry_point.py
timmens, line 1: Revert.

src/lcm/error_handling.py
```diff
@@ -42,12 +44,30 @@ def validate_value_function_array_integrity(
         states=state_action_space.states,
         period=period,
     )
-
+    extrapolation = False
```
hmgaudecker, line 47: Please call this `invalid_extrapolation` or so. In many cases, extrapolation beyond the grid is perfectly fine. E.g., our last gridpoint on the consumption grid could be 1M; if then we end up with the founder of a Unicorn company in our simulations, she will likely consume more, which would be perfectly fine ofc...

src/lcm/error_handling.py
```diff
@@ -42,12 +44,30 @@ def validate_value_function_array_integrity(
...
+                extrapolation = True
     log_dir = get_log_dir()
     file_path = log_dir / f"invalid_states_{_generate_unique_suffix()}.csv"
 
     invalid_states.to_csv(file_path, index=False, header=True)
-
+    if extrapolation:
+        raise InvalidValueFunctionError(
+            f"The value function array in period {period} contains NaN or infinite "
+            f"values. This might occur because the value function is extrapolated "
```
hmgaudecker, line 65: I don't think we ever extrapolate the value function. 

In general, I am not sure I follow when this special case would happen (sorry it's been a while...). Could you maybe add a test for this particular error to show up?

mj023: You are right, I will add test with an example of the error.

mj023: The test I added should better illustrate the problem, during the solution of the model no `-inf` occurs in the value functions, but then during the simulation the agent is suddenly in a state where he has no valid actions. 

If you think the model is just too misspecified I'm fine with that, then I will just remove the error I added. I thought the behaviour was very odd, so I wanted to warn the user in cases where it can happen.

src/lcm/error_handling.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Args:
+        V_arr: The value function array to validate.
+        period: The period for which the value function is being validated.
+
+    Raises:
+        InvalidValueFunctionError: If the value function array contains NaN values.
+
+    """
+    if jnp.any(jnp.isnan(V_arr)):
```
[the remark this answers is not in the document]
mj023: A model of this size would probably not be solvable without commiting a considerable amount of hardware (Multiple GPU's, Big CPU Clusters). The hardware would then speed up the check a lot, especially on a GPU, because the memory is faster. If I remeber correctly, I did a similar test with a GPU and the time for the check was fine, compared to the time it took to solve the model.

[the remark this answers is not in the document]
hmgaudecker: *(and yes, model seems rather large, hard to think of needing 300-500 grid points in some dimension, though I have seen models with a couple of more discrete states)*

hmgaudecker:
Last thing: Please modify the PR description before merging.

PR #145, opened 2025-08-22
Install Jax with PyPi

mj023:
Update the `pyproject.toml` to install Jax with PyPi instead of Anaconda on GPU as well as CPU. The PyPi repository is usually updated faster by Jax and is also the official way to install according to their website.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,6 @@ description = "Specification, solution, and simulation of finite-horizon dynamic
 requires-python = ">=3.12"
 dependencies = [
     "dags>=0.4.1",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -8,8 +8,6 @@ description = "Specification, solution, and simulation of finite-horizon dynamic
 requires-python = ">=3.12"
 dependencies = [
     "dags>=0.4.1",
-    "jax>=0.7",
```
timmens, line 11: I don't think we can remove these here, since this is the list of the library dependencies.

timmens (changes requested):

PR #146, opened 2025-08-26
Add LEP-01: Model Blocks

timmens:


docs/ep-01-index-representation.md
```diff
@@ -0,0 +1,193 @@
...
+and actions to influence stochastic transitions.
+
+## Backwards Compatibility
+
+All changes are fully backward compatible.
+
+## Motivation
+
+To Be Written.
+
```
hmgaudecker, line 18:
```suggestion
```
Irrelevant / obvious. Don't waste time on fitting a scheme that does not fit yet.

docs/ep-01-index-representation.md
```diff
@@ -0,0 +1,193 @@
...
+## Index Representation
+
+To Be Written.
+
+## Period-dependent Discrete Actions
+
+Consider a model where agents have to work until they are 60, then they can decide
+whether they want to work or retire until they are 70, and then they must retire. This
+is a working decision (action) with categories `WORK` and `RETIRE`, where the valid
```
hmgaudecker, line 27:
```suggestion
is an action with categories `WORK` and `RETIRE`, where the valid
```

docs/ep-01-index-representation.md
```diff
@@ -0,0 +1,193 @@
...
+
+@dataclass
+class WorkingAction:
+    work: int = 0
+    retire: int = 1
+
+actions={
+	"working": DiscreteGrid(
+		WorkingAction,
+		valid={
```
hmgaudecker, line 162: Looking at this and other parts ("Period-dependent *Discrete* Actions", "Time-dependent Markov Transitions") makes me wonder to what extent it is helpful to start with this while not being explicit about model blocks yet. In particular, this version of the interface (not necessarily the grid-approach) feels already (too) complex in this baby example, I don't want to imagine how it may look like when we have 4 different blocks in a period... 

My view from a glance is to either keep this as small as possible, sticking only to a period-dependent index definition, which we are sure about that it will directly extent to period / model block-dependent index definitions. Or to treat both period-dependent discrete action definitions and period-dependent state-choices spaces (a.k.a. model blocks) at the same time right away.

docs/ep-01-index-representation.md
hmgaudecker, line 1: Please use "LEP" instead of "EP" throughout (titles, file names), too hard to search otherwise.

hmgaudecker (commented):
Could not resist having a look... I think it is worth it delineating the scope very clearly before continuing.

timmens:
Done in #160

PR #156, opened 2025-09-29
Vmap next state calculation during simulation

mj023:
Currently the calculation of the next state is handled differently during the solution and during the simulation. During the solution, the `next_state`-functions are vmapped while during the simulation the `next_state`-functions are used as if they were vectorized functions. This can lead to problems if the `next_state`-functions are not vectorized or can't handle input arrays of different sizes like in GETTSIM.

This PR introduces vmap to the next state calculation during the simulation.

### DIFF
--- a/src/lcm/random.py
+++ b/src/lcm/random.py
@@ -28,7 +28,7 @@ def generate_simulation_keys(
 
     Returns:
         - Updated random key.
-        - Dict with random keys for each id in ids.
+        - Dict with n=n_initial_states new random keys for each name in names.
 
     """
     simulation_keys = {}


src/lcm/random.py
```diff
@@ -1,40 +1,13 @@
...
-@partial(jax.vmap, in_axes=(0, 0, None))
-def _vmapped_choice(key: Array, probs: Array, labels: Array) -> Array:
-    return jax.random.choice(key, a=labels, p=probs)
-
-
 def generate_simulation_keys(
-    key: Array, ids: list[str]
+    key: Array, names: list[str], n_initial_states: int
 ) -> tuple[Array, dict[str, Array]]:
     """Generate pseudo-random number generator keys (PRNG keys) for simulation.
```
hmgaudecker, line 12: Unrelated to this PR, but I am not familiar with the "simulator keys" terminology. Can we explain that somewhere? Or do we do it already and I should have searched?

hmgaudecker (commented):
Thanks a lot! 

Looks great to me, but I think @timmens should have the final word.

src/lcm/random.py
```diff
@@ -50,18 +23,20 @@ def generate_simulation_keys(
 
     Args:
         key: Random key.
-        ids: List of names for which a key is to be generated.
+        names: List of names for which a key is to be generated.
+        n_initial_states: Number of initial states.
 
     Returns:
         - Updated random key.
         - Dict with random keys for each id in ids.
```
timmens, line 31: This text should be updated.

src/lcm/simulation/simulate.py
```diff
@@ -159,25 +160,42 @@ def simulate(
...
-                params=params,
-                keys=stochastic_variables_keys,
+                **stochastic_variables_keys,
+            } | {"_period": period, "params": params}
+
+            states_with_next_prefix = next_state_vmapped(
+                **{
+                    parameter: next_state_keywords[parameter]
+                    for parameter in parameters
+                }
```
timmens, line 198: Why is this intermediate step necessary now, or was that actually necessary before as well?

timmens (approved):
Very nice fix, and very clean implementation! Approved.

I only have a minor comment, but feel free to merge this once you're done with it.

PR #161, opened 2025-10-15
Split functions dictionary

timmens:
To be done:

- [x] Fix tests
- [x] Remove functions_info
- [x] Fix mypy
- [ ] ~Think about whether the keys in the transition dictionary should start with next or not.~

### DIFF
--- a/src/lcm/model_initialization.py
+++ b/src/lcm/model_initialization.py
@@ -35,7 +35,6 @@ def initialize_model_components(model: Model) -> None:
         ModelInitilizationError: If initialization fails
 
     """
-    components = _get_model_components(model)
     try:
         components = _get_model_components(model)
         for name, component in components.items():
--- a/src/lcm/simulation/processing.py
+++ b/src/lcm/simulation/processing.py
@@ -59,12 +59,15 @@ def process_simulated_data(
     out["_period"] = jnp.repeat(jnp.arange(n_periods), n_initial_states)
 
     if additional_targets is not None:
+        model_functions = (
+            internal_model.functions
+            | {"utility": internal_model.utility}
+            | internal_model.constraints
+        )
         calculated_targets = _compute_targets(
             out,
             targets=additional_targets,
-            model_functions=internal_model.functions
-            | {"utility": internal_model.utility}
-            | internal_model.constraints,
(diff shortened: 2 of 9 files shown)


mj023:
Regarding the transition keys, I'm wondering, if this might be a problem with DAGs, because then the input would be named the same as some of the functions we provide for `concatenate_functions`. We might then have to internally append something to the key then.

hmgaudecker:
Broken record chiming in: Sounds like this is for the "Use `dags.tree`" PR :laughing:

mj023:
I have now also removed the `function_info`, you were right that it is unnecessary. I replaced querying the dataframe with looking directly at the transition functions.

src/lcm/simulation/processing.py
```diff
@@ -62,7 +62,9 @@ def process_simulated_data(
         calculated_targets = _compute_targets(
             out,
             targets=additional_targets,
-            model_functions=internal_model.functions,
+            model_functions=internal_model.functions
+            | {"utility": internal_model.utility}
+            | internal_model.constraints,
```
timmens: Can you define `model_functions` as a variable before?

src/lcm/simulation/simulate.py
```diff
@@ -160,9 +160,11 @@ def simulate(
         # Update states
         # ------------------------------------------------------------------------------
         if not is_last_period:
-            stochastic_next_function_names = internal_model.function_info.query(
-                "is_stochastic_next"
-            ).index.tolist()
+            stochastic_next_function_names = [
+                key
```
timmens: - key -> next_fn_name
- ... `if is_stochastic_transition(next_fn)`

src/lcm/model_initialization.py
```diff
@@ -35,6 +35,7 @@ def initialize_model_components(model: Model) -> None:
         ModelInitilizationError: If initialization fails
 
     """
+    components = _get_model_components(model)
```
timmens: Needs to be deleted.

tests/model_mock.py
timmens: There is a pre-commit error with this file because it is in the tests directory but not named test_.... This needs to be ignored.

PR #166, opened 2025-10-30
Support multiple regimes per model

mj023:
With this PR we add support for models that are composed of multiple regimes.

### DIFF
--- a/src/lcm/Q_and_F.py
+++ b/src/lcm/Q_and_F.py
@@ -218,7 +218,7 @@ def get_Q_and_F_non_terminal(
         # Calculate the instantaneous utility and feasibility
         # ------------------------------------------------------------------------------
 
-        return jnp.asarray(Q_arr), jnp.asarray(F_arr)
+        return Q_arr, F_arr
 
     return Q_and_F
 
@@ -279,7 +279,7 @@ def get_Q_and_F_terminal(
             params=params[regime.name],
         )
 
-        return jnp.asarray(U_arr), jnp.asarray(F_arr)
+        return U_arr, F_arr
 
     return Q_and_F
 
--- a/src/lcm/input_processing/regime_components.py
+++ b/src/lcm/input_processing/regime_components.py
@@ -276,6 +276,6 @@ def build_regime_transition_probs_functions(
     )
 
     return PhaseVariantContainer(
-        solve=next_regime,
+        solve=jax.jit(next_regime) if enable_jit else next_regime,
         simulate=jax.jit(next_regime_vmapped) if enable_jit else next_regime_vmapped,
     )
--- a/src/lcm/max_Q_over_c.py
+++ b/src/lcm/max_Q_over_c.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import functools
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING
 
 import jax.numpy as jnp
 from jax import Array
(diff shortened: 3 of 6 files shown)


src/lcm/model.py
```diff
@@ -42,12 +42,12 @@ class Model:
     description: str | None = None
     n_periods: int
     enable_jit: bool = True
-    regimes: Regime
-    internal_regime: InternalRegime
+    regimes: dict[Regime]
```
timmens:
```suggestion
    regimes: dict[str, Regime]
```

src/lcm/Q_and_F.py
```diff
@@ -214,7 +247,7 @@ def get_Q_and_F_terminal(
     U_and_F = _get_U_and_F(internal_functions)
 
     arg_names_of_Q_and_F = _get_arg_names_of_Q_and_F(
-        [U_and_F],
+        [U_and_F, *list(internal_functions.constraints.values())],
```
timmens: Why is this required? Aren't the constraints in the F-part of U_and_F?

src/lcm/model.py
```diff
@@ -62,19 +62,21 @@ def __init__(
...
+        self.internal_regimes = {}
+
+        for regime in regimes:
+            _validate_model_inputs(
+                n_periods=n_periods,
+                regime=regime,
+            )
+            self.regimes[regime.name] = regime
+            self.internal_regimes[regime.name] = process_regime(
+                regime=regime, n_periods=n_periods, enable_jit=enable_jit
```
timmens: Don't we have to update the regime processing to include all regimes? I thought that we need to do the processing of certain regime components first, as we then need to pass the processed components of all regimes into each regime? I.e., the Q-and-F function of a regime depends on all of the other regimes?

src/lcm/input_processing/regime_components.py
```diff
@@ -202,4 +227,55 @@ def build_next_state_simulation_functions(
...
+        func=next_regime_accepting_all,
+        variables=tuple(
+            parameter
+            for parameter in parameters
+            if parameter not in ("period", "params")
+        ),
+    )
+
+    return PhaseVariantContainer(
+        solve=next_regime,
```
timmens: The `next_regime` function for the solve phase should be jitted as well, right?

src/lcm/simulation/simulate.py
```diff
@@ -78,199 +66,206 @@ def simulate(
...
+    states: dict[str, Array],
+    subject_regime_ids: Array,
+    new_subject_regime_ids: Array,
+    V_arr_dict: dict[int, dict[RegimeName, FloatND]],
+    params: dict[RegimeName, ParamsDict],
+    regime_name_to_id: dict[RegimeName, int],
+    key: Array,
+    *,
+    is_last_period: bool,
+) -> tuple[SimulationResults, dict[str, Array], Array, Array]:
```
timmens: And similarly of the output type hints.

src/lcm/simulation/util.py
```diff
@@ -0,0 +1,254 @@
...
+
+def calculate_next_states(
+    internal_regime: InternalRegime,
+    optimal_actions: dict[str, Array],
+    period: int,
+    params: dict[RegimeName, ParamsDict],
+    states: dict[str, Array],
+    state_action_space: StateActionSpace,
+    key: Array,
+    subjects_in_regime: Array,
```
timmens: Here we know that it is a `Bool1D` array.

src/lcm/max_Q_over_c.py
timmens: We probably should not cast these types here but rather update the protocols if required.

timmens (commented):
Very nice PR. Some of the comments are probably due to things I've changed. In this case, just tell me and I take care of it.

Most of my comments are about type annotations. While we definitely require massive refactoring, I do not want to do this in the current PR.

PR #169, opened 2025-11-19
Add new example model

mj023:
This PR adds a new example model based on the paper 'Lifestyle Behaviors and Wealth-Health Gaps in Germany' by Lukas Mahler and Minchul Yum (2024, https://doi.org/10.3982/ECTA20603).

### DIFF
--- a/examples/mahler_yum_2024/model.py
+++ b/examples/mahler_yum_2024/model.py
@@ -494,7 +494,7 @@ START_PARAMS = {
 # ======================================================================================
 
 
-def create_phigrid(nu: list[Float1D]) -> FloatND:
+def create_phigrid(nu: dict[str, list[float]]) -> FloatND:
     phi_interp_values = jnp.array([1, 8, 13, 20])
     phigrid = jnp.zeros((retirement_age + 1, 2, 2))
     health = ["u", "h"]
@@ -502,15 +502,19 @@ def create_phigrid(nu: list[Float1D]) -> FloatND:
         for j in range(2):
             interp_points = jnp.arange(1, retirement_age + 2)
             spline = interp1d(
-                np.asarray(phi_interp_values), np.asarray(nu[health[j]]), kind="cubic"
+                np.asarray(phi_interp_values),
+                np.asarray(nu[health[j]]),
+                kind="cubic",
             )
             temp_grid = jnp.asarray(spline(interp_points))
-            temp_grid = jnp.where(i == 0, temp_grid * jnp.exp(nu["ad"]), temp_grid)
+            temp_grid = jnp.where(
+                i == 0, temp_grid * jnp.exp(jnp.array(nu["ad"])), temp_grid
+            )
             phigrid = phigrid.at[:, i, j].set(temp_grid)
     return phigrid
 
 
-def create_xigrid(xi: list[list[Float1D]]) -> FloatND:
+def create_xigrid(xi: dict[str, dict[str, list[float]]]) -> FloatND:
     xi_interp_values = jnp.array([1, 12, 20, 31])
     xigrid = jnp.zeros((n, 2, 2))
     edu = ["hs", "cl"]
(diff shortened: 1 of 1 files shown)


mj023:
I tried to add `interpax`, but it broke some of our dependencies. I just switched to the SciPy Interpolator, it does gives the same values.

examples/long_running/README.md
```diff
@@ -1,14 +1,14 @@
-# Example model specifications
+# Example model specification
 
-## Choosing an example
+## Example Stats
```
hmgaudecker: It does not make much sense to have a table with one entry. I would keep the old structure, but move it up by one level.

Further helicopter-level notes:
- harmonise the way how the models are run (probably keep a `simulate_model.py` (renamed `run_model.py`) in each of them?
- Rename `long_running` to something more descriptive. At least I'd expect `mahler_yum_2024` to run even longer?
- Can we access `examples` when running the tests? In that case, we should move the test models over here.

examples/mahler_yum_2024/Mahler_Yum_2024.py
hmgaudecker: Name should be `mahler_yum_2024.py`

examples/mahler_yum_2024/Mahler_Yum_2024.py
```diff
@@ -0,0 +1,544 @@
...
+    states={"dead": DiscreteGrid(Dead)},
+    actions={},
+    transitions={"dead": {"next_dead": lambda dead: Dead.dead}},  # noqa: ARG005
+    regime_transition_probs=lambda: {"alive": 0.0, "dead": 1.0},
+)
+
+MAHLER_YUM_MODEL = Model([ALIVE_REGIME, DEAD_REGIME], n_periods=n)
+
+
+########################
```
hmgaudecker: Not crucial for now, but I'd really like us to use the possibilities of pytrees here. This would be much clearer if it were in a hierarchical structure, like:

```py
params = {
    nu: {
        ad: 0.8,
        h: jnp.array([2.63, 1.66, 1.27, 1.71]),
        u: jnp.array([2.41, 1.18, 1.39, 2.41]),
    },
    ...
}
```

src/lcm/simulation/processing.py
```diff
@@ -45,7 +45,7 @@ def process_simulated_data(
     """
     n_initial_states = len(results[0].V_arr)
 
-    nan_array = jnp.full(n_initial_states, jnp.nan, dtype=jnp.float64)
+    nan_array = jnp.full(n_initial_states, jnp.nan, dtype=jnp.float32)
```
timmens: Why don't we use `float64` here anymore?

src/lcm/Q_and_F.py
```diff
@@ -218,7 +218,7 @@ def Q_and_F(
         # Calculate the instantaneous utility and feasibility
         # ------------------------------------------------------------------------------
 
-        return Q_arr, F_arr
+        return jnp.asarray(Q_arr), jnp.asarray(F_arr)
```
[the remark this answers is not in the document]
mj023: Because in cases where you only have one state, the vmap will return a value of type `float` or `bool`. This then clashes with some of the later operations, that are only possible on real arrays.

src/lcm/simulation/processing.py
mj023: This throws a warning similar to this: 

``` 
UserWarning: Explicitly requested dtype float64 requested in arange is not available, and will be 
truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the 
JAX_ENABLE_X64 shell environment variable. See https://github.com/jax-ml/jax#current-gotchas for more.
```

Maybe it would be best to avoid specifiying the dtype, then the user can control it with `JAX_ENABLE_X64`.

src/lcm/Q_and_F.py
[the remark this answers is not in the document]
timmens: Good point, I'd say we discuss this on Wednesday. Max and I already discussed other ways of getting rid of this terminal vs. non-terminal logic.

src/lcm/simulation/processing.py
timmens: > We definitely should avoid hard-coding dtypes like this

**We don't do that internally**. The instance you see above was in the processing of the simulation results and there is no other in the codebase.

> Maybe it would be best to avoid specifiying the dtype, then the user can control it with `JAX_ENABLE_X64`

I fully agree and once we have a documentation / how-to guides we should mention this. But I'd go with
```python
import jax
jax.config.update('jax_enable_x64', True)
```

> My question would be whether that global setting would be fine-grained enough? Put differently, is the value function / utility the only thing with a significant memory footprint where the Model does not control the dtype?

I believe it would be fine-grained enough for now. The main issue is the memory footprint during the value function array computation. One thing that should be doable, I think, is to have different dtypes for different regimes.

---

In general though, I think we should discuss this later (also with @mj023), since I believe that GPUs are a magnitude faster when working with float32. So this should be the recommended default?!

PR #170, opened 2025-11-24
Add shock discretization functionality

mj023:
This PR will add a way to specify multiple common shock types directly through the interface of LCM.

- [ ] Add shock discretization functions
- [ ] Integrate shocks into the current interface


timmens:
Is continued in #206

PR #190, opened 2025-12-06
Implement flat transitions interface for user-facing API

timmens:
In this PR, we simplify the user-facing interface by changing from nested to flat format for both transitions and initial states.

### Previously

Transitions were specified in a nested format where each regime had its own transition functions:

```python
transitions={
    "work": {"next_wealth": next_wealth, "next_health": next_health},
    "retirement": {"next_wealth": next_wealth, "next_health": next_health},
    "next_regime": next_regime_fn,
}
```

### Now

Transitions use a flat format where the same function is used for all target regimes:

```python
transitions={
    "next_wealth": next_wealth,
    "next_health": next_health,
    "next_regime": next_regime_fn,
}
```

### Previously

Initial states were specified in a nested format by regime:

```python
initial_states={
    "work": {"wealth": wealth_arr, "health": health_arr},
    "retirement": {"wealth": wealth_arr, "health": health_arr},
}
```

### Now

Initial states use a flat format:

```python
initial_states={
    "wealth": wealth_arr,
    "health": health_arr,
}
```

The system automatically maps each state to the regimes that use it. Validation ensures all required states are provided and all arrays have the same length.

### DIFF
--- a/tests/input_processing/test_create_params_template.py
+++ b/tests/input_processing/test_create_params_template.py
@@ -30,7 +30,6 @@ def test_create_params_without_shocks(binary_category_class):
         grids={regime.name: get_grids(regime)},  # type: ignore[arg-type]
         n_periods=3,
     )
-    # With flat transitions, param keys are flat (no regime prefix)
     assert got == {
         "beta": jnp.nan,
         "utility": {"c": jnp.nan},
--- a/tests/test_solution_on_toy_model.py
+++ b/tests/test_solution_on_toy_model.py
@@ -424,7 +424,6 @@ def test_stochastic_solve(beta, n_wealth_points, health_transition):
     model = Model([STOCHASTIC_REGIME.replace(states=new_states)], n_periods=2)
 
     # Solve model using LCM
-    # With flat transitions, param keys are flat (no regime prefix)
     # ==================================================================================
     params = {
         "beta": beta,
@@ -473,7 +472,6 @@ def test_stochastic_simulate(beta, n_wealth_points, health_transition):
     model = Model([STOCHASTIC_REGIME.replace(states=new_states)], n_periods=2)
 
     # Simulate model using LCM
-    # With flat transitions, param keys are flat (no regime prefix)
     # ==================================================================================
     params = {
         "beta": beta,


tests/input_processing/test_create_params_template.py
```diff
@@ -85,8 +101,9 @@ def next_a(a, a_transition):
         index=["a"],
     )
 
+    # _create_stochastic_transition_params expects nested format internally
     regime = RegimeMock(
-        transitions={"mock": {"next_a": next_a}},
+        transitions={"mock": {"next_a": next_a}},  # type: ignore[dict-item]
```
timmens: Why are the user-facing transitions for the `RegimeMock` class still nested? This should not be the case.

tests/input_processing/test_regime_processing.py
timmens: This test file is great, but you need to account for the test cases where there are multiple target regimes, and then test the major cases (shared state-space, overlapping state space, independent state space over the two regimes).

timmens (commented):
Going in the right direction, but the major thing that is missing is that transitions from one regime must contain transitions to the state space to all other target regimes, unless its an absorbing regime.

Ideally, to test this more efficiently, we should also extend the test case in test_multi_regime.py so that the state-spaces in the two regimes are overlapping, but neither of them is a subset of the other.

examples/mahler_yum_2024/model.py
```diff
@@ -398,18 +398,15 @@ def savings_constraint(
...
-            "next_health": next_health,
-            "next_productivity_shock": next_productivity_shock,
-            "next_discount_factor": next_discount_factor,
-            "next_adjustment_cost": next_adjustment_cost,
-            "next_effort_t_1": next_effort_t_1,
-            "next_health_type": next_health_type,
-            "next_education": next_education,
-            "next_productivity": next_productivity,
-        },
-        "dead": {"next_dead": lambda: Dead.dead},
```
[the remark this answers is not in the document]
timmens: Fixed - the `next_dead` transition is now present in the alive regime. The file was restored to include it.

src/lcm/input_processing/create_params_template.py
```diff
@@ -9,17 +9,18 @@
...
     import pandas as pd
 
     from lcm.regime import Regime
-    from lcm.typing import GridsDict, ParamsDict
+    from lcm.typing import GridsDict, ParamsDict, UserFunction
 
 
 def create_params_template(
     regime: Regime,
+    nested_transitions: dict[str, dict[str, UserFunction] | UserFunction],
```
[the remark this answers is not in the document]
timmens: Acknowledged - this is a cleanup item. The `nested_transitions` parameter and related unused params (`grids`, `n_periods`) were likely intended for `_create_stochastic_transition_params` but the call was never added. Will address in a follow-up cleanup.

src/lcm/input_processing/regime_processing.py
```diff
@@ -96,6 +96,18 @@ def _default_next_regime() -> Array:
...
 
+    # ----------------------------------------------------------------------------------
+    # Convert flat transitions to nested format
+    # ----------------------------------------------------------------------------------
+    # User provides flat format, internal processing uses nested format.
+    nested_transitions = {
+        regime.name: convert_flat_to_nested_transitions(
+            flat_transitions=regime.transitions,
+            current_regime_name=regime.name,
+        )
```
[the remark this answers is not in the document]
timmens: Fixed in commit faee971. The function now takes `states_per_regime: dict[str, set[str]]` and only includes target regimes where ALL states have transition functions. This handles cases where one regime cannot transition to another (e.g., old cannot go back to young because it lacks `next_education`).

tests/input_processing/test_create_params_template.py
timmens: The `RegimeMock` class uses nested format here because `_create_stochastic_transition_params` specifically expects nested format - it iterates over `regime.transitions.items()` expecting regime names as keys. These tests call `_create_stochastic_transition_params` directly (not through the normal processing pipeline), so they need to provide the format that function expects. The function is not called through the normal `create_params_template` path.

tests/input_processing/test_regime_processing.py
timmens: Added in commit faee971:
- `test_convert_flat_to_nested_multi_regime_complete`: Tests case with overlapping states (young: wealth/education, old: wealth/pension) where both regimes have complete transitions
- `test_convert_flat_to_nested_multi_regime_incomplete`: Tests case where one regime has incomplete transitions and gets excluded

Also added `test_multi_regime_with_overlapping_states` in test_multi_regime.py as an integration test.

PR #191, opened 2025-12-11
Add terminal regime support with active periods

timmens:
In this PR, we introduce terminal regimes and active periods, enabling explicit modeling of absorbing states and period-specific regime activation.

### Previously

There was no explicit concept of a terminal regime. Users had to manually handle the last period's value function logic, and there was no way to specify that a regime should only be active during certain periods of the model.

```python
# No explicit terminal handling
Regime(
    name="retirement",
    utility=utility_retired,
    transitions={"next_wealth": next_wealth, "next_regime": ...},  # Required even for last period
    states={"wealth": wealth_grid},
)
```

### Now

We introduce two new `Regime` attributes:

1. **`terminal`**: Marks a regime as terminal (absorbing). Terminal regimes:
   - Cannot have transitions (no `next_*` functions needed)
   - Must have at least one state
   - The utility function serves as the terminal value function

2. **`active`**: Specifies which periods a regime is active in (required for all regimes)

```python
# Non-terminal regime with explicit active periods
Regime(
    name="working",
    active=range(n_periods - 1),  # Active in all periods except last
    utility=utility,
    transitions={"next_wealth": next_wealth, "next_regime": next_regime},
    states={"wealth": wealth_grid, "health": health_grid},
    actions={"consumption": consumption_grid},
)

# Terminal regime - no transitions needed
Regime(
    name="retirement",
    active=[n_periods - 1],  # Only active in the last period
    terminal=True,
    utility=utility_retired,  # Acts as terminal value function
    states={"wealth": wealth_grid, "health": health_grid},
)
```

### Model Validation

The `Model` class now enforces:
- At least one terminal regime must be provided
- At least one non-terminal regime must be provided
- Non-terminal regimes must include `next_regime` in their transitions
- Non-terminal regimes must have transitions for ALL states across ALL regimes (since they can potentially transition to any regime)

### DIFF
--- a/src/lcm/simulation/util.py
+++ b/src/lcm/simulation/util.py
@@ -29,8 +29,6 @@ def get_regime_name_to_id_mapping(regime_id_cls: type) -> dict[RegimeName, int]:
 def create_regime_state_action_space(
     internal_regime: InternalRegime,
     states: dict[str, Array],
-    *,
-    is_last_period: bool,
 ) -> StateActionSpace:
     """Create the state-action space containing only the relevant subjects in a regime.
 
@@ -38,16 +36,12 @@ def create_regime_state_action_space(
         internal_regime: The internal regime instance.
         states: The current states of all subjects.
         subject_ids_in_regime: Indices of subjects in the current regime.
-        is_last_period: Whether we are in the last period of the model.
 
     Returns:
         The state-action space for the subjects in the regime.
 
     """
-    if is_last_period:
-        query = "is_state and enters_concurrent_valuation"
-    else:
-        query = "is_state and (enters_concurrent_valuation | enters_transition)"
+    query = "is_state and (enters_concurrent_valuation | enters_transition)"
 
     relevant_states_names = internal_regime.variable_info.query(query).index
 
@@ -59,7 +53,6 @@ def create_regime_state_action_space(
         variable_info=internal_regime.variable_info,
         grids=internal_regime.grids,
         states=states_for_state_action_space,
-        is_last_period=is_last_period,
     )
 
 
--- a/examples/consumption_saving/model.py
+++ b/examples/consumption_saving/model.py
@@ -64,7 +64,7 @@ def utility_retired(
     wealth: ContinuousState,
     health: ContinuousState,
 ) -> FloatND:
-    return jnp.log(wealth) * health
+    return jnp.array([jnp.log(wealth) * health])
 
 
 # --------------------------------------------------------------------------------------
--- a/src/lcm/input_processing/regime_components.py
+++ b/src/lcm/input_processing/regime_components.py
@@ -60,7 +60,6 @@ def build_state_action_space(
     return create_state_action_space(
         variable_info=variable_info,
         grids=grids,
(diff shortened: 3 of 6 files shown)


src/lcm/Q_and_F.py
```diff
@@ -28,60 +28,28 @@
...
         RegimeName,
     )
 
 
 def get_Q_and_F(
     regime: Regime,
+    active_periods: dict[str, list[int]],
     internal_functions: InternalFunctions,
     next_state_space_infos: dict[str, StateSpaceInfo],
     grids: dict[RegimeName, Any],
```
hmgaudecker: Changed the order; please adjust like this elsewhere, too. (thinking of `period` like a state)

src/lcm/input_processing/regime_processing.py
```diff
@@ -211,6 +174,25 @@ def _default_next_regime() -> Array:
...
+    if regime.terminal and n_periods - 1 not in regime.active:
+        return [n_periods - 1, *list(regime.active)]
+
+    if not regime.terminal and n_periods - 1 in regime.active:
+        return [p for p in regime.active if p == n_periods - 1]
+
+    return list(regime.active)
+
+
+def _get_default_active_periods(regime: Regime, n_periods: int) -> list[int]:
```
hmgaudecker: I'd get rid of this. It is not obvious this is a sensible default. E.g., one could argue for terminal just being active in the last period. And any default stops making sense once you have multiple terminal regimes.

src/lcm/simulation/util.py
```diff
@@ -276,7 +278,7 @@ def validate_flat_initial_states(
         internal_regimes: Dict of internal regime instances.
 
     Raises:
-        ValueError: If validation fails with descriptive message.
+        InvalidInitialStatesError: If validation fails with descriptive message.
```
hmgaudecker:
```suggestion
        InvalidInitialStatesError: If validation fails.
```

hmgaudecker (commented):
Very nice! Did not manage to look at the tests.

tests/simulation/test_initial_states.py
```diff
@@ -1,66 +1,130 @@
...
+
+    def next_regime(period: int) -> ScalarInt:
+        return jnp.where(
+            period + 1 >= 2,
+            RegimeID.terminal,
+            RegimeID.active,
+        )
+
+    alive = Regime(
+        name="active",
```
hmgaudecker:
```suggestion
        name="alive",
```

tests/simulation/test_initial_states.py
```diff
@@ -1,66 +1,130 @@
...
+        [alive, dead],
+        n_periods=2,
+        regime_id_cls=RegimeID,
+    )
+
+
+# ==============================================================================
+# Tests
+# ==============================================================================
+def test_convert_flat_to_nested_single_regime(model: Model) -> None:
```
hmgaudecker: Name seems confusing now that we only ever have models with min 2 regimes. *(unless you think that supporting the static case as a special one is worth it, but then this would definitely be a misnomer)*

PR #198, opened 2025-12-22
Inject temporal context into model functions

timmens:
## Add `Time` object for temporal context

User functions now receive a `Time` object instead of separate `period`, `n_periods`, `last_period` arguments:

```python
def next_regime(time: Time) -> int:
    forced_retirement = time.period == time.last_period - 1
    return jnp.where(forced_retirement, RegimeId.retirement, RegimeId.working)
```

### Why this is better

1. **Single point of extension**: Adding new temporal variables (e.g., `periods_remaining`) requires changes in one place only - the `Time` dataclass.

1. **Self-documenting**: Users discover available temporal info through IDE autocomplete on `time.`, rather than checking documentation for magic variable names.

1. **Type safety**: `time: Time` in function signatures is explicit about what's being received, vs implicit `period: int` that could be confused with other integers.

The `last_period` property is computed from `n_periods`, eliminating redundant storage and ensuring consistency.

### DIFF
--- a/src/lcm/typing.py
+++ b/src/lcm/typing.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import Any, Protocol
 
+import jax
 from jax import Array
 from jaxtyping import Bool, Float, Int, Scalar
 
@@ -26,6 +27,14 @@ class Time:
     def last_period(self) -> int:
         """Index of the last period (n_periods - 1)."""
         return self.n_periods - 1
+
+
+# Register Time as a JAX pytree so it can be passed through JIT-compiled functions
+jax.tree_util.register_dataclass(
+    Time,
+    data_fields=["period", "n_periods"],
+    meta_fields=[],
+)
 
 
 type ContinuousState = Float[Array, "..."]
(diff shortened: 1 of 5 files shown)


hmgaudecker:
Just occurred to me: Would it make sense to take that thought further and call it `AgeGrid` with obvious contents?

We would get rid of the period / age distinction that way, like we do in other cases. Age would be the grid value, period the grid index.

timmens:
After talking on the phone, we decided to:

- Only support/include  `period` and `age` as "special" context in the model functions. That is, after the change, users will be able to access the current `period` and the current `age` in their functions. The `period` will represent the index of an age grid. Things like `last_period` should be easy enough to figure out by the user herself.

- Add an `ages` attribute to `lcm.Model`, where the user specifies which ages to "loop over". E.g. `ages = lcm.AgeGrid(start=18, end=70, step=2)`.
   
   This also gives us the possibility to allow for "custom" age grids in the future, for example through `ages = lcm.AgeGrid(custom=[18, 30, 50, 60, 62, 65])`. In this case, the class would print a warning however, since we do not want to adjust any future-weighting conditional on the time-interval-length.
   
   This will also remove the necessity to supply `n_periods` to the model, since we, effectively, only need to loop over `ages`.

timmens:
Thanks to the both of you. Development of this feature continues at #203

PR #200, opened 2025-12-25
Normalize regime transition probabilities

timmens:
In this PR, we add normalization of regime transition probabilities, which is necessary because otherwise agents assign probability to inactive regimes.

### DIFF
--- a/tests/test_normalize_regime_probs.py
+++ b/tests/test_normalize_regime_probs.py
@@ -1,5 +1,8 @@
 import jax.numpy as jnp
+import pytest
 
+from lcm.exceptions import InvalidRegimeTransitionProbabilitiesError
+from lcm.simulation.util import _validate_normalized_regime_transition_probs
 from lcm.utils import normalize_regime_transition_probs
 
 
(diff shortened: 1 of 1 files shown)


src/lcm/error_handling.py
```diff
@@ -24,5 +24,9 @@ def validate_value_function_array(
     """
     if jnp.any(jnp.isnan(V_arr)):
         raise InvalidValueFunctionError(
-            f"The value function array in period {period} contains NaN values."
+            f"The value function array in period {period} contains NaN values. This "
+            "may be due to various reasons:\n"
+            "- The user defined functions returned invalid values.\n"
+            "- It is impossible to reach an active regime, resulting in NaN regime "
+            "transition probabilities in the normalized transition probabilities."
         )
```
hmgaudecker:
```suggestion
            "- The user-defined functions returned invalid values.\n"
            "- It is impossible to reach an active regime, resulting in NaN regime\n"
            "  transition probabilities in the normalized transition probabilities."
        )
```

hmgaudecker (approved):
Would be great to test the exceptions, too. But no show-stopper!

PR #201, opened 2025-12-29
Add `to_pickle` and `from_pickle` methods to `SimulationResult` object

timmens:
In this PR, we add the `to_pickle` and `from_pickle` methods to `SimulationResult` object.

Saving an entire simulation result object can be important in projects. This PR adds the functionality to save and load such objects via the `cloudpickle` library.

### Implementation

- `to_pickle`: To write the file we use an atomic strategy (all-or-nothing). We do this to keep an existing file intact until the new file is completely written to avoid a state where writing failed and the user ends up with only a corrupted file.
- `from_pickle`: This is a straightforward pickle read.

### DIFF
--- a/src/lcm/Q_and_F.py
+++ b/src/lcm/Q_and_F.py
@@ -14,6 +14,7 @@ from lcm.functools import get_union_of_arguments
 from lcm.input_processing.util import is_stochastic_transition
 from lcm.interfaces import InternalFunctions, Target
 from lcm.next_state import get_next_state_function, get_next_stochastic_weights_function
+from lcm.utils import normalize_regime_transition_probs
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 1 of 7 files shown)


src/lcm/simulation/result.py
```diff
@@ -570,3 +629,48 @@ def _create_target_function(
...
+
+    # Verify parent directory exists
+    p = Path(path)
+    parent = p.parent
+    if parent and not parent.is_dir():
+        raise FileNotFoundError(f"Parent directory does not exist: {parent}")
+
+    # Write to a temporary file first, then replace the target. This avoids leaving a
+    # corrupted file at `path` behind if writing fails mid-way.
+    tmp = p.with_suffix(p.suffix + ".tmp")
```
hmgaudecker: Just use Python's tempfile mechanism here, saves us the manual cleanup.

PR #203, opened 2026-01-02
Add age grid

timmens:
In this PR we introduce a model-wide age grid.

- Adds `ages` argument to `Model` init.
- Ages must be an `AgeGrid` object that can be specified using a range-like syntax or custom values.
   - The range-like syntax allows to specify steps as time-sensible deltas, e.g., `"Y"` for year, `"3M"` for three months, or `"Q"` for quarterly. These will then be transformed into year-level fractions.
- Since float-ages do not work well with equality comparisons, the `active` attribute of `Regimes` had to be changed to be `Callable[[float], bool]`, which returns `True` if the regime is active during an `age`.

### DIFF
--- a/examples/mahler_yum_2024/model.py
+++ b/examples/mahler_yum_2024/model.py
@@ -7,6 +7,7 @@ Wealth-Health Gaps in Germany" by Lukas Mahler and Minchul Yum (Econometrica, 20
 from __future__ import annotations
 
 from dataclasses import dataclass, make_dataclass
+from functools import partial
 from typing import TYPE_CHECKING
 
 import jax
@@ -43,7 +44,8 @@ if TYPE_CHECKING:
 # --------------------------------------------------------------------------------------
 avrgearn_not_normalized: float = 57706.57
 theta_val: Float1D = jnp.array([jnp.exp(-0.2898), jnp.exp(0.2898)])
-n: int = 38
+ages = AgeGrid(start=25, stop=101, step="2Y")
+n: int = ages.n_periods
 retirement_age: int = 19
 taul: float = 0.128
 lamda: float = 1.0 - 0.321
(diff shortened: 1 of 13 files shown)


src/lcm/ages.py
```diff
@@ -0,0 +1,203 @@
...
+            IndexError: If period is out of bounds.
+
+        """
+        if period < 0 or period >= self.n_periods:
+            raise IndexError(
+                f"Period {period} out of bounds for grid with {self.n_periods} periods."
+            )
+        return float(self._ages[period])
+
+    def get_periods_where(self, predicate: Callable[[float], bool]) -> list[int]:
```
hmgaudecker:
```suggestion
    def get_periods_where(self, predicate: Callable[[float], bool]) -> tuple[int, ...]:
```

Would prefer an immutable object?

examples/mahler_yum_2024/model.py
```diff
@@ -397,18 +397,20 @@ def savings_constraint(
...
     name="dead",
     terminal=True,
     utility=lambda: 0.0,
-    active=[n - 1],
+    active=lambda age: age >= n - 1,
 )
 
 MAHLER_YUM_MODEL = Model(
-    [ALIVE_REGIME, DEAD_REGIME], n_periods=n, regime_id_cls=RegimeId
+    [ALIVE_REGIME, DEAD_REGIME],
```
hmgaudecker: I think we should make `Model` keyword-only.

src/lcm/typing.py
```diff
@@ -49,7 +50,7 @@ class InternalUserFunction(Protocol):
     """
 
     def __call__(  # noqa: D102
-        self, *args: Array | int, params: ParamsDict, **kwargs: Array | int
+        self, *args: Array | float, params: ParamsDict, **kwargs: Array | float
```
hmgaudecker: Does `float` imply `float|int` or can we not pass period any more?

hmgaudecker (commented):
Very nice!!! Thanks a lot.

src/lcm/ages.py
hmgaudecker: Implemented in #204.

PR #206, opened 2026-01-06
Add shock discretization functionality

mj023:
This PR is duplicate of #170. Merging the previous PR with the main branch became unfeasible because there were too many merge conflicts.

This PR will add a way to specify multiple common shock types directly through the interface of LCM.

### New Classes
- Class `ShockGrids`
    - Used to specify some default shock types for discretization with LCM
    - On initialization creates a `Shock`
    - Params can be supplied later through the fixed parameters, or directly to the `ShockGrid`class (This allows the user to use the gridpoints and transition probabilities independently from the model)
    - All shocks have default parameters that cant be changed, e.g. $\sigma = 1$ ( => the user has to scale the shock himself)
    - If extra paremeters are needed (eg. $\rho$ for AR1-shocks), they always also have to be supplied via the models fixed params
 - Class `Shock`
     - Attribute of `ShockGrid` 
     - Interface to the collections of shock discretization functions

#### Regime initialization
A weights function that returns the shocks transition probabilities is added. A function that returns the possible values in the next period (simply the discretization points) is added.
#### Solution
`ShockGrids` act similar to discrete grids, the expectation is calculated over the discretized shock grid in the next period,  weighted with the transition probabilities.
#### Simulation
The values of the shocks in the next period are drawn from the continuous distribution of the shock. To find the optimal actions at shock values that are not on the discretization grid, a linear interpolation with `map_coordinates` is used.

### DIFF
--- a/src/lcm/simulation/result.py
+++ b/src/lcm/simulation/result.py
@@ -18,8 +18,10 @@ from lcm.ages import AgeGrid
 from lcm.dispatchers import vmap_1d
 from lcm.exceptions import InvalidAdditionalTargetsError
 from lcm.grids import DiscreteGrid
+from lcm.input_processing.util import is_stochastic_transition
 from lcm.interfaces import InternalRegime, PeriodRegimeSimulationData
 from lcm.typing import FloatND, ParamsDict, RegimeName
+from lcm.utils import flatten_regime_namespace
 
 CLOUDPICKLE_IMPORT_ERROR_MSG = (
     "Pickling SimulationResult objects requires the optional dependency 'cloudpickle'. "
@@ -349,10 +351,27 @@ def _collect_all_available_targets(
 
 def _get_available_targets_for_regime(regime: InternalRegime) -> set[str]:
     """Get available target names for a single regime."""
+    stochastic_weight_names = _get_stochastic_weight_function_names(regime)
     targets = {"utility"}
-    targets.update(regime.functions.keys())
+    targets.update(
+        name for name in regime.functions if name not in stochastic_weight_names
+    )
     targets.update(regime.constraints.keys())
     return targets
+
+
+def _get_stochastic_weight_function_names(regime: InternalRegime) -> set[str]:
+    """Get names of internal stochastic weight functions.
+
+    These are functions named `weight_{transition_name}` that return probability arrays
+    for stochastic state transitions. They should not be exposed as available targets.
+    """
+    flat_transitions = flatten_regime_namespace(regime.transitions)
+    return {
+        f"weight_{name}"
+        for name, fn in flat_transitions.items()
+        if is_stochastic_transition(fn)
+    }
 
 
 # ======================================================================================
--- a/tests/simulation/test_simulate.py
+++ b/tests/simulation/test_simulate.py
@@ -421,6 +421,44 @@ def test_additional_targets_all(regression_simulation_result):
     assert set(result.available_targets) <= set(df.columns)
 
 
+def test_additional_targets_all_with_stochastic_transitions():
+    """Test that additional_targets='all' works with stochastic transition models.
+
+    This is a regression test for issue #215: stochastic weight functions
+    (e.g., weight_next_health) were incorrectly included in available_targets,
+    causing additional_targets='all' to fail.
+    """
+    from tests.test_models.stochastic import (  # noqa: PLC0415
+        HealthStatus,
+        PartnerStatus,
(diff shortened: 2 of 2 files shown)


timmens (commented):
Very nice! I have a few general comments. Excited to discuss this soon! 🙂 


 **Bugs**                                                                
                                                                      
  1. ShockGrid not exported - Listed in __all__ but not imported in src/lcm/__init__.py                                                 
                                                                      
  **API Design**                                                          
                                                                      
  2. Inconsistent pass pattern - Built-in shock transitions use `def next_x(): pass` while all other transitions return values. Consider inferring transitions automatically from ShockGrid - if a state has a ShockGrid, generate its transition implicitly (no next_ function needed). This also removes the redundant distribution_type in both grid and decorator. Only custom shocks would need a next function then. Counter argument would be that this breaks the "contract" that each state variable (which a shock is) has a next function.                                   
  3. ShockGrid.get_coordinate breaks interface - Takes extra params argument unlike other ContinuousGrid subclasses (hence the # ty: ignore). Could consider a different abstraction?
                                                                      
  **Implementation**                                                      
                                                                      
  4. ShockGrid.to_jax() returns placeholders - Returns [0, 1, 2, ...] until update_sas_with_shocks fills in real values. This implicit contract could cause subtle bugs.                                   
  5. InternalRegime now mutable - Changed from frozen=True to frozen=False. Could use dataclasses.replace() instead to preserve immutability?                                           
  6. Python loops in discretization - _fill_tauchen and rouwenhorst use Python loops. Could vectorize with jax.vmap?                       
                                                                      
  **Minor**                                                               
                                                                      
  7. Outdated docstring - ShockGrid docstring mentions nonexistent start/stop attributes

hmgaudecker (commented):
This is coming along very nicely, thanks a lot!!! I only looked at the interface so far. Looks neat, though a bit more streamlining could be helpful (maybe documentation is enough).

I also like the `fixed_params` attribute as a start, we'll flesh it out elsewhere then.

Thanks for tracking down those bugs, they sound like they would have become a nightmare!

examples/mahler_yum_2024/model.py
```diff
@@ -684,9 +654,13 @@ def create_inputs(
     sigma: int,
 ) -> tuple[dict[RegimeName, Any], dict[RegimeName, Any], list[RegimeName]]:
     # Create variable grids from supplied parameters
-    income_grid = create_income_grid(income_process)
+    income_grid = create_income_grid(income_process)  # ty: ignore[invalid-argument-type]
     chimax_grid = create_chimaxgrid(chi)
-    xvalues, xtrans = rouwenhorst(rho, jnp.sqrt(income_process["sigx"]), 5)  # ty: ignore[invalid-argument-type]
+    prod_shock = Shock(
```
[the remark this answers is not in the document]
mj023: This is mostly caused by the new `fixed_param`. The ShockGrid can only be fully initialized when the Model gets created, because it's missing the params. The only other way to do this would be to repeat the params, eg when creating the ShockGrid or the Shock and then in the fixed params. Right now the Shock Class is therefore only a streamlined way to access the internal shock discretization functions. Therefore the repetition can at least be avoided if one does not need the Shock outside the model. 

The next_state function would also optimally be eliminated for Shock states, but this currently clashes with the way we process transitions. This should be part of a different PR that eliminates the duplication of all next_state functions for all regimes.

src/lcm/model.py
```diff
@@ -261,47 +270,31 @@ def _validate_model_inputs(  # noqa: C901
...
+    mising_fixed_params = _validate_fixed_params_present(regimes, fixed_params)
+    if mising_fixed_params:
+        error_messages.extend(mising_fixed_params)
 
     if error_messages:
         msg = format_messages(error_messages)
         raise ModelInitializationError(msg)
 
 
-def _validate_transition_completeness(regimes: Mapping[str, Regime]) -> list[str]:
```
timmens: Why was `_validate_transition_completeness` deleted?

hmgaudecker:
> What the role of fixed_params, why it is necessary now, and how (if) it can help outside of shocks

I don't think we should spend time on this — the serious interface still needs to be spelled out. Happy to chat at any point starting tomorrow afternoon.

src/lcm/model.py
timmens: No, you were absolutely right in deleting it. Thanks!

timmens (approved):

PR #207, opened 2026-01-06
Require Python 3.14, Cuda 13

hmgaudecker:
- Bump Python to 3.14, Cuda 13 (let's take advantage of most recent tech, see below)
- No big changes necessary, just a couple of ty issues that might have been due to numpy 2.4 or updated ty, actually

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -208,7 +208,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/zeromq-4.3.5-h387f397_9.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/zipp-3.23.0-pyhcf101f3_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/zstd-1.5.7-hb78ec9c_6.conda
-      - pypi: git+https://github.com/opensourceeconomics/dags.git?branch=fix-no-annotation-found-mismatch#51b9393ecb5cc68efa92871e5d708cd6de43158a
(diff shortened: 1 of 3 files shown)


src/lcm/simulation/util.py
```diff
@@ -9,12 +10,15 @@
     InvalidRegimeTransitionProbabilitiesError,
 )
 from lcm.input_processing.util import is_stochastic_transition
-from lcm.interfaces import InternalRegime, StateActionSpace
 from lcm.random import generate_simulation_keys
 from lcm.state_action_space import create_state_action_space
-from lcm.typing import Bool1D, Float1D, Int1D, ParamsDict, RegimeName
+from lcm.typing import Bool1D, Float1D, Int1D  # noqa: TC001
```
hmgaudecker: Note: required at runtime in Python 3.14

hmgaudecker:
@mj023, can you briefly check whether the setup in this branch works for you?

timmens:
Looks good to me, very exciting! At some point we should start benchmarking some model to actually measure whether there are performance increases with new (jax/cuda/etc.) versions.

I'd like to wait with merging until we are sure that we can use Cuda13 on all relevant devices.

@MImmesberger, do you know whether Marvin supports Cuda13 this?

hmgaudecker:
> I'd like to wait with merging until we are sure that we can use Cuda13 on all relevant devices.

I tend to disagree. We can work around that one machine (should that be necessary at all) just like we work around on Marvin right now. After all, it only affects the test environment, whereas projects have their own. I cannot imagine anything we would do in the next couple of weeks that would work on Cuda 13 but not on Cuda 12. But maybe that is my ignorance!

PR #208, opened 2026-01-11
Introduce decorator to describe 1d spaces and regime ids. Fixes #192. 

### LINKED ISSUE #192: ENH: Single action sets and state spaces
We currently have constructs like:

```py
@dataclass
class LaborSupply:
    work: int = 0
    retire: int = 1
```

in order to be able to write `LaborSupply.work`. 

Similarly, we have (in #191/ #160):

```py
@dataclass
class RegimeID:
    working: int = 0
    dead: int = 1
```

and 

```py
def next_regime(period: int, n_periods: int) -> ScalarInt:
    certain_retirement = period >= n_periods - 2
    return jnp.where(certain_retirement, RegimeID.retirement, RegimeID.working)
```

The dataclasses really span the space one of these three types of variables can do. That we use that to not write out integers is one thing, but it would be great if we could also use it for type hinting like so:

```py
def next_regime(period: int, n_periods: int) -> RegimeID:
```

We may be able to do that via a custom combination of Enums and classes, syntax would ideally look like:

```py
@lcm.regime_space
class RegimeID:
    working: int
    dead: int
```

```py
@lcm.one_action_set
class LaborSupply:
    work: int
    retire: int


@lcm.one_state_space
class LaggedLaborSupply:
    working: int
    retired: int
```

Whether we can use that to further restrict the type hints for state-related next-functions is a different story, but probably not crucial

hmgaudecker:
This gets rid of having to specify consecutive integers in describing 1d spaces manually.

Simple enough, but I am not perfectly sure about the naming yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,8 @@ jobs:
       - name: Upload coverage report
         if: runner.os == 'Linux' && matrix.python-version == '3.14'
         uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
   run-ty:
     name: Run ty
     runs-on: ubuntu-latest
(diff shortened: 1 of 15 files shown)


hmgaudecker:
Sorry for the mess...

timmens:
1. This seems to be outdated. The dataclass should only allow `ints`. Also: The naming should not be a showstopper. We can always rename afterwards.
2. I remember that I actually proposed to use something similar to `enum.auto()` to generate the indices automatically. Since then I changed my mind, and think that the indices might be helpful after all. But also here, we could go ahead and merge, and if users miss the indices we revert that change.

hmgaudecker:
> I remember that I actually proposed to use something similar to enum.auto() to generate the indices automatically. Since then I changed my mind, and think that the indices might be helpful after all. But also here, we could go ahead and merge, and if users miss the indices we revert that change.

I'd prefer that. We won't have too many outcomes, typically. And it is near-trivial to look at the actual `DiscreteGrid` or a dataclass like `LaborSupply` by just importing file or setting a breakpoint in the user entry code.

Issue #187, opened 2025-12-05
ENH: Use regime transitions with array returns internally


timmens:
In #186 we aligned the user-interface for regime transitions with that of stochastic variable transitions. However, internally we still rely on the `dict[str, float | Array]` interface.

Ideally, we should not need to perform such a conversion but use the Array-return interface also internally. To decide whether this is a better approach we need to:

1. Estimate the resulting code complexity after the change
2. Consider performance trade-offs

PR #214, opened 2026-01-18
Simplify `normalize_regime_transition_probs` and other cleanups

hmgaudecker:
Address some parts of #187. This simplifies the code a bit. I did not benchmark, but I do not see where it could lead to a slowdown.


src/lcm/utils.py
```diff
@@ -50,25 +50,71 @@ def unflatten_regime_namespace(d: dict[str, Any]) -> dict[RegimeName, Any]:
     return unflatten_from_qnames(d)  # ty: ignore[invalid-return-type]
 
 
-@overload
 def normalize_regime_transition_probs(
```
timmens: `normalize_regime_transition_probs` is tested but never used in production code?

PR #223, opened 2026-01-28
Fix: Exclude stochastic weight functions from available_targets

### LINKED ISSUE #215: BUG: Selecting stochastic state transition as `additional_targets` returns `ValueError`
#### Problem description

When selecting additional targets in the `.to_dataframe()` method, lcm lists all model functions as potential targets. But when selecting one of the stochastic state transitions as a target (my expectation was that this would return the realisation of the next shock) I get a `ValueError` in pandas: 

`ValueError: Data must be 1-dimensional, got ndarray of shape (10, 2) instead`

The ValueError occurs because lcm returns the transition probabilities, not the shock realisation in that case. 

Here is a reproducer. Just import `get_model` and `get_params` from the `stochastic.py` test model and run

```python
model = get_model(n_periods=4)
params = get_params(n_periods=4)
result = model.solve_and_simulate(
    params=params,
    initial_states={
        "health": jnp.array([1, 1, 0, 0]),
        "partner": jnp.array([0, 0, 1, 0]),
        "wealth": jnp.array([10.0, 50.0, 30, 80.0]),
    },
    initial_regimes=["working"] * 4,
)
df = result.to_dataframe(additional_targets=["weight_retired__next_health"])
```

timmens:
## Summary

- Exclude internal `weight_{transition_name}` functions from `available_targets` to fix ValueError when calling `to_dataframe(additional_targets="all")` on models with stochastic transitions

## Test plan

- [x] Verified the reproducer from issue #215 no longer crashes

Fixes #215

### DIFF
--- a/src/lcm/simulation/result.py
+++ b/src/lcm/simulation/result.py
@@ -18,6 +18,7 @@ from lcm.ages import AgeGrid
 from lcm.dispatchers import vmap_1d
 from lcm.exceptions import InvalidAdditionalTargetsError
 from lcm.grids import DiscreteGrid
+from lcm.input_processing.util import is_stochastic_transition
 from lcm.interfaces import InternalRegime, PeriodRegimeSimulationData
 from lcm.typing import FloatND, ParamsDict, RegimeName
 from lcm.utils import flatten_regime_namespace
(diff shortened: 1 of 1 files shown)


MImmesberger (approved):
Looks good to me, thanks! Maybe the fact that `additional_targets="all"` didn't work previously warrants a new end-to-end test? Your call, I don't know much about this codebase

PR #231, opened 2026-02-09
Replace `internal_regime_params` with flat qnames and factor out aggregation function `H`

hmgaudecker:
- **`internal_regime_params` removed.** Function parameters are renamed to qualified names
(e.g. `risk_aversion` → `utility__risk_aversion`) via `dags.signature.rename_arguments`
and passed as flat keyword arguments. The old mechanism wrapped every user function to
accept a single `internal_regime_params: MappingProxyType` dict and extract its params
at call time; that indirection is gone.
- `InternalRegimeParams` is now a flat `MappingProxyType[str, bool | float | Array]`
(was nested two levels). `process_params` produces per-regime flat dicts; call sites
use `**internal_params[name]` instead of `internal_regime_params=internal_params[name]`.
- **Aggregation function H factored out.** `Q = U + discount_factor * E[V']` was hardcoded
inline. Now a default `_default_H(utility, continuation_value, discount_factor)` is
injected into non-terminal `Regime.functions`, and `Q_and_F` calls
`H_fn(utility=U_arr, continuation_value=..., **H_kwargs)`. Users can supply a custom H
(e.g. CES/Epstein-Zin) via `functions={"H": my_H}`.
- `H` is excluded from the dags function pool (`_get_U_and_F`, `_build_functions_pool`) to
avoid annotation conflicts. Terminal regimes don't get `H` (`Q` = `U` directly, no
continuation value).
- `create_regime_params_template` uses a restricted exclusion set for `H` (no states/actions)
and raises `InvalidNameError` if an `H` parameter name shadows a state or action.

## Key files

| Area                                | Files                                                                                                                                              |
| ----------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
| Core                                | `Q_and_F.py`, `regime_processing.py`, `params_processing.py`, `create_regime_params_template.py`, `regime_components.py`                           |
| Optimization                        | `max_Q_over_a.py`, `max_Q_over_c.py`, `max_Qc_over_d.py`                                                                                           |
| Solution/sim                        | `solve_brute.py`, `simulate.py`, `util.py`, `result.py`, `next_state.py`                                                                           |
| Model def                           | `regime.py`, `typing.py`                                                                                                                           |
| Tests                               | `test_custom_aggregator.py` (new), plus updates to 9 existing test files                                                                           |
| Examples                            | `mahler_yum_2024/model.py`: renamed state `discount_factor` →`stochastic_discount_factor` to avoid collision with H's `discount_factor` parameter. |
| Dead regime params changed to `{}`. |

## ToDo

- [ ] Partial `fixed_params` upon model creation; `params_template` will only show what is left.
- [ ] Allow passing only `n_points` for `IrregSpacedGrids`, values will then show up in the params template
- [ ] Allow passing only `n_points` for `ShockGrid` (and any set of other parameters that might be required), unspecified parameters will then show up in the params template; values and/or transition probabilities will be computed on the fly.

### DIFF
--- a/src/lcm/input_processing/regime_processing.py
+++ b/src/lcm/input_processing/regime_processing.py
@@ -112,17 +112,15 @@ def process_regimes(
     variable_info = MappingProxyType(
         {n: get_variable_info(r) for n, r in regimes.items()}
     )
-    internal_fixed_params = {
-        n: _extract_regime_fixed_params(
-            regime=r, regime_name=n, fixed_params=fixed_params
-        )
+    shock_init_params = {
+        n: _extract_shock_params(regime=r, regime_name=n, fixed_params=fixed_params)
         for n, r in regimes.items()
     }
     gridspecs = MappingProxyType(
         {
             n: _init_shock_gridspecs(
                 gridspecs=get_gridspecs(r),
-                internal_fixed_params=internal_fixed_params[n],
+                shock_init_params=shock_init_params[n],
             )
             for n, r in regimes.items()
         }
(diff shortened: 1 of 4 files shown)


hmgaudecker:
Will not be merged, making three different PRs out of this one.

hmgaudecker:
Nothing left here that would be worth looking at.

Issue #219, opened 2026-01-23
ENH: Improved flow -> `solve(Model, params)`, `simulate(Model, params, initial_states, solution)`


hmgaudecker:
```mermaid
  flowchart TD                                                                                                                                       
      subgraph "Model.__init__"                                                                                                                      
          A[Regime definitions] --> B[process_regimes]                                                                                               
          B --> C[Build state_action_space<br/>with empty shock grids]                                                                               
          B --> D[Build Q_and_F functions]                                                                                                           
          B --> E[Build max_Q_over_a]                                                                                                                
          C --> F[InternalRegime]                                                                                                                    
          D --> F                                                                                                                                    
          E --> F                                                                                                                                    
      end                                                                                                                                            
                                                                                                                                                     
      subgraph "Model.solve"                                                                                                                         
          G[params] --> H[pre_compute_shock_probs]                                                                                                   
          H --> I[update_sas_with_shocks<br/>⚠️ mutates SAS]                                                                                         
          I --> J[Backward induction]                                                                                                                
          J --> K[V_arr_dict]                                                                                                                        
      end                                                                                                                                            
                                                                                                                                                     
      subgraph "Model.simulate"                                                                                                                      
          L[params] --> M[pre_compute_shock_probs]                                                                                                   
          M --> N[update_sas_with_shocks<br/>⚠️ mutates SAS]                                                                                         
          N --> O[Forward simulation]                                                                                                                
          K --> O                                                                                                                                    
          O --> P[SimulationResult]                                                                                                                  
      end                                                                                                                                            
                                                                                                                                                     
      F --> G                                                                                                                                        
      F --> L
```

##  Problem with Current Design                                                                                                                        

1. Model contains incomplete grids (shock grids empty until params provided)                                                                       
2. update_sas_with_shocks mutates internal state at runtime                                                                                        
3. Unclear separation: what belongs to model vs what belongs to solution                                                                           

---

## Proposed Alternative Architecture

Basic idea: Everything except for `params` belongs to the `Model`. This includes transitions, shock descriptions, etc.

```mermaid
  flowchart TD                                                                                                                                       
      subgraph "Model = pure structure"                                                                                                              
          A1[Regime definitions]                                                                                                                     
          A2[Fixed grids<br/>'LinSpacedGrid' and friends]                                                                                          
          A3[Functions: utility,<br/>constraints, transitions]                                                                                       
          A4[Variable info]
          A5[Fixed params<br/>think partialling]
[...]

Issue #174, opened 2025-11-25
ENH: Make aggregation step in Q_and_F more flexible


timmens:
Currently, LCM only supports the aggregation function

$$
H(u, V) = u + \beta \cdot V.
$$

There should be the possibility to use a custom aggregation function, with some typical ones pre-implemented and selectable via an enum.

hmgaudecker:
Fixed in #275

PR #233, opened 2026-02-10
Replace `internal_regime_params` with flat kwargs via `dags.signature.rename_arguments`.

hmgaudecker:
Instead of wrapping each function to accept an `internal_regime_params` dict and extracting parameters at call time, use `dags.signature.rename_arguments()` to qualify parameter names with function prefixes (e.g., `risk_aversion` becomes `utility__risk_aversion`).

This makes the parameter flow explicit and removes the `internal_regime_params` indirection from all function signatures.

It is a crucial prerequisite for solving #174 and the partialling of fixed_params described in #219, which will be done in subsequent PRs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: check-hooks-apply
       - id: check-useless-excludes
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: v2.12.1
+    rev: v2.16.0
     hooks:
       - id: pyproject-fmt
   - repo: https://github.com/lyz-code/yamlfix
@@ -48,7 +48,7 @@ repos:
     hooks:
       - id: yamllint
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.14
+    rev: v0.15.1
     hooks:
       - id: ruff-check
         args:
--- a/pixi.lock
+++ b/pixi.lock
@@ -10808,8 +10808,8 @@ packages:
   timestamp: 1750615908735
 - pypi: ./
   name: pylcm
-  version: 0.0.2.dev72+g5f3761098.d20260211
-  sha256: 7dc8f80d2a78300a4dae2a4d2cdfdd7fd005008a7763c48ca09853f10828ae8f
+  version: 0.0.2.dev72+g24f07b5b7.d20260213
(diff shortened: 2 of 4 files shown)


tests/test_next_state.py
```diff
@@ -43,15 +42,11 @@ def test_get_next_state_function_with_solve_target():
...
-    internal_regime_params = MappingProxyType(
-        {
-            "discount_factor": 1.0,
-            "utility": {"disutility_of_work": 1.0},
-            "next_wealth": {
-                "interest_rate": 0.05,
-            },
-        }
-    )
+    flat_regime_params = {
```
timmens: Is there a reason why this dictionary does not need to be a `MappingProxyType` anymore?

src/lcm/input_processing/params_processing.py
```diff
@@ -183,3 +188,44 @@ def create_params_template(  # noqa: C901
...
+    {"discount_factor": float} are included as-is.
+
+    """
+    result = set()
+    for key, value in regime_params_template.items():
+        if isinstance(value, Mapping):
+            for param_name in value:
+                result.add(f"{key}{QNAME_DELIMITER}{param_name}")
+        else:
+            # Top-level param (e.g., "discount_factor": float)
```
timmens: This exact comment was renamed above to 

`Scalar param (currently unused - all params are under namespaces)`

src/lcm/input_processing/regime_components.py
```diff
@@ -171,20 +178,17 @@ def build_next_state_simulation_functions(
...
-    parameters = list(signature.parameters)
+    params = tuple(inspect.signature(next_state).parameters)
 
     next_state_vmapped = vmap_1d(
         func=next_state,
         variables=tuple(
-            parameter
-            for parameter in parameters
-            if parameter not in ("period", "age", "internal_regime_params")
+            p for p in params if p not in _get_non_vmap_params(regime_params_template)
```
timmens: I think `params` and/or the `p` in the loop should be replaced by something like `argnames` and `var`. I was slightly confused for which reason we would vmap over parameter, which we don't, we vmap over states and actions that are arguments of the next_state function.

timmens (approved):
While starting the review I was a bit skeptical about the change, but was quickly convinced that this is a big improvement. Very nice PR! Only minor comments, but already approved.

hmgaudecker:
These are great comments, thanks! 

Could you live with me applying them to #235 ? Would make life much easier...

PR #234, opened 2026-02-10
Factor out the time aggregation function, stop special-casing utility.

hmgaudecker:
- First part of tackling #174, though it will not be solved in all generality, so leaving that open. We'll take `H(utility, continuation_value; params)`, but no more complex forms.
- Include utility among the user functions and remove it from the attributes of `Regime`.


hmgaudecker:
@timmens:
- The crucial changes to time aggregation are in  if you want to have a look at them in isolation. Just be careful with `Q_and_F`, a chunk of code had accidentally slipped in here, which I removed in 2eb5b92dfb2a again.
- 0f235c5ad190 moves `utility` into the user functions. All those special cases had confused the hell out of me many times, this PR seemed like a good spot to do it since the structure is very similar to `H`. Apologies for the bloat.
- Rest are just simplifications / clarifications.

PR #235, opened 2026-02-10
Partial `fixed_params` into functions upon model creation. Stateless treatment of dynamic `IrregSpacedGrids`, in particular `ShockGrids`.

hmgaudecker:
Ended up a bit of a monster, sorry...

- This allows passing any parameter at model creation via `fixed_params` -> will be partialled into functions.
- Strongly-typed shock grids, inheritance diagram below.
- Update the `IrregSpacedGrids` / `ShockGrids` so they do the same thing: For both `IrregSpacedGrid`s and `ShockGrid`s, we can pass params directly to the instantiation, via `fixed_params`, or at runtime.
- Using terminology "fixed grid" / "points specified at runtime" instead of "static" / "dynamic" (too confusing in a library for dynamic economic models...). Also using mean / std (/ ar1_coeff for 
- Remove defaults from grid parameters (better be explicit)
- There should be a clear, stateless flow now for setting up a Model instance.

```mermaid
classDiagram
    ContinuousGrid <|-- ShockGrid
    ShockGrid <|-- ShockGridIID
    ShockGrid <|-- ShockGridAR1
    ShockGridIID <|-- ShockGridIIDUniform
    ShockGridIID <|-- ShockGridIIDNormal
    ShockGridAR1 <|-- ShockGridAR1Tauchen
    ShockGridAR1 <|-- ShockGridAR1Rouwenhorst

    class ShockGrid {
        +n_points: int
        +params
        +params_to_pass_at_runtime
        +is_fully_specified
        +compute_gridpoints()
        +compute_transition_probs()
        +get_gridpoints()
        +get_transition_probs()
        +to_jax()
        +get_coordinate()
    }

    class ShockGridIID {
        +draw_shock(params, key)
    }

    class ShockGridAR1 {
        +draw_shock(params, key, current_value)
    }

    class ShockGridIIDUniform {
        +start: float
        +stop: float
    }

    class ShockGridIIDNormal {
        +mean: float
        +std: float
        +n_std: float
    }

    class ShockGridAR1Tauchen {
        +ar1_coeff: float
        +std: float
        +mean: float
        +n_std: float
    }

    class ShockGridAR1Rouwenhorst {
        +ar1_coeff: float
        +std: float
        +mean: float
    }
```

### DIFF
--- a/src/lcm/input_processing/regime_processing.py
+++ b/src/lcm/input_processing/regime_processing.py
@@ -202,6 +202,7 @@ def process_regimes(
             next_state_simulation_function=next_state_simulation_function,
             # currently no additive utility shocks are supported
             random_utility_shocks=ShockType.NONE,
+            _base_state_action_space=state_action_spaces[name],
         )
 
     return ensure_containers_are_immutable(internal_regimes)
--- a/src/lcm/solution/solve_brute.py
+++ b/src/lcm/solution/solve_brute.py
@@ -5,13 +5,8 @@ import jax.numpy as jnp
 
 from lcm.ages import AgeGrid
 from lcm.error_handling import validate_value_function_array
-from lcm.grids import Grid, IrregSpacedGrid
-from lcm.interfaces import (
-    InternalRegime,
-    StateActionSpace,
-)
-from lcm.shocks import _ShockGrid
-from lcm.typing import FlatRegimeParams, FloatND, InternalParams, RegimeName
+from lcm.interfaces import InternalRegime
+from lcm.typing import FloatND, InternalParams, RegimeName
 
 
 def solve(
@@ -51,11 +46,8 @@ def solve(
         }
 
         for name, internal_regime in active_regimes.items():
-            state_action_space = _replace_runtime_states(
-                internal_regime.state_action_space,
-                internal_params[name],
-                internal_regime.gridspecs,
-                internal_regime.resolved_fixed_params,
+            state_action_space = internal_regime.state_action_space(
+                flat_regime_params=internal_params[name],
             )
             max_Q_over_a = internal_regime.max_Q_over_a_functions[period]
 
@@ -75,48 +67,3 @@ def solve(
         logger.info("Age: %s", ages.values[period])
 
     return MappingProxyType(solution)
-
-
-def _replace_runtime_states(
-    state_action_space: StateActionSpace,
-    params: FlatRegimeParams,
-    gridspecs: MappingProxyType[str, Grid],
-    fixed_params: FlatRegimeParams = MappingProxyType({}),
-) -> StateActionSpace:
-    """Complete state grids whose values are supplied at runtime via params.
-
-    For IrregSpacedGrid with runtime-supplied points, the grid points come from
-    params as ``{state_name}__points``. For _ShockGrid with runtime-supplied params,
-    the grid points are computed from shock params in the params dict or
-    fixed_params.
-
-    """
-    all_params = {**fixed_params, **params}
-    replacements: dict[str, object] = {}
-    for state_name, spec in gridspecs.items():
-        if state_name not in state_action_space.states:
-            continue
-        if isinstance(spec, IrregSpacedGrid) and spec.pass_points_at_runtime:
-            points_key = f"{state_name}__points"
-            if points_key not in all_params:
-                continue
-            replacements[state_name] = all_params[points_key]
-        elif isinstance(spec, _ShockGrid) and spec.params_to_pass_at_runtime:
-            all_present = all(
-                f"{state_name}__{p}" in all_params
-                for p in spec.params_to_pass_at_runtime
-            )
-            if not all_present:
-                continue
-            shock_kw = dict(spec.params)
-            for p in spec.params_to_pass_at_runtime:
-                shock_kw[p] = all_params[f"{state_name}__{p}"]
-            replacements[state_name] = spec.compute_gridpoints(
-                spec.n_points, **shock_kw
-            )
-
-    if not replacements:
-        return state_action_space
-
-    new_states = dict(state_action_space.states) | replacements
-    return state_action_space.replace(states=MappingProxyType(new_states))
--- a/src/lcm/interfaces.py
+++ b/src/lcm/interfaces.py
@@ -6,7 +6,7 @@ from types import MappingProxyType
 import pandas as pd
 from jax import Array
 
-from lcm.grids import ContinuousGrid, DiscreteGrid, Grid
+from lcm.grids import ContinuousGrid, DiscreteGrid, Grid, IrregSpacedGrid
 from lcm.shocks import _ShockGrid
 from lcm.typing import (
     ArgmaxQOverAFunction,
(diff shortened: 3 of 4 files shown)


examples/mahler_yum_2024/model.py
```diff
@@ -21,7 +20,7 @@
 import lcm
 from lcm import AgeGrid, DiscreteGrid, LinSpacedGrid, Model, Regime, categorical
 from lcm.dispatchers import _base_productmap
-from lcm.grids import ShockGrid
+from lcm.shock_grids import ShockGridAR1Rouwenhorst, ShockGridIIDUniform
```
timmens: For a more pleasant user-interface, I would leave out the "ShockGrid" prefix. Potentially, I would even rename the shock_grids module to shocks.

I can also see the argument that you want to make clear to the user that they are describing grids. Personally, I still prefer the minimalistic UI (without prefix), which focuses on the actual object they are trying to model (the shock) and not the implementation (the grid).

src/lcm/shock_grids.py
```diff
@@ -0,0 +1,337 @@
...
+    ) -> Float1D:
+        return params["mean"] + params["std"] * jax.random.normal(key=key)
+
+
+@dataclass(frozen=True, kw_only=True)
+class ShockGridAR1Tauchen(ShockGridAR1):
+    r"""AR(1) shock discretized via Tauchen (1986).
+
+    The process is
+    :math:`y_t = \mu_\varepsilon + \rho \, (y_{t-1} - \mu_\varepsilon) + \varepsilon_t`,
```
mj023: This AR1 Process is a bit odd. The function that draws the shock is $\rho Y_{t-1} - \rho \mu_{\epsilon} + \epsilon_t$. Which is not what is stated here, but also generally not what I would expect. The first  $\mu_{\epsilon}$ probably needs to be new parameter if we want something that just shifts the mean.

src/lcm/shock_grids.py
```diff
@@ -0,0 +1,337 @@
...
+        params: MappingProxyType[str, float],
+        key: FloatND,
+        current_value: Float1D,
+    ) -> Float1D:
+        mu = params["mean"]
+        return (
+            mu
+            + params["ar1_coeff"] * (current_value - mu)
+            + params["std"] * jax.random.normal(key=key)
+        )
```
mj023: Same as in the other function.

src/lcm/shock_grids.py
```diff
@@ -0,0 +1,337 @@
...
+
+        """
+        if not self.is_fully_specified:
+            return jnp.full(self.n_points, jnp.nan)
+        return self.compute_gridpoints(self.n_points, **self.params)
+
+    def get_transition_probs(self) -> FloatND:
+        """Get the transition probabilities at the gridpoints.
+
+        Returns uniform probabilities when required params are missing.
```
timmens: Do we actually want that, or do we want to throw an error if its not fully specified and we request the transition probs?

timmens (commented):
Very nice PR! 🚀 

I have a few comments before full approval, but nothing serious. In addition, I have two comments here:

1. I think this PR is missing an economically sensible test for the shocks. A test where we predict (some part of) the result of the solution or simulation given the chosen shock. Currently there is no end-to-end test like this. Only the regression test, but that does not test a behavior.
2. At some point I believe we must think about a clearer separation of code/objects that are used for the solution and code/objects used for the simulation. Definitely not in this PR -- was just reminded of that while reading the changes.

src/lcm/model.py
```diff
@@ -408,36 +445,191 @@ def _validate_all_variables_used(regimes: Mapping[str, Regime]) -> list[str]:
...
+
+
+def _filter_kwargs_for_func(
+    func: Callable, kwargs: Mapping[str, object]
+) -> Mapping[str, object]:
+    """Filter kwargs to only those accepted by func's signature."""
+
+    try:
+        sig = inspect.signature(func)
+    except ValueError, TypeError:
```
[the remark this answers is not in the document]
hmgaudecker: It's actually fine, I believe (the comma creates a tuple), but agree that making it explicit is nicer.

CLAUDE.md
```diff
@@ -277,6 +277,11 @@ initial_regimes = ["working", "working", "retired"]
 - All functions require type annotations
 - Pre-commit hooks ensure code quality
 
+### Testing Style
+
+- Use plain pytest functions, never test classes (`class TestFoo`)
```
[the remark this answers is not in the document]
hmgaudecker: Yip, I noted it at some point but apparently only fixed later occurrences... Sorry for the noise.

src/lcm/shock_grids.py
hmgaudecker: Good one. Almost -- returning NaN now because we need the shapes and dtypes when building the model (tl;dr).


### Why NaN instead of raising?

These methods are called during `Model.__init__` → `process_regimes`(). At that stage JAX is tracing the computation graph — it needs concrete shapes and dtypes, but the actual values don't matter yet:

- Not-fully-specified grids have `params_to_pass_at_runtime`. The processing code checks this and takes the runtime branch, which closes over `compute_gridpoints` / `compute_transition_probs` directly and calls them inside JIT with the user-supplied params.
- The static branch only runs when the grid is fully specified, so the `get_transition_probs()` call there always gets real values.
- `to_jax()` at line 115 builds the grids dict used for StateActionSpace construction. For an unspecified shock grid, this produces a NaN array, but that array is only used as a shape/placeholder — the actual gridpoints are recomputed at runtime from params.

So in normal usage through `Model`, the NaN arrays are never consumed for arithmetic — they're structural placeholders. An exception would break model initialization for any shock whose params are deferred to runtime, which is the core use case.

When could a user actually see NaN?

Only by calling `grid.get_gridpoints()` or `grid.get_transition_probs()` directly on a partially-specified grid object outside of the model pipeline — e.g. `Tauchen(n_points=5).get_transition_probs()`. That's a reasonable thing to do for inspection, and NaN makes the "these aren't real values" situation obvious (and will propagate loudly through any downstream computation), whereas the old uniform matrix silently looked correct.

src/lcm/shock_grids.py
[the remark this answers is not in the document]
hmgaudecker: Referring to QuantEcon for now -- @mj023 pleas chime in in case that is wrong!

tests/test_shock_grids.py
```diff
@@ -0,0 +1,238 @@
...
+from lcm._config import TEST_DATA
+from tests.conftest import DECIMAL_PRECISION, X64_ENABLED
+from tests.test_models.shocks import get_model, get_params
+
+
+@pytest.mark.skipif(not X64_ENABLED, reason="Not working with 32-Bit because of RNG")
+@pytest.mark.parametrize(
+    "distribution_type", ["uniform", "normal", "tauchen", "rouwenhorst"]
+)
+def test_model_with_shock(distribution_type):
```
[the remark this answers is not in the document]
hmgaudecker: I opened #240 for that -- I do agree, ofc, but while this PR refactors the shocks, it does not really add anything new in that respect (maybe the possibility to pass parameters other that mu=0, sigma=1, but I would argue that that is not enough to add a test right here).

hmgaudecker:
> 1. I think this PR is missing an economically sensible test for the shocks. A test where we predict (some part of) the result of the solution or simulation given the chosen shock. Currently there is no end-to-end test like this. Only the regression test, but that does not test a behavior.

See #240 

> 2. At some point I believe we must think about a clearer separation of code/objects that are used for the solution and code/objects used for the simulation. Definitely not in this PR -- was just reminded of that while reading the changes.

See [note in #219](https://github.com/OpenSourceEconomics/pylcm/issues/219#issuecomment-3906474580).

src/lcm/input_processing/regime_processing.py
```diff
@@ -441,8 +423,42 @@ def next_func(**kwargs: Any) -> Int1D:  # noqa: ARG001, ANN401
...
+                input=transition_probs,
+                coordinates=[
+                    jnp.full(n_points, fill_value=coord),
+                    jnp.arange(n_points),
+                ],
+            )
+
+        return weights_func_runtime
+
+    transition_probs = gridspec.get_transition_probs()  # ty: ignore[unresolved-attribute]
```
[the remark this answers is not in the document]
hmgaudecker: Better now?

src/lcm/solution/solve_brute.py
```diff
@@ -48,7 +51,12 @@ def solve(
         }
 
         for name, internal_regime in active_regimes.items():
-            state_action_space = internal_regime.state_action_space
+            state_action_space = _replace_runtime_states(
+                internal_regime.state_action_space,
+                internal_params[name],
+                internal_regime.gridspecs,
+                internal_regime.resolved_fixed_params,
+            )
```
timmens: What I actually had in mind was:

```python
for name, internal_regime in active_regimes.items():
    state_action_space = internal_regime.build_state_action_space(
        params=internal_params[name],
    )
```

The "uncompleted" state-action space (currently `internal_regime.state_action_space`) could then even become private, because AFAIK, we don't need it for anything else except to build the full state-action space above.

This would also move the code (`_replace_runtime_states`) that, IMO, belongs to the regime out of the solution module. If we add a new solution module we would have to move this into another module anyways, then why not add it to internal regime directly.

hmgaudecker: Hope I got it now. Apologies, should have seen that in the morning already, but it is not my sharpest day.

Issue #236, opened 2026-02-10
BUG: State variable only used in transitions causes ValueError in vmap_1d at last regime period


hmgaudecker:
## Description

When a state variable is only used in transition functions (not in utility or constraints), and the regime transitions to a period for which that particular transition is irrelevant, `Model` construction fails with `ValueError: list.index(x): x not in list` in `vmap_1d`.

## Root cause

At the last period of a non-terminal regime (where the only active target for the next period is a terminal regime like "dead"), the `Q_and_F` function's parameter list is computed via `_get_arg_names_of_Q_and_F`, which takes the union of parameters across all dependency functions. Since state transitions are not needed when transitioning to a terminal regime (which has no states), any state variable that **only** appears in transition functions is pruned from the `Q_and_F` signature.

However, `build_argmax_and_max_Q_over_a_functions` calls `simulation_spacemap` with `states_names=tuple(state_action_space.states)`, which still includes **all** states defined in the current regime. When `vmap_1d` tries to find the pruned state variable's position in the `Q_and_F` function's parameters, it fails because the variable is not in the signature.

The mismatch is between:
- `state_action_space.states` (all states declared for the regime — includes `type_var`)
- Q_and_F function signature at last period (only states actually needed — excludes `type_var`)

## Minimal reproducer

```python
import jax.numpy as jnp
from lcm import AgeGrid, DiscreteGrid, LinSpacedGrid, Model, Regime, categorical


@categorical
class RegimeId:
    alive: int
    dead: int


@categorical
class TypeVar:
    low: int
    high: int


def utility(consumption, wealth):
    return jnp.log(consumption) + 0.01 * wealth


def dead_utility():
    return 0.0


def next_wealth(wealth, consumption, type_var):
    """type_var affects wealth transition but does NOT appear in utility."""
    return (1 + 0.05 * type_var) * (wealth - consumption)


def next_type_var(type_var):
    return type_var


def next_regime(age):
    return jnp.where(age >= 2, RegimeId.dead, RegimeId.alive)


ages = AgeGrid(start=0, stop=3, step="Y")

alive = Regime(
    utility=utility,
    states={
        "wealth": LinSpacedGrid(start=1, stop=100, n_points=10),
        "type_var": DiscreteGrid(TypeVar),
    },
    actions={
        "consumption": LinSpacedGrid(start=1, stop=50, n_points=10),
    },
    transitions={
        "next_wealth": next_wealth,
        "next_type_var": next_type_var,
[...]

PR #237, opened 2026-02-11
Allow more deeply nested mappings in `params` while retaining tight typing.

hmgaudecker:
## Summary

Small update in order to work sensibly with GETTSIM functions.

- Add `MappingLeaf`, a JAX-pytree-compatible wrapper that prevents `flatten_regime_namespace` from recursing into its contents, allowing nested parameter dicts to be passed through the params pipeline as opaque leaves. Same for `SequenceLeaf`.
- Add an `as_leaf` function, wich is how these will typically be generated, make it available via `lcm.params.as_leaf`
- Teach `_make_immutable` / `_make_mutable` to handle `MappingLeaf` (converting inner data while preserving the wrapper)
- *[somewhat unrelated, just because this bug crept up in the model ]* Add reproducer test for #236 (state used only in transitions with terminal regime)

## Test plan
- [x] `test_mapping_leaf.py` — construction, repr, equality, JAX pytree round-trips, immutable/mutable conversions, `flatten_regime_namespace` leaf behavior
- [x] `test_model.py::test_state_only_in_transitions_with_terminal_regime` — reproducer for #236. 

  *Somewhat unrelated, just because the bug crept up in the model where we are already using this branch.*

  This passes out of the box; must have been fixed along the way somewhere.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -283,6 +283,11 @@ initial_regimes = ["working", "working", "retired"]
 - Use plain pytest functions, never test classes (`class TestFoo`)
 - Use `@pytest.mark.parametrize` for test variations
 
+### Testing Style
+
+- Use plain pytest functions, never test classes (`class TestFoo`)
+- Use `@pytest.mark.parametrize` for test variations
+
 ### Key Dependencies
 
 - **jax**: Numerical computation


hmgaudecker:
@timmens @MImmesberger -- do you like `lcm.params.MappingLeaf` ?

hmgaudecker:
@MImmesberger, can you switch to `.as_leaf`? That should be the idiomatic way of using it, I am thinking about the classes more in terms of typing.

hmgaudecker:
@timmens -- I thought you had approved already, but either I misremember or it got lost because of subsequent changes. Anything left?

Issue #185, opened 2025-12-02
ENH: Retrieve stochastic transition array shapes


timmens:
In #182 we disabled the calculation of the stochastic transition array shapes for the parameter template, since it has gotten more complicated because of the fact that in #182 we now allow for more flexible stochastic transition functions.

This feature is, however, really nice and should be re-added in the future because of two main factors:

1. The user can quickly check the params template for the shape of the transition array that she needs to provide
2. We can use this information to check the validity of the user provided transition array shape at runtime and raise a more readable error

hmgaudecker:
Closing [as discussed](https://github.com/OpenSourceEconomics/pylcm/pull/250#discussion_r2848210733) in #250 -- this cannot possibly work in the new stochastic transitions where users can write any transition code they want.

Issue #152, opened 2025-09-13
ENH: Do not require next state functions for single period models


felixschmitz:
> - **Feature**: simplify user model initialisation.

#### Is your feature request related to a problem?

Currently, PyLCM requires for the initialisation of a user model for each state a corresponding next state function. If the next state function is missing for a state, `_validate_logical_consistency` inside the `__post_init__` raises a `ModelInitilizationError`. 
If the user model is specified with `n_periods=1`, the next state function is never executed.

#### Describe the solution you'd like

I wish to use a PyLCM user model for a single period without having to specify next state functions. 
I suspect that a first step is to adapt the `__post_init__` of the user model.

Issue #63, opened 2024-03-22
BUG: Transition matrices with incorrect shapes do not throw error


buddejul:
I came across unexpected behavior for stochastic next functions with the `_period` argument.

For `N_PERIODS = 3` and a binary state I thought we needed a 2x2 matrix as discussed somewhere above (one for each period transition). But the following does not throw an error:

```python
@lcm.mark.stochastic
def next_dummy_state(_period):
    pass

DUMMY_STATE_TRANSITION = jnp.array(
    [[0.5, 0.5]]
)
```

I checked it's actually included in the model, providing an empty transition matrices raises an indexing error as expected.

Providing a matrix that's "too large" (e.g. `(4, 2)`) here also doesn't result in an error.

timmens:
Thank you for opening the issue!

This seems to be related to JAX's out-of-bounds indexing. The following code runs without error:

```python
import jax.numpy as jnp

jnp.arange(10)[11]

> Array(9, dtype=int32)
```

For reference,see: https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#out-of-bounds-indexing

I will have to check if it is not a more serious problem however. In any case, we have to throw an informative error on the LCM side.

hmgaudecker:
Great detective work. Sounds like we'll need to add bound checks up front.

Also, a good example why you want to test with asymmetric inputs. I wonder whether things still work with `[0.4, 0.6]` or you'd see a violation of add-to-one constraints?

buddejul:
Re asymmetry: I originally noticed this for a state in our replication and the row had asymmetric entries (but would need to confirm). So I guess the indexing is not at the individual entry level but the rows (at least in this example with just period as argument).

FWIW I also tried different combinations of required rows (i.e..periods) and provided rows but this didn't seem to have an effect, consistent with the above.

PR #249, opened 2026-02-22
Reproducers for old issues, validate initial states upfront

### LINKED ISSUE #64: ENH: Check that initial conditions for simulations are feasible
#### Is your feature request related to a problem?

Inspired by a test case created by @buddejul:

```python
from copy import deepcopy

import numpy as np
import jax.numpy as jnp
from lcm.entry_point import get_lcm_function
from tests.test_models.deterministic import ISKHAKOV_ET_AL_2017

PARAMS = {
    "beta": 0.95,
    "utility": {"disutility_of_work": 1},
    "next_wealth": {"interest_rate": 0.02},
    "next_lagged_retirement": {},
    "consumption_constraint": {},
    "absorbing_retirement_filter": {},
    "labor_income": {"wage": 1},
    "working": {},
}

INITIAL_STATES = {
    "wealth": jnp.array([0.25]),
    "lagged_retirement": jnp.array([0]),
}


def solve_and_simulate(model, params=PARAMS, initial_states=INITIAL_STATES):
    solve, params_dict = get_lcm_function(model, targets="solve")
    simulate, _ = get_lcm_function(model, targets="simulate")

    vf_arr_list = solve(params)

    return simulate(params, vf_arr_list=vf_arr_list, initial_states=initial_states)

def test_choice_constraint():
    result = solve_and_simulate(ISKHAKOV_ET_AL_2017)
    assert np.isfinite(result["value"]).all()
```

The test will fail because:

```pycon
>>> result["value"]
period  initial_state_id
0       0                  -inf
1       0                  -inf
2       0                  -inf
Name: value, dtype: float32
```

Note that in the initial period, the consumption constraint 

```py
    return consumption <= wealth
```

cannot ever be fulfilled, because the minimum gridpoint of consumption is 1 and wealth is 0.25.

#### Describe the solution you'd like

Check that the initial states are valid in the sense that all of them admit at least one valid choice.

Maybe we can just apply filters and constraints to the initial states. Not fully obvious to me yet how to do that as part of an estimation when the outcome of this check may depend on parameters picked by the optimizer (i.e., what precisely we should do upon failure), but just sticking our head in the sand seems worse (will need to figure out much later what causes all those `-inf`s)

Another thing is that we might have to allow extrapolating beyond the grid boundaries.

hmgaudecker:
Add reproducers for 
- #64 
- #63 (xfail, tackle in new PR)
- #39 (just regression guard -- already fixed)
- #35 (just regression guard -- already fixed)
- #152 (just regression guard -- already fixed)
- #185 (xfail, tackle in new PR)

Fix #64.


tests/test_stochastic.py
```diff
@@ -216,3 +225,169 @@ def test_compare_deterministic_and_stochastic_results_value_function(
...
+    dead: int
+
+
+_STOCH_FINAL_AGE = 1  # 3 periods total
+
+
+@lcm.mark.stochastic
+def _next_shock(shock: DiscreteState, shock_transition: FloatND) -> FloatND:
+    """Default stochastic transition using a pre-computed transition matrix."""
+    return shock_transition[shock]
```
timmens: This is also a bit weird with the "_stoch" prefix. Maybe move these functions into _make_minimal_stochastic_model?

PR #250, opened 2026-02-22
Unified interface for stochastic transitions

hmgaudecker:
~~Fix attempted for #63.~~

~~Fix attempted for #185.~~ 

### Overview

Replace the dual `@stochastic` decorator + attribute-checking mechanism with a unified  
grid-type-based approach. Stochastic transitions are now determined entirely from grid
types (`DiscreteMarkovGrid`, `_ShockGrid`) and the `Regime.stochastic_transition` flag,
eliminating the `mark.py` module and `is_stochastic_transition()` helper.

### New public API: `DiscreteMarkovGrid`                                                 
  
- Extract `_DiscreteGridBase(Grid)` in `grids.py` — private base class holding shared    
  code (categories, codes, `to_jax`). Both `DiscreteGrid` and `DiscreteMarkovGrid` now
  inherit from `_DiscreteGridBase` instead of `DiscreteMarkovGrid` inheriting from       
  `DiscreteGrid`.                                                                        
- `DiscreteGrid(_DiscreteGridBase)` — optional deterministic transition (or `None`
  for fixed states).
- `DiscreteMarkovGrid(_DiscreteGridBase)` — requires a callable stochastic transition
  (raises on `None`). Exposes `n_states` property. Has its own `transition` property
  typed as `Callable[..., FloatND]` (probability distribution), distinct from
  `DiscreteGrid.transition` which returns a `DiscreteState`.
- Export from `lcm.__init__` and add to `__all__`.
- Users now write `DiscreteMarkovGrid(HealthStatus, transition=next_health)` instead of
  `DiscreteGrid(HealthStatus, transition=stochastic(next_health))`.
- Update all `isinstance(grid, DiscreteGrid)` checks downstream to
  `isinstance(grid, DiscreteGrid | DiscreteMarkovGrid)` — in `interfaces.py`,
  `state_action_space.py`, `simulation/result.py`, and `simulation/util.py`.

### New regime-level flag: `Regime.stochastic_transition`

- Add `stochastic_transition: bool = False` field to `Regime`.
- Validates that terminal regimes cannot have `stochastic_transition=True`.
- Replaces `stochastic(self.transition)` wrapping in `get_all_functions()`.

### Delete `mark.py` and `is_stochastic_transition()`

- Delete `src/lcm/mark.py` (`StochasticInfo`, `stochastic()` decorator).
- Remove `is_stochastic_transition()` from `input_processing/util.py`.
- Remove `lcm.mark` from `lcm.__init__.__all__`.

### Name-set-based stochastic detection

- Add `stochastic_transition_names: frozenset[str]` field to `InternalFunctions`.
- In `regime_processing._get_internal_functions`, compute the set from grid types:
  - `DiscreteMarkovGrid` states → `next_{name}`
  - Shock states (from `variable_info.is_shock`) → `next_{name}`
- Fix latent bug: `is_stochastic_regime_transition` was always `False` because
  `_extract_transitions_from_regime` passed the raw `regime.transition` (never wrapped
  with `stochastic()`). Now reads `regime.stochastic_transition` directly.

### Thread `stochastic_transition_names` through downstream code

- `Q_and_F.py`: `get_Q_and_F`, `_get_joint_weights_function` — use name-in-set.
- `next_state.py`: `get_next_stochastic_weights_function`,
  `get_next_state_function_for_simulation`, `_extend_transitions_for_simulation` —
  accept and use set parameter. Clean up `continuous_stochastic_targets` to drop
  unused `func` variable.
- `simulation/util.py`: `calculate_next_states` — use set from `internal_regime`.
- `simulation/result.py`: `_get_stochastic_weight_function_names` — use set from
  `internal_regime`.
- `regime_components.py`: `build_next_state_simulation_functions` — pass set through.
- `regime.py`: `_collect_state_transitions` — drop `stochastic()` wrappers, remove
  dead `DiscreteGrid + _stochastic_info` error check, add `DiscreteMarkovGrid` branch.

### Improved validation

- ~~`create_regime_params_template`: for `DiscreteMarkovGrid` transitions, replace~~
  ~~`FloatND` type annotations with `(n_states, n_states)` shape tuples in the params~~
  ~~template so users see the expected matrix dimensions.~~
  
  This does not work because the user can write any code for stochastic transitions now.  
  
- ~~`params_processing._validate_param_shapes`: new validation that checks array~~
  ~~parameters match expected shape suffixes from the template.~~

- `typing.py`: widen `RegimeParamsTemplate` and `MutableRegimeParamsTemplate` to
  accept `type | tuple[int, ...]`.
- `model.py`: widen `_remove_fixed_from_template` type annotations to match.
- ~~Un-xfail two tests: `test_wrong_transition_matrix_shape_rejected` (issue #63) and~~
  ~~`test_params_template_includes_stochastic_transition_shape` (issue #185).~~

### Fix initial-state feasibility check with age/period-dependent constraints

- `validate_initial_state_feasibility` and `_check_regime_feasibility` now accept
  `initial_age: float` and supply `period=0` / `age=initial_age` to the feasibility
  function when its signature requires them.
- `Model.simulate` and `Model.solve_and_simulate` pass
  `initial_age=float(self.ages.values[0])`.

### Example model updates

- `examples/mahler_yum_2024/model.py`: replace `@lcm.mark.stochastic` with
  `DiscreteMarkovGrid` for health, add `stochastic_transition=True` for regime
  transition.
- `examples/consumption_saving/model.py`: benefits from the feasibility fix.

### DIFF
--- a/src/lcm/input_processing/params_processing.py
+++ b/src/lcm/input_processing/params_processing.py
@@ -106,8 +106,6 @@ def process_params(
         regime_name, remainder = key.split(QNAME_DELIMITER, 1)
         result[regime_name][remainder] = value
 
-    _validate_param_shapes(result, params_template)
-
     return cast("InternalParams", ensure_containers_are_immutable(result))
 
 
@@ -191,38 +189,6 @@ def create_params_template(  # noqa: C901
     # E.g., labor_income is a function in 'working' but a param in 'retired'.
 
     return ensure_containers_are_immutable(template)
-
-
-def _validate_param_shapes(
-    resolved: dict[str, dict[str, Any]],
-    template: ParamsTemplate,
-) -> None:
-    """Validate that array parameters match expected shape suffixes.
-
-    When a template value is a tuple of ints (e.g. (2, 2) for a transition matrix),
-    check that the corresponding resolved value has a matching shape suffix.
-
-    """
-    for regime_name, regime_template in template.items():
-        for func_name, func_params in regime_template.items():
-            if not isinstance(func_params, Mapping):
-                continue
-            for param_name, expected in func_params.items():
-                if not (
-                    isinstance(expected, tuple)
-                    and all(isinstance(d, int) for d in expected)
-                ):
-                    continue
-                flat_key = f"{func_name}{QNAME_DELIMITER}{param_name}"
-                value = resolved.get(regime_name, {}).get(flat_key)
-                if value is None or not hasattr(value, "shape"):
-                    continue
-                n = len(expected)
-                if len(value.shape) < n or value.shape[-n:] != expected:
-                    raise InvalidParamsError(
-                        f"Parameter '{regime_name}/{func_name}/{param_name}' "
-                        f"has shape {value.shape}, expected suffix {expected}."
-                    )
 
 
 def get_flat_param_names(regime_params_template: RegimeParamsTemplate) -> set[str]:


src/lcm/input_processing/create_regime_params_template.py
```diff
@@ -63,6 +63,18 @@ def create_regime_params_template(
...
 
+    # Replace FloatND type annotations with (n_states, n_states) shape tuples
+    # for DiscreteMarkovGrid transitions, so users know the expected matrix shape.
+    for state_name, grid in regime.states.items():
+        if isinstance(grid, DiscreteMarkovGrid):
+            func_key = f"next_{state_name}"
+            if func_key in function_params:
+                n = grid.n_states
+                function_params[func_key] = {
+                    pname: (n, n) if ptype is FloatND else ptype
```
timmens: I have a concern about the (n, n) shape hints here:

In the previous approach (where the transition function had no body and was just a signature), the argument ordering of the function directly determined the ordering of the transition
array dimensions. That meant the framework had full control over the mapping between dimensions and variables, and there was no ambiguity.

Now the transition function has a body that indexes into a user-provided array (partner_transition[period, labor_supply, partner]), so the dimension ordering is entirely determined by
how the user writes that indexing expression. The framework has no way to verify that the array dimensions actually correspond to the right variables. If someone provides an array with
the axes in the wrong order but all dimensions happen to have the same size (which is common: e.g., n_labor == n_partner == 2), validation passes, the model solves without error, and
produces silently incorrect results.

Given that, I think showing (2, 2) in the params template is actively misleading as it suggests the framework understands and validates the expected shape, when in reality it can only
check a suffix that may match by coincidence. I'd argue we should go back to `FloatND`, because I don't think we can get the shape information in general.

timmens: Yes, let's close it!

PR #258, opened 2026-02-26
Allow for Gauss-Hermite points; add log-normal shocks.

### LINKED ISSUE #248: ENH: Add log-normal shocks and improve quadrature points
- [ ] Add log-normal shocks to `lcm.shocks.iid`
- [ ] Discuss whether we want to support optimal nodes for the IID shocks
   
   - For instance, for the Normal shock we could use the Gauss-Hermite quadrature points, which would also remove the `n_std` parameter.

   - For the LogNormal shock there might be similar nodes.

*(edit: idd -> iid; in case that was misguided, lmk!)*

hmgaudecker:
Fixes #248.

API change: When requesting Normal / Tauchen / log-Normal, users now need to explicitly set a Boolean flag `gauss_hermite`. 

Cleanup of prior PRs: Make example models run again; these had not included age among the initial states yet.

### DIFF
--- a/src/lcm/shocks/ar1.py
+++ b/src/lcm/shocks/ar1.py
@@ -142,15 +142,6 @@ class Rouwenhorst(_ShockGridAR1):
     mu: float | None = None
     """Intercept (drift) of the AR(1) process."""
 
-    def __post_init__(self) -> None:
-        if self.n_points % 2 == 0:
-            msg = (
-                f"n_points must be odd (got {self.n_points}). Odd n guarantees a"
-                " quadrature node at the mean (Abramowitz & Stegun, 1972,"
-                " Table 25.10)."
-            )
-            raise GridInitializationError(msg)
-
     def compute_gridpoints(self, n_points: int, **kwargs: float) -> Float1D:
         rho, sigma, mu = kwargs["rho"], kwargs["sigma"], kwargs["mu"]
         nu = jnp.sqrt((n_points - 1) / (1 - rho**2)) * sigma
--- a/tests/test_shock_grids.py
+++ b/tests/test_shock_grids.py
@@ -293,11 +293,10 @@ def test_ar1_draw_shock_unconditional_moments(grid_cls):
         (lcm.shocks.iid.Normal, {"gauss_hermite": True}),
         (lcm.shocks.iid.LogNormal, {"gauss_hermite": True}),
         (lcm.shocks.ar1.Tauchen, {"gauss_hermite": True}),
-        (lcm.shocks.ar1.Rouwenhorst, {}),
     ],
 )
 def test_even_n_points_rejected(grid_cls_and_kwargs):
-    """Normal, LogNormal, Tauchen, and Rouwenhorst reject even n_points."""
+    """Gauss-Hermite grids reject even n_points."""
     grid_cls, extra_kw = grid_cls_and_kwargs
     with pytest.raises(GridInitializationError, match="n_points must be odd"):
         grid_cls(n_points=4, **extra_kw)


src/lcm/shocks/ar1.py
```diff
@@ -32,10 +33,18 @@ class Tauchen(_ShockGridAR1):
     $y_t = \mu + \rho \, y_{t-1} + \varepsilon_t$,
     where $\varepsilon_t \sim N(0, \sigma_\varepsilon^2)$.
 
-    Original implementation follows [QuantEcon](https://quanteconpy.readthedocs.io/en/latest/markov/approximation.html#quantecon.markov.approximation.tauchen).
+    When `gauss_hermite=True`, the grid uses Gauss-Hermite quadrature nodes
+    with importance-sampling weights following
+    [Tauchen & Hussey (1991)](https://doi.org/10.2307/2938229).
```
timmens:
```suggestion
    [Tauchen & Hussey (1991)](https://doi.org/10.2307/2938261).
```

src/lcm/shocks/ar1.py
```diff
@@ -48,34 +57,56 @@ class Tauchen(_ShockGridAR1):
...
         x = jnp.linspace(-x_max, x_max, n_points)
         return x + mu / (1 - rho)
 
     def compute_transition_probs(self, n_points: int, **kwargs: float) -> FloatND:
         rho, sigma = kwargs["rho"], kwargs["sigma"]
+        if self.gauss_hermite:
+            nodes, weights = _gauss_hermite_normal(n_points, 0.0, sigma)
+            f_cond = pdf(nodes[None, :], loc=rho * nodes[:, None], scale=sigma)
+            g_prop = pdf(nodes, loc=0.0, scale=sigma)
+            raw = weights * f_cond / g_prop
```
timmens: We do a division by `g_prop` here which might be very small. Maybe we should make this numerically more robust?

timmens (commented):
Ah, and one more comment: I find it a bit hard to completely judge the mathematical correctness here, as I haven't worked a lot with quadrature and Markov-chains. To make the review a bit easier (also for the future) we could add a page to the docs where the shock math is explained. On this level we can also then ask our friends with a background in numerical mathematics to check it.

src/lcm/shocks/ar1.py
```diff
@@ -48,34 +57,62 @@ class Tauchen(_ShockGridAR1):
...
+            exclude.add("n_std")
+        return tuple(f.name for f in fields(self) if f.name not in exclude)
+
     def compute_gridpoints(self, n_points: int, **kwargs: float) -> Float1D:
-        rho, sigma = kwargs["rho"], kwargs["sigma"]
-        mu, n_std = kwargs["mu"], kwargs["n_std"]
+        rho, sigma, mu = kwargs["rho"], kwargs["sigma"], kwargs["mu"]
+        if self.gauss_hermite:
+            std_y = jnp.sqrt(sigma**2 / (1 - rho**2))
+            nodes, _weights = _gauss_hermite_normal(n_points, 0.0, std_y)
```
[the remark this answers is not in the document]
mj023: The AI definitely seems to have read more papers than me. I checked the references and the $\sigma_y$ is from Floden 2008, and it sounds convincing, even though Floden 2008 finds an even better value for $\sigma$. For the Tauchen-Binning stuff I'm not so sure, I think it's not in the papers being cited and also it just seems suspect to me to go the distance to get the hermitian points and then just not use the associated weights at all. But in the end the Rouwenhorst method seems to be the most accurate anyways.

src/lcm/shocks/ar1.py
hmgaudecker: This would happen only if you went very far into the tails -- not sure how and why that would happen in practice, have not seen more than 3-7 quadrature points. I'd wait until it actually becomes a problem.

timmens: I am convinced. I let Claude run a few experiments and it should not be an issue under normal usage.

PR #270, opened 2026-03-08
Support heterogeneous initial ages in simulation

### LINKED ISSUE #253: ENH: Support heterogeneous initial ages in simulation
All simulated subjects are forced to start at period 0 (`ages.minimum`). The simulation
loop iterates `for period, age in enumerate(ages.values)` from period 0 for every
subject, regardless of the post-#249-mandatory state variable "age".

This is too restrictive for many applications, where we start observing subjects at
different points in their life-cycle.

**Reproducer in #251:** `tests/test_heterogeneous_initial_ages.py`

hmgaudecker:
## Summary
- Subjects can now start simulation at different ages instead of all being forced to period 0 (`ages.minimum`)
- Not-yet-started subjects use a `MISSING_CAT_CODE` sentinel regime ID, which the existing `jnp.where(subjects_in_regime, ...)` guards naturally handle
- Adds `_compute_starting_periods()` helper that converts per-subject initial ages to period indices with validation

Closes #253

## Test plan
- [x] `tests/test_heterogeneous_initial_ages.py` passes (removed `xfail`)
- [x] Full test suite passes (660 passed, 1 skipped)

### DIFF
--- a/tests/simulation/test_compute_starting_periods.py
+++ b/tests/simulation/test_compute_starting_periods.py
@@ -1,0 +1,124 @@
+"""Tests for _compute_starting_periods."""
+
+import jax.numpy as jnp
+import pytest
+
+from lcm import AgeGrid
+from lcm.simulation.simulate import _compute_starting_periods
+
+# =============================================================================
+# Happy path
+# =============================================================================
+
+
+def test_all_subjects_start_at_first_age():
+    ages = AgeGrid(start=25, stop=75, step="Y")
+    initial_ages = jnp.array([25.0, 25.0, 25.0])
+    result = _compute_starting_periods(initial_ages=initial_ages, ages=ages)
+    expected = jnp.array([0, 0, 0])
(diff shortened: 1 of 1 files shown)


src/lcm/simulation/simulate.py
```diff
@@ -315,3 +327,43 @@ def _lookup_values_from_indices(
 # vmap jnp.unravel_index over the first axis of the `indices` argument, while holding
 # the `shape` argument constant (in_axes = (0, None)).
 vmapped_unravel_index = vmap(jnp.unravel_index, in_axes=(0, None))
+
+
+def _compute_starting_periods(
```
timmens: Could you add a few unit tests for this function? Especially for the validation logic.

timmens (changes requested):
Very nice, and super simple implementation actually, nice!

Would only like to see a few unit tests so that it's a bit easier to understand (1) what `_compute_starting_periods` does; and (2) its validation logic.

PR #280, opened 2026-03-20
Add batched productmap

mj023:
### Problem

Sometimes running a model is not possible because of memory restrictions. The nested `vmaps` can lead to JAX creating large arrays for intermediate results, that can be temporarily saved in the GPU memory. Usually these arrays have the dimensions of the State-Action-Space of the model, so looping over batches of half the grid size along one of its dimensions can already halve the peak memory usage. The batching comes at a cost though, the execution time will get progressively worse the smaller the batch size. For big batches the drop in speed is bigger than I would have expected, given that not all the computations can happen at the same time anyways.

### New feature

This PR implements a batched version of `productmap`. The user can for each grid specify the batch size for each states grid. Instead of using `vmap` to map the `Q_and_F_Function` along this grid, `jax.lax.map` will be used, which will then either loop over the batches of gridpoints or if `batch_size=0`, work like `vmap`. The batched version will only be used during the solution, as the State-Action-Space for the simulation is already much smaller, as it only depends on the number of simulated subjects.

### Tasks

- [x] Add batched productmap
- [x] Refactor so not two versions of `_base_productmap` are needed
- [x] Fix typing, tests
- [x] Investigate speed drop

### DIFF
--- a/src/lcm/grids/base.py
+++ b/src/lcm/grids/base.py
@@ -7,6 +7,7 @@ class Grid(ABC):
     """LCM Grid base class."""
 
     @property
+    @abstractmethod
     def batch_size(self) -> int:
         """Size of the batches looped over during the solution.
 
@@ -14,7 +15,6 @@ class Grid(ABC):
         `DiscreteGrid` overrides this via its own property.
 
         """
-        return 0
 
     @abstractmethod
     def to_jax(self) -> Int1D | Float1D:


mj023:
I wasn't yet able to create a good benchmark for a model where batching actually helps. It needs a model that is sufficiently complex, so the compiler can't optimize it well, but it needs to still be runnable somewhat quickly. If using batches actually helps seems to really depend on the model. With Marvins retirement model it nearly cut the memory usage in half, for the Mahler & Yum Model it does very little.

I also fixed an error in the MY model input creation and removed one of the tests, because `productmap` can now handle scalar inputs.

hmgaudecker:
### Code review

Found 7 issues:

1. `batch_size` field leaks into `_ShockGrid.params` — `ContinuousGrid.batch_size` is a dataclass field, but `_ShockGrid._param_field_names` only excludes `"n_points"`. This means every shock grid's `.params` dict includes `batch_size: 0` as if it were a distribution parameter, violating the API contract that `.params` contains "distribution's parameters' names to their specified values."

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/shocks/_base.py#L82-L95

2. `Grid` base class missing `batch_size` attribute — `processing.py` calls `grid.batch_size` on `MappingProxyType[str, Grid]`, but `Grid` has no `batch_size`. The `# ty: ignore[unresolved-attribute]` suppresses the type error rather than fixing the root cause. Adding `batch_size` as an abstract property or concrete field on `Grid` would make the type system enforce the contract.

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/regime_building/processing.py#L1267-L1269

3. `batch_sizes` parameter missing from `get_max_Q_over_a` docstring Args section (AGENTS.md says "Google-style docstrings" with all parameters documented)

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/regime_building/max_Q_over_a.py#L21-L57

4. Multiple stale `vmap` references in `dispatchers.py` — docstrings say "Like vmap" and "arguments over which we apply vmap", comments say "just vmap over all vars", and the result variable is named `vmapped`, but the implementation uses `jax.lax.map`. The `_base_productmap_batched` docstring also says "Cannot have keyword-only arguments" but the implementation actively handles them via `**kwargs` closures.

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/utils/dispatchers.py#L221-L232

5. No tests for `batch_size > 0` — the core new feature (memory-saving batched execution) has zero tests verifying correct numerical output when `batch_size > 0`. All test changes only use `batch_size=0`.

6. Action grid `batch_size` values are silently ignored — `processing.py` builds `batch_sizes` from all grids (states + actions), but `get_max_Q_over_a` only passes `batch_sizes` to the outer `productmap` over `state_names` (line 84), not the inner `productmap` over `action_names` (line 64). Similarly, `simulation_spacemap` hardcodes `batch_sizes=dict.fromkeys(action_names, 0)`. Users setting `batch_size` on action grids would see no effect.

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/regime_building/max_Q_over_a.py#L60-L85

7. Stale cross-reference in docstring — `_base_productmap_batched` Returns section says "See `product_map` for details" but the function is named `productmap` (no underscore).

https://github.com/OpenSourceEconomics/pylcm/blob/c729e8e9eeec0bb0216cac0b965109728c4f0f7e/src/lcm/utils/dispatchers.py#L228-L231

PR #282, opened 2026-03-22
Use integer ages for annual/multiannual models

hmgaudecker:
## Summary

`AgeGrid` now auto-detects whether ages are integer-valued. Annual/multiannual
grids (e.g. `step="Y"`, `step="10Y"`) produce `int32` arrays and `int` scalars;
sub-annual grids (`step="Q"`, `step="M"`) produce floats as before. This lets
user functions annotate `age: int` in annual models.

- `AgeGrid.__init__` checks if all exact values are integer-valued after
  computing them; if so, stores `int` exact values and an `int32` JAX array
- Add `is_integer` property, `age_to_period()` method, and `_age_to_period_map`
- `period_to_age()` returns `int` or `float` based on `is_integer`
- `get_periods_where()` passes `int` or `float` to predicates accordingly
- Add `type Age = int | float` in `typing.py`; widen `age` parameter types in
  `error_handling.py` and `logging.py` to `ScalarInt | ScalarFloat`
- Replace ad-hoc `age_to_period` dicts in `validation.py` and `pandas_utils.py`
  with the new `AgeGrid.age_to_period()` method

## Test plan

- [x] 32 dedicated tests in `test_ages.py` (int auto-detection, float
  sub-annual, period_to_age/age_to_period types, integration with Model)
- [x] Full test suite passes (784 tests)
- [x] ty type checking passes
- [x] Pre-commit hooks pass


hmgaudecker:
Cool. The main gain is really being able to use `age: int` in all type annotations of functions that consume it.

The rest is not terribly important indeed!

PR #305, opened 2026-04-08
More sensible tracking of GPU peak memory usage in benchmarks

hmgaudecker:
- Fix `track_gpu_peak_mem` benchmarks to actually measure GPU peak memory in isolation.
  The previous implementation read `peak_bytes_in_use` from the parent ASV process after
  `setup()` had already run, which returned the cumulative high-water mark of the ASV
  process rather than the peak of a single benchmark run. Since ASV reuses one process
  for all benchmarks, this number grew monotonically and mixed memory from unrelated
  benchmarks.
- The new implementation spawns an isolated subprocess per measurement via
  `benchmarks/_gpu_mem.py`. The subprocess builds the model from scratch, runs it once
  cold (compilation + execution), and reports the true peak. A `GpuPeakMem` base class
  provides dedicated ASV benchmark classes with no-op setup so the parent process never
  touches the GPU.
- Use on-demand GPU memory allocation instead of JAX's default of pre-allocating
  75% of GPU memory. This plays nicely with other GPU processes, makes nvidia-smi
  reflect actual usage, and enables meaningful GPU memory benchmarks. Users can
  override by setting XLA_PYTHON_CLIENT_PREALLOCATE=true before importing lcm.
- Fix `productmap` calls in `function_representation.ipynb` and `dispatchers.ipynb` that broke
  after `batch_sizes` became required (#280).

### DIFF
--- a/benchmarks/_gpu_mem.py
+++ b/benchmarks/_gpu_mem.py
@@ -1,0 +1,107 @@
+"""Measure GPU peak memory for a benchmark in an isolated subprocess.
+
+ASV runs all benchmarks in one process, so ``peak_bytes_in_use`` accumulates
+across runs and warm-up calls.  This module spawns a fresh Python process that
+builds the model, runs it once cold (compilation + execution), and reports the
+true peak.
+
+The ``GpuPeakMem`` base class provides a ready-made ASV benchmark with a no-op
+``setup`` so the parent process does not touch the GPU before spawning the
+subprocess.  Subclass it and set ``bench_module`` / ``bench_class``::
+
+    class MahlerYumGpuPeakMem(GpuPeakMem):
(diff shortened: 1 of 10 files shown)


mj023 (approved):
The changes look fine overall, but I'm wondering a little why the benchmark results don't seem to change, when there was supposedly some memory accumulation through different benchmarks.

hmgaudecker:
@mj023 Good catch. The numbers were identical because the `GpuPeakMem` benchmark classes are new (different class names from the old ones on main). The `_backfill_base_results` logic was copying HEAD values into the base results file for missing benchmarks, so the comparison was HEAD vs. itself.

Fixed in d89ab25: removed the backfill logic. Benchmarks without a baseline now show empty before/ratio columns instead of fake identical numbers.

PR #306, opened 2026-04-08
Accept shock grid columns in initial_conditions_from_dataframe

hmgaudecker:
## Summary

`initial_conditions_from_dataframe` rejected shock grid state columns (e.g. `productivity_shock` from a `Rouwenhorst` grid) as "unknown columns", even though `validate_initial_conditions` in `simulate()` requires them. This made it impossible to use the DataFrame API with models that have shock grids without a workaround.

Shock grid states are continuous states — they should pass through as float columns, just like `wealth`.

## Changes

- Split `_collect_all_state_names` into `_collect_state_names` returning `(required, optional)` sets
- Required: non-shock states + age (must be present in the DataFrame)
- Optional: shock grid states (accepted if present, not flagged as missing)
- `_validate_state_columns` updated to use the new required/optional split

## Test plan

- [x] `test_shock_state_columns_accepted` — DataFrame with shock columns converts without error
- [x] `test_shock_state_columns_optional` — DataFrame without shock columns also works
- [x] Full `test_pandas_utils.py` suite (69/69 pass)
- [x] `prek run --all-files`
- [x] `pixi run -e type-checking ty`

### DIFF
--- a/src/lcm/pandas_utils.py
+++ b/src/lcm/pandas_utils.py
@@ -800,25 +800,19 @@ def _validate_state_columns(
     initial_regimes: list[str],
 ) -> None:
     """Validate that DataFrame columns match model states."""
-    required, optional = _collect_state_names(
-        regimes=regimes, initial_regimes=initial_regimes
-    )
-    all_known = required | optional
+    expected = _collect_state_names(regimes=regimes, initial_regimes=initial_regimes)
 
-    unknown = state_columns - all_known
+    unknown = state_columns - expected
     if unknown:
         msg = (
             f"Unknown columns not matching any model state: {sorted(unknown)}. "
-            f"Expected states: {sorted(all_known)}."
+            f"Expected states: {sorted(expected)}."
         )
         raise ValueError(msg)
 
-    missing = required - state_columns
+    missing = expected - state_columns
     if missing:
-        msg = (
-            f"Missing required state columns: {sorted(missing)}. "
-            f"All non-shock states must be provided."
-        )
+        msg = f"Missing required state columns: {sorted(missing)}."
         raise ValueError(msg)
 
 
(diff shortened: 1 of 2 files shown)


hmgaudecker:
### Code review

Found 3 issues:

1. **Inconsistency between `initial_conditions_from_dataframe` and `validate_initial_conditions`:** This PR makes shock states optional in `initial_conditions_from_dataframe` (the test `test_shock_state_columns_optional` omits shock columns and passes). But `validate_initial_conditions` in `simulate()` uses `variable_info.query("is_state")`, which includes shock states as required. A user who omits shock columns from their DataFrame will get a dict without those keys from `initial_conditions_from_dataframe`, then `model.simulate()` will raise `InvalidInitialConditionsError` saying the shock state is missing. Either `validate_initial_conditions` also needs to treat shock states as optional, or `initial_conditions_from_dataframe` should keep requiring them. (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/6d7fb9688d06849861b258f4b2748f25301a7973/src/lcm/pandas_utils.py#L818-L822

2. **No test for heterogeneous initial regimes where one regime has a shock state and the other does not.** Both new tests use a single-regime DataFrame (`regime=["alive", "alive"]`). The `_collect_state_names` function iterates over `set(initial_regimes)` and accumulates required/optional states across all regimes, but this cross-regime logic is untested. (Score: 65)

3. **Docstring imprecision in `_collect_state_names`.** Says "Required includes all non-shock states plus age" but `"age"` is hardcoded into the `required` set before the regime loop, not collected from regime states. The phrasing implies age comes from regime states. (Score: 50)

https://github.com/OpenSourceEconomics/pylcm/blob/6d7fb9688d06849861b258f4b2748f25301a7973/src/lcm/pandas_utils.py#L830-L835

mj023 (approved):
All looks good, I wonder why the benchmark is sometimes so inconsistent, there is nothing here that should affect it.

hmgaudecker:
> I wonder why the benchmark is sometimes so inconsistent,

Probably load on the GPU. I was sending other stuff there while these PRs came in.

PR #307, opened 2026-04-09
Fix missing discount type assignment for high-education agents

hmgaudecker:
## Summary

The type-assignment loop in `create_inputs` set `discount[index]` for indices 0–7 (low education) but never set `discount[index + 8]` for indices 8–15 (high education). All high-education agents were incorrectly assigned to the low discount factor type (default 0).

One-line fix: add `discount = discount.at[index + 8].set(i - 1)`.

Found by comparing against the original replication code (Replication_MY2024) which correctly sets discount type for all 16 type indices.

## Test plan

- [x] `prek run --all-files`


hmgaudecker:
will be superseded soon.

PR #314, opened 2026-04-09
Enable persistent JAX compilation cache by default

hmgaudecker:
## Summary

- Set `XLA_PYTHON_CLIENT_PREALLOCATE=false` on import — disables JAX's
  default of reserving 75% of GPU memory upfront. Makes `nvidia-smi`
  reflect actual usage and plays nicely with other GPU processes.
- Enable persistent JAX compilation cache at `~/.cache/jax` by default.
  Large models can take minutes to compile; the cache makes subsequent
  runs near-instant. The default is skipped (rather than crashing at
  import time) on systems where `Path.home()` raises — notably HPC
  containers without `HOME`/`USERPROFILE`. An explicitly-empty
  `JAX_COMPILATION_CACHE_DIR` is treated as unset.
- Add JAX Settings section to installation docs with HPC override
  instructions.

Users can override both settings via environment variables before
importing `lcm`.

## Test plan

- [x] Full test suite passes
- [x] `ty` clean
- [x] `prek` clean

Generated with Claude Code.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -6,7 +6,7 @@ concurrency:
 on:
   pull_request:
     branches:
-      - '*'
+      - '**'
   workflow_dispatch: null
 jobs:
   run-benchmarks:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
       - main
   pull_request:
     branches:
-      - '*'
+      - '**'
 jobs:
   run-tests:
     name: Run tests for ${{ matrix.os }} on ${{ matrix.python-version }}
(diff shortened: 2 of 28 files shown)


hmgaudecker:
### Code review (relative to PR #316)

Found 1 issue:

1. **`$USER` shell variable not expanded in Python code example.** The HPC override snippet uses `"/scratch/$USER/.cache/jax"` which is a literal string in Python — `$USER` is not expanded. A user copying this verbatim gets a broken cache path. Should use `os.path.expandvars("/scratch/$USER/.cache/jax")` or `f"/scratch/{os.environ['USER']}/.cache/jax"`. (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/d05aa5da17c04f50294e6813e72f820bf26a6b18/docs/user_guide/installation.md#L66-L69

mj023 (approved):
All good, great that the environment variables we set are now mentioned in the docs.

PR #315, opened 2026-04-09
Support heterogeneous state sets in initial conditions

hmgaudecker:
## Summary

Support models where different regimes have different state sets in
`initial_conditions_from_dataframe` — e.g., a `claimed_ss` state that only
exists for retired subjects.

**`pandas_utils.py`:**
- `initial_conditions_from_dataframe`: skip columns that aren't states of
  the current regime (fixes NA/string conversion errors for regime-specific
  states).
- Pre-allocate result arrays with NaN so unused slots surface bugs.
- Replace remaining NaN in discrete columns with `MISSING_CAT_CODE`
  (`jnp.iinfo(jnp.int32).min`, imported from `simulation.initial_conditions`
  — single source of truth) before the int32 cast, so JAX indexing
  produces obviously wrong values instead of silently selecting the last
  element.
- `_validate_state_columns`: "Missing required state columns" error names
  which regime(s) require each missing column. The message for
  pseudo-states vs regime-required vs fallback cases is now formatted via
  `_format_missing_state_detail`; "Unknown columns" clarifies it's about
  columns not matching any state of an *initial* regime.

**`simulation/initial_conditions.py`:**
- `_validate_discrete_state_values`: validate each discrete state only for
  subjects in regimes that actually have it. Previously the sentinel for
  missing states was falsely rejected.
- `_raise_feasibility_type_error` (`-> Never`): adds diagnostic context
  when a feasibility check `TypeError` is likely caused by a discrete
  state with the wrong dtype (e.g. float where int is expected).

**`ages.py`:**
- New module-level `PSEUDO_STATE_NAMES = frozenset({"age"})` constant.
  Used by both the validator and the data-loader so the two paths stay
  in sync on what counts as a pseudo-state.

## Test plan

- [x] `test_initial_conditions_heterogeneous_state_sets` — regimes with
      different state sets.
- [x] `test_initial_conditions_shock_grid_heterogeneous_state_sets` — one
      regime has a shock state, another does not.
- [x] Heterogeneous discrete grids in `test_regime_state_mismatch.py`.

Generated with Claude Code.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -6,7 +6,7 @@ concurrency:
 on:
   pull_request:
     branches:
-      - '*'
+      - '**'
   workflow_dispatch: null
 jobs:
   run-benchmarks:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
       - main
   pull_request:
     branches:
-      - '*'
+      - '**'
 jobs:
   run-tests:
     name: Run tests for ${{ matrix.os }} on ${{ matrix.python-version }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
     hooks:
       - id: check-github-workflows
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.9
+    rev: v0.15.10
     hooks:
       - id: ruff-check
         args:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -223,7 +223,7 @@ Model(
 ### Derived Categoricals
 
 When parameters are indexed by a DAG function output (not a model state/action), declare
-`derived_categoricals={"name": DiscreteGrid(...)}` on the `Regime` that uses it. For
+`derived_categoricals={"name": CategoryClass}` on the `Regime` that uses it. For
 convenience, model-level `derived_categoricals` on `Model(...)` are broadcast to all
 regimes. Functions used as derived categoricals must return **integer** types, not
 booleans — JAX cannot use booleans as array indices inside JIT. Use `jnp.int32(...)` to
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -222,10 +222,12 @@ Model(
 
 ### Derived Categoricals
 
-When `solve()` / `simulate()` parameters are indexed by a DAG function output (not a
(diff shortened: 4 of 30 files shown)


hmgaudecker:
### Code review (relative to PR #314)

Found 4 issues:

1. **NaN-to-int32 cast for discrete states relies on undefined behavior.** Discrete-state columns are pre-allocated as `np.full(n_subjects, np.nan)` (float). For subjects whose regime lacks the state, NaN remains. The cast `jnp.array(arr, dtype=jnp.int32)` converts NaN to INT32_MIN with a `RuntimeWarning`. This happens to match `MISSING_CAT_CODE` by coincidence, but relies on platform-undefined NaN-to-int cast behavior. Should pre-allocate integer buffers with `np.iinfo(np.int32).min` for known-discrete columns. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/0ec706a9267f387e7a1d26d34f6acde59ee22174/src/lcm/pandas_utils.py#L96-L140

2. **`_raise_feasibility_type_error` raises `TypeError` instead of `InvalidInitialConditionsError`.** All other validation failures raise `InvalidInitialConditionsError` (documented in `validate_initial_conditions` docstring). This function breaks the contract — callers catching `InvalidInitialConditionsError` would miss this error. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/0ec706a9267f387e7a1d26d34f6acde59ee22174/src/lcm/simulation/initial_conditions.py#L653-L683

3. **`_raise_feasibility_type_error` missing `Args:` docstring section.** Has 4 parameters but only a one-liner docstring. All other non-trivial functions in the same file document their parameters per Google-style convention. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/0ec706a9267f387e7a1d26d34f6acde59ee22174/src/lcm/simulation/initial_conditions.py#L653-L660

4. **Test doesn't assert sentinel value for heterogeneous discrete state.** `test_initial_conditions_heterogeneous_state_sets` comments that `result["status"][2]` is "unused (NaN->int32 gives a sentinel)" but never asserts the actual value. A future change that accidentally writes a real value to that slot would not be caught. (Score: 50)

https://github.com/OpenSourceEconomics/pylcm/blob/0ec706a9267f387e7a1d26d34f6acde59ee22174/tests/test_pandas_utils.py#L470-L475

hmgaudecker (commented):
Autoreview.

hmgaudecker:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

hmgaudecker (commented):
Autoreview.

PR #316, opened 2026-04-10
Skip unreachable target regimes in continuation value loop

hmgaudecker:
## Summary

Detect and reject user errors where a regime's `state_transitions` does not
cover all reachable target regimes, pre-solve. Previously these cases
produced opaque NaN propagation or shape mismatches during the solve.

A quick glossary for this PR:
- **Regime transition** — `regime.transition`, the function that returns
  the next regime.
- **State transition** — `regime.state_transitions[state_name]`, the
  function that returns `next_<state_name>`.
- **Per-target-regime state transition** — a state transition supplied
  as a dict keyed by target regime name, i.e.
  `state_transitions["health"] = {"working": ..., "retired": ...}`.

**Reachable-target derivation (`processing.py`):**
- `_get_reachable_targets`: when at least one state has per-target-regime
  state transitions, start from the explicitly named target regimes and
  add any target whose state needs are covered by simple (non-per-target)
  transitions. Without per-target-regime state transitions, all regimes
  are reachable.
- `_extract_transitions_from_regime`: only build transition entries for
  reachable target regimes.
- `_process_regime_core`: scope shock-transition construction to reachable
  target regimes, avoiding spurious entries for unreachable regimes.

**Complete/incomplete target partitioning (`Q_and_F.py`):**
- `get_Q_and_F` enumerates all regimes active in the next period from
  `regime_to_v_interpolation_info` (not just those in `transitions`, so
  completely-absent targets are detected). Targets whose stochastic state
  needs are all covered by `transitions` go into `complete_targets`; the
  traced function loops only over these.

**Pre-solve validation (`error_handling.py`):**
- `_validate_no_reachable_incomplete_targets`: for every active target
  (including self-transitions) whose stochastic state transitions are not
  fully covered, raise `InvalidRegimeTransitionProbabilitiesError` when
  `regime_transition_probs[target] > 0`. The error lists the missing
  `next_*` transitions and tells the user what to fix.
- `validate_V` gains a bullet for the incomplete-target case.

Moving detection outside the JIT-traced function keeps the HLO (and the
persistent compilation-cache key) independent of the complete/incomplete
partition — an earlier iteration of this PR that put the check inside the
trace invalidated the compilation cache on every change.

## Test plan

- [x] `test_incomplete_per_target_reachable_target` — raises
      `InvalidRegimeTransitionProbabilitiesError` when a source regime
      can reach a target it does not list in its per-target-regime dict.
- [x] `test_incomplete_target_zero_prob_succeeds` — solve succeeds (and
      V arrays are NaN-free) when an incomplete target has zero
      transition probability.

Generated with Claude Code.

### DIFF
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -69,9 +69,9 @@ def get_Q_and_F(
     # Enumerate all active targets, not just those in transitions — targets
     # entirely absent from per-target dicts must also be detected.
     all_active_next_period = tuple(
-        name
-        for name in regime_to_v_interpolation_info
-        if period + 1 in regimes_to_active_periods.get(name, ())
+        regime_name
+        for regime_name in regime_to_v_interpolation_info
+        if period + 1 in regimes_to_active_periods.get(regime_name, ())
     )
 
     # Keep only targets whose stochastic state needs are all covered by
@@ -79,15 +79,17 @@ def get_Q_and_F(
     # from the traced function; pre-solve validation in
     # `_validate_no_reachable_incomplete_targets` raises if any such target
     # has non-zero transition probability.
-    complete_targets: list[str] = []
-    for name in all_active_next_period:
+    complete_targets: list[RegimeName] = []
+    for regime_name in all_active_next_period:
         target_stochastic_needs = {
             f"next_{s}"
-            for s in regime_to_v_interpolation_info[name].state_names
+            for s in regime_to_v_interpolation_info[regime_name].state_names
             if f"next_{s}" in stochastic_transition_names
         }
-        if name in transitions and target_stochastic_needs.issubset(transitions[name]):
-            complete_targets.append(name)
+        if regime_name in transitions and target_stochastic_needs.issubset(
+            transitions[regime_name]
+        ):
+            complete_targets.append(regime_name)
 
     next_V_extra_param_names: dict[str, frozenset[str]] = {}
 
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -76,9 +76,10 @@ def get_Q_and_F(
 
     # Keep only targets whose stochastic state needs are all covered by
     # `transitions`. Targets with missing stochastic transitions are dropped
-    # from the traced function; pre-solve validation in
-    # `_validate_no_reachable_incomplete_targets` raises if any such target
-    # has non-zero transition probability.
+    # from the traced function; `validate_regime_transitions_all_periods`
+    # (via `_validate_no_reachable_incomplete_targets` in
+    # `lcm.utils.error_handling`) raises pre-solve if any such target has
+    # non-zero transition probability.
     complete_targets: list[RegimeName] = []
     for regime_name in all_active_next_period:
         target_stochastic_needs = {
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -1,9 +1,7 @@
-import logging
-from collections.abc import Callable, Mapping
+from collections.abc import Callable
 from types import MappingProxyType
 from typing import Any, cast
 
-import jax
 import jax.numpy as jnp
 from dags import concatenate_functions, with_signature
 from jax import Array
@@ -28,7 +26,7 @@ from lcm.utils.dispatchers import productmap
 from lcm.utils.functools import get_union_of_args
 
 
-def get_Q_and_F(  # noqa: C901, PLR0915
+def get_Q_and_F(
     *,
     flat_param_names: frozenset[str],
     age: float,
@@ -162,27 +160,6 @@ def get_Q_and_F(  # noqa: C901, PLR0915
         exclude=frozenset({"period", "age"}),
     )
 
-    # Guard callback for incomplete targets — defined at closure scope so JAX
-    # sees the same function object across calls (avoids JIT re-compilation).
-    # Only active when the logger is at DEBUG level; otherwise a no-op.
-    if incomplete_targets:
-
-        def _check_zero_probs(probs: Mapping[str, Array]) -> None:
-            """Validate that incomplete targets have zero transition probability."""
-            if not logging.getLogger("lcm").isEnabledFor(logging.DEBUG):
-                return
-            for target in incomplete_targets:
-                prob = float(probs[target])
-                if prob > 0:
-                    msg = (
-                        f"Regime transition probability to '{target}' "
-                        f"is {prob} > 0, but no stochastic state "
-                        f"transition was provided for this target. "
-                        f"Add the missing entries to the per-target "
-                        f"dict in state_transitions."
-                    )
-                    raise ValueError(msg)
-
     @with_signature(
         args=arg_names_of_Q_and_F, return_annotation="tuple[FloatND, BoolND]"
     )
@@ -212,10 +189,12 @@ def get_Q_and_F(  # noqa: C901, PLR0915
             period=period,
             age=age,
         )
-        # Filter to active regimes only — inactive regimes must have 0
-        # probability (validated before solve).
+        # Use only complete targets for the traced function — incomplete
+        # target validation happens outside JIT to keep the HLO (and thus
+        # the persistent compilation cache key) independent of the
+        # partition.
         active_regime_probs = MappingProxyType(
-            {r: regime_transition_probs[r] for r in all_active_next_period}
+            {r: regime_transition_probs[r] for r in complete_targets}
         )
 
         E_next_V = jnp.zeros_like(U_arr)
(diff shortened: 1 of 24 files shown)


hmgaudecker:
### Code review

Found 5 issues:

1. **Comment claims incomplete targets are "unreachable" but the code adds a runtime check precisely because this cannot be guaranteed statically.** The comment should say something like "assumed to be unreachable (validated at runtime via `_check_zero_probs`)" rather than flatly declaring them unreachable. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/688f4648c7b56aef460c019953c42df21e6e8fab/src/lcm/regime_building/Q_and_F.py#L76-L80

2. **`jax.debug.callback` may silently drop exceptions under JIT.** The callback is documented as "innocuous" — its effect can be dropped, duplicated, or reordered by JAX transformations. If a user misconfigures a model where an incomplete target has non-zero transition probability, the `ValueError` might not propagate, leading to silently wrong results. Consider `jax.experimental.io_callback` or a build-time check instead. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/688f4648c7b56aef460c019953c42df21e6e8fab/src/lcm/regime_building/Q_and_F.py#L213-L215

3. **No test added for the new code path.** The partition into complete/incomplete targets and the `_check_zero_probs` runtime guard have no corresponding test. A regression in the partitioning logic would go undetected. (Score: 75)

4. **Missing docstring on `_check_zero_probs`.** CLAUDE.md requires Google-style docstrings; all other private helpers in the same file have them. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/688f4648c7b56aef460c019953c42df21e6e8fab/src/lcm/regime_building/Q_and_F.py#L165-L177

5. **`dict[str, Array]` annotation on `probs` parameter instead of `Mapping[str, Array]`.** CLAUDE.md says "Type annotations use `Mapping` for read-only dict-like interfaces" and AGENTS.md says "Use `collections.abc` for abstract types: `Sequence`, `Mapping`, `Iterable`". The function only reads from the dict. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/688f4648c7b56aef460c019953c42df21e6e8fab/src/lcm/regime_building/Q_and_F.py#L165

hmgaudecker:
### Code review (relative to PR #321)

Found 3 issues:

1. **`io_callback` missing `ordered=True`.** Without it, JAX may reorder the validation callback relative to the computation that uses the (potentially invalid) transition probabilities. For a safety check that should fire before wrong values propagate, `ordered=True` is the correct semantic. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/f9593cd749f76ec1a40b30427583bd544f27064f/src/lcm/regime_building/Q_and_F.py#L189-L192

2. **One-assertion-per-test violation.** `test_incomplete_target_skipped_and_zero_prob_validated` tests two distinct behaviors: (a) solve succeeds with zero probability to incomplete target, (b) raises with non-zero probability. AGENTS.md says "One assertion per test." Should be split into two test functions. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/f9593cd749f76ec1a40b30427583bd544f27064f/tests/test_Q_and_F.py#L267-L394

3. **`io_callback` fires per scalar state-action evaluation.** `_check_zero_probs` is inside `Q_and_F`, which is called N x M times through `productmap`. The validation runs tens of thousands of times with identical arguments instead of once. Each call does a host synchronization. For realistic models (~50k state-action combinations), this adds ~4s of overhead per regime per period. (Score: 55)

https://github.com/OpenSourceEconomics/pylcm/blob/f9593cd749f76ec1a40b30427583bd544f27064f/src/lcm/regime_building/Q_and_F.py#L189-L192

hmgaudecker:
Note: Tests run locally, for cascading PRs we need the `"**"` wildcard in CI, which I only added in #314

src/lcm/regime_building/Q_and_F.py
```diff
@@ -173,11 +210,19 @@ def Q_and_F(
...
+            {r: regime_transition_probs[r] for r in all_active_next_period}
         )
 
+        if incomplete_targets:
+            jax.experimental.io_callback(
+                _check_zero_probs,
+                None,
+                dict(active_regime_probs),
+                ordered=True,
+            )
```
mj023: Just from looking at the [Jax Docs](https://docs.jax.dev/en/latest/external-callbacks.html#exploring-io-callback) I am wondering why this is not throwing an error, as they state that `ordered=True` inside `vmap` is not allowed. Also from a performance standpoint I am worried that this might be costly, I think the benchmarks are not covering a case where incomplete targets exist, right?

src/lcm/regime_building/Q_and_F.py
```diff
@@ -67,14 +69,31 @@ def get_Q_and_F(
...
+    for name in all_active_next_period:
+        target_stochastic_needs = {
+            f"next_{s}"
+            for s in regime_to_v_interpolation_info[name].state_names
+            if f"next_{s}" in stochastic_transition_names
+        }
+        if target_stochastic_needs.issubset(transitions[name]):
+            complete_targets.append(name)
+        else:
+            incomplete_targets.append(name)
```
mj023: It seems like `incomplete_targets` is always empty in the tests below.

tests/test_Q_and_F.py
```diff
@@ -255,3 +261,126 @@ def utility_func(
...
+    def _next_regime(age: float) -> ScalarInt:
+        return jnp.where(
+            age >= 2, _IncompleteTargetRegimeId.dead, _IncompleteTargetRegimeId.work
+        )
+
+    model, params = _build_incomplete_target_model(next_regime_func=_next_regime)
+    model.solve(params=params)
+
+
+@pytest.mark.xfail(
```
mj023: We should try to get this test to work, otherwise the feature is not properly tested I think.

src/lcm/regime_building/Q_and_F.py
hmgaudecker: Good catch on both points.

**Why it doesn't throw:** `Q_and_F` is wrapped by `productmap`, which uses
`jax.lax.map` (sequential), not `jax.vmap`. So the `ordered=True`
restriction doesn't apply here. But it's still an experimental API with
unreliable error propagation — which is why the xfail test exists.

**Fixed in e8618f0:** Replaced `io_callback` with a two-layer approach:
1. `jax.debug.callback` (stable API, only active at `log_level="debug"`)
  for a specific error message.
2. `jnp.where` NaN-poisoning (always active) as a reliable fallback —
  pure JAX, zero callback overhead, caught by `validate_V` outside JIT.

The `if incomplete_targets:` guard is a Python-level check at trace time,
so neither layer is compiled into the function when there are no incomplete
targets (the common case). Should not have any performance impact AFAICT.

src/lcm/regime_building/Q_and_F.py
hmgaudecker: Correct — and not just a test gap, it was structural. `_partition_targets`
enumerated targets from `tuple(transitions)`, but
`_extract_transitions_from_regime` already removed absent targets from
`transitions`. So a target missing from the per-target dict was invisible
and `incomplete_targets` was always empty.

Fixed in e8618f0: `_partition_targets` now enumerates all active targets
from `regime_to_v_interpolation_info` (not just those in `transitions`).
Targets absent from `transitions` but with stochastic needs are correctly
classified as incomplete. Also added `_get_reachable_targets` to include
targets fully served by simple transitions.

tests/test_Q_and_F.py
hmgaudecker: Agreed — fixed in e8618f0:
- Removed the xfail. The NaN-poisoning path reliably raises
  `InvalidValueFunctionError` on all backends.
- Removed this duplicate test from `test_Q_and_F.py` — the same scenario
  is covered by `test_incomplete_per_target_reachable_target` in
  `test_regime_state_mismatch.py` (also has its xfail removed).
- Kept `test_incomplete_target_zero_prob_succeeds` here since it's a
  Q_and_F unit concern.

src/lcm/regime_building/Q_and_F.py
mj023: `jax.lax.map` uses `vmap` for batches, so it should still throw the error, but thaths not our problem I guess.

I am just concerned that larger models with incomplete targets incur a lot of overhead because of the callbacks, as there is one callback for each state. Another option would be to make a new static argument `debug=True` and pass it to each jitted function and only run the callback if the debug mode is activated. Could also be useful in other cases, but maybe we can try it like this and if we run into problems then we can try a bigger overhaul of the debug functionality.

tests/test_Q_and_F.py
mj023: I think just the xfail test from `test_Q_and_F.py` was deleted, at least I can't find the `test_incomplete_per_target_reachable_target` test and the changes do not show that xfail was removed anywhere.

src/lcm/regime_building/Q_and_F.py
hmgaudecker: The problem with debug as a static argument is that compilation is pretty slow on large models -- so each time you'd want to use debug mode, you'd recompile. 

I'd say let's try like this for now.

tests/test_Q_and_F.py
hmgaudecker: Oh, sorry -- looks like both me and Claude got confused in the cascade. The test has been added in #315 (downstream from here), without xfail now:                                                                                                                                   
                                                                                                                                                                                                             
https://github.com/OpenSourceEconomics/pylcm/pull/315/changes#diff-3a5da5bc4a380922c6d568f35a727347c54feb5a65233bc0db9cd94e3180e993R617

mj023 (approved):

src/lcm/regime_building/Q_and_F.py
hmgaudecker: You were completely right, ofc! This caused the performance regression we chatted about elsewhere...

`incomplete_targets` being non-empty changes the traced function via:

1. `if incomplete_targets`: → callback + NaN-poisoning compiled in
2. `all_active_next_period` = `(*complete_targets, *incomplete_targets)` → `active_regime_probs` dict traced with more keys

Both change the HLO → cache invalidation

tests/test_Q_and_F.py
hmgaudecker: Pulled it in here now in b168b84.

hmgaudecker (commented):
Autoreview.

PR #317, opened 2026-04-10
Lazy NaN diagnostics for value function validation

hmgaudecker:
## Summary

When `validate_V` detects a NaN in the computed value function, run a
diagnostic pass that pinpoints which intermediate (U, F, E[V], Q) is the
source. Zero overhead on the normal (no-NaN) solve path.

**Diagnostic closure (`Q_and_F.py`, `processing.py`):**
- `get_compute_intermediates` mirrors `get_Q_and_F` but returns all
  intermediates. Its target enumeration matches `get_Q_and_F`: iterate
  `regime_to_v_interpolation_info`, keep targets whose stochastic state
  needs are all in `transitions`. The returned scalar closure is typed
  `tuple[FloatND, FloatND, FloatND, FloatND, MappingProxyType[RegimeName, Array]]`.
- `_productmap_over_state_action_space` wraps the scalar closure over
  the full state-action space (same outer-states / inner-actions pattern
  as `max_Q_over_a`).
- `_wrap_with_reduction` fuses the productmap'd closure with on-device
  reductions (`jnp.isnan` → `jnp.mean` over per-dimension axes) in a
  single JIT region. This keeps the full-shape U/F/E/Q arrays from ever
  materialising in host-visible memory.
- `_build_compute_intermediates_per_period` builds one fused closure per
  period, stored on `SolveFunctions.compute_intermediates` and invoked
  lazily only when NaN is detected.

**Error enrichment (`error_handling.py`):**
- `validate_V`: when given `compute_intermediates` and `state_action_space`,
  calls `_enrich_with_diagnostics`. Flat regime params that collide with
  state/action names are filtered out before splatting into kwargs.
- `_summarize_diagnostics` restructures the flat reduction pytree into the
  nested `{"U_nan_fraction": {"overall": ..., "by_dim": {...}}, ...}`
  summary. Pure host-side — no device computation.
- `_format_diagnostic_summary` turns the summary into a multi-line note
  attached to the raised `InvalidValueFunctionError`.
- Diagnostic failures are caught and logged — the original NaN error still
  propagates.

**Snapshot on failure (`solve_brute.py`, `model.py`):**
- Partial solution arrays are attached to the exception and saved via
  `save_solve_snapshot` when `log_path` is set, so users can resume
  debugging from the point of failure.

**Simulation log formatting (`logging.py`):**
- `log_regime_transitions` now prints transitions as a bulleted list.

Debugging guide added at `docs/user_guide/debugging.md`.

## Test plan

- [x] `test_nan_diagnostics.py` covers lazy enrichment and graceful
      degradation when the diagnostic path itself fails.

Generated with Claude Code.

### DIFF
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -255,6 +255,7 @@ def get_compute_intermediates(
     *,
     age: float,
     period: int,
+    flat_param_names: frozenset[str],
     functions: FunctionsMapping,
     constraints: FunctionsMapping,
     transitions: TransitionFunctionsMapping,
@@ -272,6 +273,9 @@ def get_compute_intermediates(
     Args:
         age: The age corresponding to the current period.
         period: The current period.
+        flat_param_names: Frozenset of flat parameter names for the regime;
+            used to build the explicit signature via `with_signature` so
+            productmap can introspect the closure correctly.
         functions: Immutable mapping of function names to internal user functions.
         constraints: Immutable mapping of constraint names to constraint functions.
         transitions: Immutable mapping of target regime names to state transition
@@ -362,6 +366,24 @@ def get_compute_intermediates(
         get_union_of_args([_H_func]) - {"utility", "E_next_V"}
     )
 
+    arg_names_of_compute_intermediates = _get_arg_names_of_Q_and_F(
+        [
+            U_and_F,
+            compute_regime_transition_probs,
+            *list(state_transitions.values()),
+            *list(next_stochastic_states_weights.values()),
+        ],
+        include=frozenset({"next_regime_to_V_arr"} | flat_param_names),
+        exclude=frozenset({"age", "period"}),
+    )
+
+    @with_signature(
+        args=arg_names_of_compute_intermediates,
+        return_annotation=(
+            "tuple[FloatND, FloatND, FloatND, FloatND, "
+            "MappingProxyType[RegimeName, Array]]"
+        ),
+    )
     def compute_intermediates(
         next_regime_to_V_arr: FloatND,
         **states_actions_params: Array,
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -69,9 +69,9 @@ def get_Q_and_F(
     # Enumerate all active targets, not just those in transitions — targets
     # entirely absent from per-target dicts must also be detected.
     all_active_next_period = tuple(
-        name
-        for name in regime_to_v_interpolation_info
-        if period + 1 in regimes_to_active_periods.get(name, ())
+        regime_name
+        for regime_name in regime_to_v_interpolation_info
+        if period + 1 in regimes_to_active_periods.get(regime_name, ())
     )
 
     # Keep only targets whose stochastic state needs are all covered by
@@ -79,15 +79,17 @@ def get_Q_and_F(
     # from the traced function; pre-solve validation in
     # `_validate_no_reachable_incomplete_targets` raises if any such target
     # has non-zero transition probability.
-    complete_targets: list[str] = []
-    for name in all_active_next_period:
+    complete_targets: list[RegimeName] = []
+    for regime_name in all_active_next_period:
         target_stochastic_needs = {
             f"next_{s}"
-            for s in regime_to_v_interpolation_info[name].state_names
+            for s in regime_to_v_interpolation_info[regime_name].state_names
             if f"next_{s}" in stochastic_transition_names
         }
-        if name in transitions and target_stochastic_needs.issubset(transitions[name]):
-            complete_targets.append(name)
+        if regime_name in transitions and target_stochastic_needs.issubset(
+            transitions[regime_name]
+        ):
+            complete_targets.append(regime_name)
 
     next_V_extra_param_names: dict[str, frozenset[str]] = {}
 
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -76,9 +76,10 @@ def get_Q_and_F(
 
     # Keep only targets whose stochastic state needs are all covered by
     # `transitions`. Targets with missing stochastic transitions are dropped
-    # from the traced function; pre-solve validation in
-    # `_validate_no_reachable_incomplete_targets` raises if any such target
-    # has non-zero transition probability.
+    # from the traced function; `validate_regime_transitions_all_periods`
+    # (via `_validate_no_reachable_incomplete_targets` in
+    # `lcm.utils.error_handling`) raises pre-solve if any such target has
+    # non-zero transition probability.
     complete_targets: list[RegimeName] = []
     for regime_name in all_active_next_period:
         target_stochastic_needs = {
--- a/src/lcm/regime_building/Q_and_F.py
+++ b/src/lcm/regime_building/Q_and_F.py
@@ -1,9 +1,7 @@
-import logging
-from collections.abc import Callable, Mapping
+from collections.abc import Callable
 from types import MappingProxyType
 from typing import Any, cast
 
-import jax
 import jax.numpy as jnp
 from dags import concatenate_functions, with_signature
 from jax import Array
@@ -28,7 +26,7 @@ from lcm.utils.dispatchers import productmap
 from lcm.utils.functools import get_union_of_args
 
 
-def get_Q_and_F(  # noqa: C901, PLR0915
+def get_Q_and_F(
     *,
     flat_param_names: frozenset[str],
     age: float,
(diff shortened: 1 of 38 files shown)


hmgaudecker:
### Code review (relative to PR #315)

Found 5 issues:

1. **Diagnostic arrays have wrong shapes — not productmapped.** `_enrich_with_diagnostics` calls `compute_intermediates` with flat 1D grid arrays from `state_action_space`, not meshed via `productmap`. The resulting `U_arr`, `E_next_V`, `Q_arr` have shapes from JAX broadcasting (1D or scalar), not the full state x action Cartesian product. `_summarize_diagnostics` then maps axis indices to state names assuming axis i = state i, which is wrong. The per-state NaN breakdown (the main diagnostic value) will be incorrect or empty. (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/b08b3be1b373204eced456bf0ae224f9d7e381e9/src/lcm/utils/error_handling.py#L131-L145

2. **Diagnostics failure can swallow the original error.** If `_enrich_with_diagnostics` raises (shape mismatch, OOM, any bug in diagnostic code), the exception replaces the original `InvalidValueFunctionError`. The user loses the carefully constructed error message with NaN debugging guidance. The call in `validate_V` is not wrapped in try/except. (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/b08b3be1b373204eced456bf0ae224f9d7e381e9/src/lcm/utils/error_handling.py#L101-L113

3. **Multiple new functions missing Args:/Returns: docstring sections.** `_enrich_with_diagnostics` (8 params), `_summarize_diagnostics` (8 params), `_format_diagnostic_summary` (1 param), `_build_compute_intermediates_per_period` (9 params) all have one-liner docstrings only. AGENTS.md requires Google-style docstrings with Args/Returns. (Score: 85)

4. **GPU→CPU fallback catches wrong exception type.** Catches `jax.errors.JaxRuntimeError` but `compute_intermediates` runs eagerly (not JIT'd). Eager GPU errors raise different exception types, so the fallback never triggers. Also, the docstring says "JIT-compiles on the fly" but it doesn't — contradicts `SolveFunctions.compute_intermediates` docstring which correctly says "NOT JIT-compiled". (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/b08b3be1b373204eced456bf0ae224f9d7e381e9/src/lcm/utils/error_handling.py#L127-L145

5. **`validate_V` docstring missing `period` parameter in Args section.** (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/b08b3be1b373204eced456bf0ae224f9d7e381e9/src/lcm/utils/error_handling.py#L38-L57

src/lcm/regime_building/Q_and_F.py
```diff
@@ -282,6 +282,154 @@ def Q_and_F(
...
+    transitions: TransitionFunctionsMapping,
+    stochastic_transition_names: frozenset[str],
+    regimes_to_active_periods: MappingProxyType[RegimeName, tuple[int, ...]],
+    compute_regime_transition_probs: RegimeTransitionFunction,
+    regime_to_v_interpolation_info: MappingProxyType[RegimeName, VInterpolationInfo],
+) -> Callable:
+    """Build a closure that computes Q_and_F intermediates for diagnostics.
+
+    Same setup as `get_Q_and_F` but returns all intermediates instead of
+    just (Q, F). NOT JIT-compiled — only called in the error path when
```
mj023: Is there a reason this is not jit-compiled? No jitting means no memory optimizations, so sometimes the period will run in the compiled version, but then the intermediate targets can throw an OOM error.

src/lcm/utils/error_handling.py
```diff
@@ -35,36 +36,225 @@
...
+        "next_regime_to_V_arr": next_regime_to_V_arr,
+        **(dict(internal_params) if internal_params else {}),
+    }
+
+    try:
+        result = compute_intermediates(**call_kwargs)
+    except Exception:  # noqa: BLE001
+        cpu = jax.devices("cpu")[0]
+        call_kwargs = jax.device_put(call_kwargs, cpu)
+        result = compute_intermediates(**call_kwargs)
```
mj023: Not sure this is a good idea. This feature will be most important on more complex models, where you don't want to run all periods again, but these models will just not be solvable on the cpu and this step will run forever.

src/lcm/utils/error_handling.py
```diff
@@ -35,36 +36,225 @@
...
+    """
+    # Mesh grid arrays so broadcasting produces the full Cartesian product.
+    all_names = (*state_action_space.state_names, *state_action_space.action_names)
+    grids = {**state_action_space.states, **state_action_space.actions}
+    n_vars = len(grids)
+    meshed: dict[str, Any] = {}
+    for i, (name, arr) in enumerate(grids.items()):
+        shape = [1] * n_vars
+        shape[i] = len(arr)
+        meshed[name] = jnp.reshape(arr, shape)
```
mj023: This can and maybe should be done with `productmap` like in the actual `max_Q_over_A`-functions.

src/lcm/utils/error_handling.py
```diff
@@ -35,36 +36,225 @@
...
+            k: float(np.mean(np.asarray(v))) for k, v in regime_probs.items()
+        },
+        variable_names=all_names,
+        regime_name=regime_name,
+        age=age,
+    )
+    exc.add_note(_format_diagnostic_summary(exc.diagnostics))
+
+
+def _summarize_diagnostics(
```
mj023, line 183: This function is suddenly using numpy instead of Jax.

src/lcm/utils/error_handling.py
```diff
@@ -35,36 +36,225 @@
...
+        cpu = jax.devices("cpu")[0]
+        call_kwargs = jax.device_put(call_kwargs, cpu)
+        result = compute_intermediates(**call_kwargs)
+
+    U_arr, F_arr, E_next_V, Q_arr, regime_probs = result
+    exc.diagnostics = _summarize_diagnostics(
+        U_arr=np.asarray(U_arr),
+        F_arr=np.asarray(F_arr),
+        E_next_V=np.asarray(E_next_V),
+        Q_arr=np.asarray(Q_arr),
```
mj023: This will surely create a bunch of very large intermediate arrays, if we combine both functions and then jit the whole thing, it might be possible to avoid this.

mj023 (commented):
Great feature, will be very helpful when debugging more complex models. But I think the calculation of the intermediate values could be made more similar to the actual computation in `max_Q_over_A`.

hmgaudecker:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

src/lcm/utils/error_handling.py
hmgaudecker: Good catch, thanks!

src/lcm/regime_building/Q_and_F.py
hmgaudecker: Good point. It's JIT-compiled now.

The scalar `compute_intermediates` stays un-jitted (it's the inner closure fed to productmap), but one level up `_build_compute_intermediates_per_period` in `src/lcm/regime_building/diagnostics.py` wraps the productmapped closure with `_wrap_with_reduction` and then `jax.jit` — so the stored per-period callable is a JIT program. The full-shape U/F/E/Q arrays never materialise; XLA fuses compute + reduction.

The builder is gated by `enable_jit` for symmetry with the other per-period builders.

src/lcm/utils/error_handling.py
hmgaudecker: The CPU-fallback path is gone. The diagnostic closure is now JIT-compiled once per period config in `_build_compute_intermediates_per_period` (deduplicated across periods sharing the same active-target set, via #318's shared-JIT), and runs on whatever device the solve is already using. No separate CPU run-through exists.

If a specific period's JIT is too heavy for the target device, the diagnostic will hit the same OOM the solve would — but that's the same budget, not an extra one. Happy to add an opt-out knob if you see cases where the diagnostic blows the budget while the solve doesn't.

src/lcm/utils/error_handling.py
hmgaudecker: Done. `_productmap_over_state_action_space` in `src/lcm/regime_building/diagnostics.py` mirrors the `get_max_Q_over_a` pattern exactly — actions as the inner unbatched productmap, states as the outer batched productmap. `_build_compute_intermediates_per_period` feeds the scalar `compute_intermediates` through it before fusing with reductions and JIT-compiling.

src/lcm/utils/error_handling.py
hmgaudecker: Agreed, and fixed — see `_wrap_with_reduction` in `src/lcm/regime_building/diagnostics.py`. The productmap'd closure and the on-device reductions (overall scalar + per-dimension vectors for U_nan, E_nan, Q_nan, F_feasible, plus regime_probs) live inside a single `jax.jit` region. XLA fuses them, so the full-shape intermediates never materialise — the closure returns only scalars and small per-dim vectors.

The NaN fractions are now also feasibility-conditional (numerator masked by F, denominator = `max(sum(F), 1)`), which dropped a source of confusion about whether U/E[V] NaN in infeasible cells was a real cause or just noise.

hmgaudecker:
> Great feature, will be very helpful when debugging more complex models. But I think the calculation of the intermediate values could be made more similar to the actual computation in `max_Q_over_A`.

Addressed across the inline threads. The diagnostic path now mirrors `max_Q_over_a`'s structure: scalar closure → productmap over states/actions (same batched/unbatched split) → JIT-compiled. The only extra step is `_wrap_with_reduction`, which fuses per-dim/per-metric reductions into the same JIT region so nothing full-size escapes.

`get_compute_intermediates` and `get_Q_and_F` in `Q_and_F.py` share their target-enumeration helper `get_complete_targets` now, so the two paths can't drift out of sync.

Many thanks for the pushback! This made the feature much more useful.

src/lcm/regime_building/diagnostics.py
```diff
@@ -0,0 +1,236 @@
...
+    *,
+    func: Callable,
+    variable_names: tuple[str, ...],
+) -> Callable:
+    """Fuse a productmap'd intermediates function with on-device reductions.
+
+    The wrapped function returns a flat pytree of scalars and per-dimension
+    vectors instead of full state-action-shaped arrays. When JIT-compiled,
+    XLA can fuse the compute and reduce steps so the full-shape
+    intermediates never materialise.
```
mj023, line 141: Sadly this is not always true, sometimes JAX can't find the right optimizations and then matrializes some large intermediate arrays on the GPU. Not a problem for this PR though, this also happens in `max_Q_over_A`.

mj023 (approved):
Very good! These changes were exactly what I was thinking.

src/lcm/regime_building/diagnostics.py
hmgaudecker: Thanks, softened it! ("can often")

PR #318, opened 2026-04-10
Share JIT compilations across periods with same target config

hmgaudecker:
## Summary

Make `age` and `period` runtime arguments (via `**states_actions_params`)
instead of closure constants. Periods with the same active-target
configuration now share a single `Q_and_F` closure and a single JIT
compilation.

For a model with many periods and regimes (e.g., 45 periods × 18 regimes),
this reduces the number of JIT compilations from hundreds to tens — only
distinct target-regime configurations need separate compilations.

**Changes:**
- `get_Q_and_F` / `get_Q_and_F_terminal`: drop `age`/`period`/
  `regimes_to_active_periods` params; take `complete_targets` as a
  precomputed parameter instead.
- `get_compute_intermediates`: same refactor (takes `complete_targets`
  as a precomputed parameter), and now wraps its inner closure in
  `with_signature(...)` just like `Q_and_F` does, so the productmap
  downstream can see the full arg list instead of a bare
  `**states_actions_params`.
- `_get_complete_targets` (processing.py): enumerate target regimes
  active in the next period from `regime_to_v_interpolation_info` and
  keep those whose stochastic state needs are covered by `transitions`.
  Target regimes missing stochastic transitions (including those absent
  from `transitions`) are dropped;
  `_validate_no_reachable_incomplete_targets` (#316) raises pre-solve if
  any dropped target has non-zero probability.
- `_build_Q_and_F_per_period` and `_build_compute_intermediates_per_period`:
  group periods by `complete_targets`, build one closure per group. The
  diagnostic variant keeps the productmap + `_wrap_with_reduction` fusion
  from #317 — its output is the flat reduction dict consumed by
  `_summarize_diagnostics`.
- `solve_brute` / `simulate`: pass `period=jnp.int32(period)` and
  `age=ages.values[period]` at call time — JAX scalars, not Python ones,
  so the shared JIT function is traced once rather than per
  distinct value.
- `_enrich_with_diagnostics`: wraps its Python-scalar `age` / `period`
  inputs as `jnp.asarray(age)` / `jnp.int32(period)` for the same reason
  before calling the diagnostic closure.

## Test plan

- [x] `test_nan_diagnostics_end_to_end` — end-to-end test that solves a
      model with NaN-producing utility and asserts the exception has a
      populated `diagnostics` attribute (exercises the full
      build → productmap → reduction → summarize chain).
- [x] Full test suite passes (841/841)
- [x] `ty` clean
- [x] `prek` clean

Generated with Claude Code.


hmgaudecker:
### Code review (relative to PR #317)

Found 3 issues:

1. **`validate_V` docstring missing `period` in Args section.** The new `period: int | None = None` parameter has no corresponding docstring entry. AGENTS.md says "Docstrings must match type annotations." (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/755f1da655b3beafc80667d1c1890cdee5599307/src/lcm/utils/error_handling.py#L47-L66

2. **Python scalar `age`/`period` may defeat JIT sharing.** JAX folds Python scalars as compile-time constants during tracing. Calling a shared `jax.jit`-wrapped function with `period=3` then `period=5` (as Python ints) triggers separate recompilations — each distinct `(period, age)` pair gets its own XLA program. This undermines the deduplication that the PR is designed to achieve. Fix: pass as `jnp.int32(period)` and `jnp.asarray(age)` so JAX traces them abstractly. The benchmark speedup (1.94 min to 16.5s) likely comes from parallel compilation, not from avoiding retraces. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/755f1da655b3beafc80667d1c1890cdee5599307/src/lcm/solution/solve_brute.py#L101-L103

3. **`_partition_targets` docstring says "unreachable" but should say "assumed to have zero transition probability."** This was already fixed in Q_and_F.py (PR #316) but the `_partition_targets` docstring still uses the old inaccurate wording. (Score: 50)

https://github.com/OpenSourceEconomics/pylcm/blob/755f1da655b3beafc80667d1c1890cdee5599307/src/lcm/regime_building/processing.py#L1316-L1320

src/lcm/regime_building/processing.py
```diff
@@ -1284,35 +1284,98 @@ def _build_Q_and_F_per_period(
...
-            )
+    if regime.terminal:
+        func = get_Q_and_F_terminal(
+            flat_param_names=flat_param_names,
+            functions=functions,
+            constraints=constraints,
+        )
+        return MappingProxyType(dict.fromkeys(range(ages.n_periods), func))
+
+    assert compute_regime_transition_probs is not None  # noqa: S101
```
mj023: We can maybe remove this, as `compute_regime_transition_probs` should only be `None` if the regime is terminal, but we already return the function early in that case.

src/lcm/regime_building/processing.py
```diff
@@ -1331,25 +1394,43 @@ def _build_compute_intermediates_per_period(
...
-    for period, age in enumerate(ages.values):
-        if regime.terminal:
-            continue
-        assert compute_regime_transition_probs is not None  # noqa: S101
-        intermediates[period] = get_compute_intermediates(
-            age=age,
+    if regime.terminal:
+        return MappingProxyType({})
+
+    assert compute_regime_transition_probs is not None  # noqa: S101
```
mj023: Same as above.

mj023 (approved):
Looks solid, I think there are some old parts of #316 in here that might need to be removed or merged in some way. I am not sure anymore how everything fits together.

src/lcm/regime_building/processing.py
hmgaudecker: Was just for `ty` (get rid of the union), but I moved the case distinction to the calling code, so we don't need it any more.

hmgaudecker:
### Code review (relative to base `improve/lazy-diagnostics`)

Found 1 issue (supersedes/augments prior comment):

1. **Diagnostic path broken: `_wrap_with_reduction` removed but `_summarize_diagnostics` still expects its dict output.** `_build_compute_intermediates_per_period` now stores `jax.jit(mapped)` directly — a productmap over the scalar `get_compute_intermediates`, which returns a tuple `(U_arr, F_arr, E_next_V, Q_arr, active_regime_probs)`. `_enrich_with_diagnostics` passes that tuple straight to `_summarize_diagnostics(reductions=...)`, which indexes it as `reductions["U_nan_overall"]`, `reductions[f"{key_in}_by_{name}"]`, `reductions["regime_probs"]`. Any real NaN in V will crash the diagnostic path with `TypeError: tuple indices must be integers or slices, not str`. Tests pass only because `test_nan_diagnostics.py` mocks `compute_intermediates` to return the flat dict directly. The `_build_compute_intermediates_per_period` docstring also still claims the closure "fuses a productmap ... with on-device reductions."

https://github.com/OpenSourceEconomics/pylcm/blob/badfb675609ba6ad75057a3ea99e95da59ae4b44/src/lcm/regime_building/processing.py#L1468-L1471

https://github.com/OpenSourceEconomics/pylcm/blob/badfb675609ba6ad75057a3ea99e95da59ae4b44/src/lcm/utils/error_handling.py#L222-L233

PR #319, opened 2026-04-10
Add parallel AOT compilation for solve functions

hmgaudecker:
## Summary

AOT (Ahead-Of-Time) compilation uses `.lower()` + `.compile()` to compile
XLA executables before the backward induction loop, rather than compiling
on first call. This PR parallelises that step.

**Parallel compilation (`solve_brute.py`):**
- `_compile_all_functions`: deduplicate `max_Q_over_a` functions across
  `(regime, period)` pairs (shared-JIT from #318 means many pairs map to
  the same callable). Lower each unique function sequentially on the main
  thread (tracing is not thread-safe), then compile all lowered programs
  in parallel via `ThreadPoolExecutor`. XLA releases the GIL during
  `.compile()`, so threads achieve true parallelism.
- `_resolve_compilation_workers`: defaults to `os.cpu_count()`; rejects
  values `< 1` with a clear error rather than silently falling back.
- `_func_dedup_key`: hashable key supporting both plain callables and
  `functools.partial` wrappers (used by `fixed_params`). Keys include
  `id()` of each keyword value (not just the keyword name), so two
  partials that share the same underlying function and keyword names but
  bind different value objects get distinct keys.
- Private helpers (`_compile_and_log`, `_resolve_compilation_workers`,
  `_func_dedup_key`) are keyword-only, following the project convention.

**Consistent pytree structure:**
- Build `next_regime_to_V_arr` with all regime keys and proper shapes from
  the start. Without this, the pytree structure varied by period and
  forced JIT re-compilation.

**Failure snapshots (`model.py`, `solve_brute.py`):**
- `validate_V` receives a snapshot that includes already-solved sibling
  regimes in the current period plus the NaN-bearing V_arr itself.
- `Model.solve` and `Model.simulate` catch `InvalidValueFunctionError` and
  save the partial solution via `save_solve_snapshot` when `log_path` is
  set.

**API:**
- `max_compilation_workers` parameter on `Model.solve()` and
  `Model.simulate()`; `None` (default) uses `os.cpu_count()`.

## Test plan

- [x] `test_solve_fixed_params_aot_parity` — AOT and non-AOT results match.
- [x] `test_aot_dedup_with_fixed_params` — partials wrapping the same JIT
      function deduplicate correctly.

Generated with Claude Code.


hmgaudecker:
### Code review (relative to PR #318)

Found 2 issues:

1. **`lower_args` passes `period`/`age` as Python scalars, causing dtype mismatch with the solve loop.** PR #318 wrapped these as `jnp.int32(period)` and `jnp.asarray(age)` at the call site to avoid per-period recompilation. But the AOT lowering in `_compile_all_functions` still passes raw Python int and numpy scalar. JAX traces the Python int as `int64` (weak type), while the call site uses `int32`. This causes `TypeError: Argument 'period' compiled with int64[] and called with int32[]`. Fix: use `jnp.int32(period)` and `jnp.asarray(ages.values[period])` in `lower_args`. (Score: 85)

https://github.com/OpenSourceEconomics/pylcm/blob/f8bc75804b1bec98e8e8fdcb885400ba26c4b622/src/lcm/solution/solve_brute.py#L224-L226

2. **`next(iter(all_functions.values()))` raises `StopIteration` if no active periods.** Theoretical — model validation requires at least one terminal + one non-terminal regime, so `all_functions` is never empty in practice. (Score: 25)

https://github.com/OpenSourceEconomics/pylcm/blob/f8bc75804b1bec98e8e8fdcb885400ba26c4b622/src/lcm/solution/solve_brute.py#L190-L192

src/lcm/solution/solve_brute.py
```diff
@@ -114,3 +146,144 @@ def solve(
...
+    # Collect all (regime, period) -> function mappings.
+    all_functions: dict[tuple[RegimeName, int], Callable] = {}
+    for name, regime in internal_regimes.items():
+        for period in regime.active_periods:
+            all_functions[(name, period)] = regime.solve_functions.max_Q_over_a[period]
+
+    # If JIT is disabled, return raw functions directly.
+    sample_func = next(iter(all_functions.values()))
+    if not hasattr(sample_func, "lower"):
+        return all_functions
```
mj023, line 207: A bit strange, we should use `enable_jit` from the model. Also hides that AOT Compilation does not work with model fixed parameters, as the functions will all be `partial(` objects and therefore don't have `.lower()`.

tests/solution/test_solve_brute.py
```diff
@@ -120,9 +112,9 @@ def _Q_and_F(
 
     max_Q_over_a = get_max_Q_over_a(
         Q_and_F=_Q_and_F,
-        action_names=("consumption", "labor_supply"),
-        state_names=("lazy", "wealth"),
-        batch_sizes={"lazy": 0, "wealth": 0},
+        action_names=("consumption", "labor_supply", "lazy"),
+        state_names=("wealth",),
+        batch_sizes={"wealth": 0},
     )
```
mj023: Something strange happened with this test, `lazy` is supposed to be like a type, but at some point it got moved to the actions.

mj023 (approved):
Never would have thought that this would come up again #73. Looks good now, I made some changes because it was not working when using fixed parameters.

src/lcm/solution/solve_brute.py
hmgaudecker: Should be fixed now -- I think we need the lower-step upfront.

tests/solution/test_solve_brute.py
hmgaudecker: Thanks!

src/lcm/solution/solve_brute.py
```diff
@@ -78,7 +107,7 @@ def solve(
                 next_regime_to_V_arr=next_regime_to_V_arr,
                 **internal_params[name],
                 period=jnp.int32(period),
-                age=jnp.asarray(ages.values[period]),
+                age=ages.values[period],
```
hmgaudecker, line 110: The previous PR (#318) explicitly wrapped this as `jnp.asarray(ages.values[period])` to avoid dtype mismatches when calling AOT-compiled functions. This PR reverts that to a raw numpy scalar. The AOT-lowering path (line 90) correctly uses `ages.values[period]` as the trace input, but the call site here must match that exact dtype. A numpy scalar and a JAX array have different weak-type semantics — calling a compiled function with a mismatched type will raise `TypeError: Argument 'age' compiled with ... and called with ...`. Should restore `jnp.asarray(ages.values[period])`.

src/lcm/solution/solve_brute.py
```diff
@@ -114,3 +159,175 @@ def solve(
...
+        logger.info("  lowered in %s", format_duration(seconds=elapsed))
+
+    # Phase 2: Compile all lowered programs in parallel (XLA releases the GIL).
+    compiled: dict[Hashable, jax.stages.Compiled] = {}
+
+    def _compile_and_log(
+        func_id: Hashable,
+        low: jax.stages.Lowered,
+        label: str,
+    ) -> tuple[Hashable, jax.stages.Compiled]:
```
hmgaudecker, line 261: Project standard: all helper/utility function parameters must be keyword-only (`*,`). `_compile_and_log` takes positional args (`func_id`, `low`, `label`). Same applies to `_resolve_compilation_workers(max_compilation_workers)`. Neither is passed to `jax.vmap` or similar, so the keyword-only exemption does not apply. Please add `*,` as the first parameter.

src/lcm/solution/solve_brute.py
```diff
@@ -114,3 +159,175 @@ def solve(
...
+    Note:
+        The dedup ignores keyword argument *values*. This relies on the
+        invariant that partials sharing the same underlying function and
+        keyword names also share identical (same-object) keyword values,
+        which holds today because `_apply_fixed_params` uses the same
+        `regime_fixed` mapping for all periods of a regime.
+
+    """
+    if isinstance(func, functools.partial):
+        return (id(func.func), tuple(sorted(func.keywords)))
```
hmgaudecker: The deduplication ignores keyword argument *values* and relies on the undocumented invariant that all partials sharing the same underlying function and keyword names also share identical (same-object) keyword values. This breaks as soon as two regimes each have a `fixed_params` partial but with different regime-level param values — they'd share the same `id(func.func)` and keyword names but different values, so both would compile to the same (wrong) XLA program. The test `test_solve_fixed_params_aot_parity` only uses one regime, so it doesn't catch this. A safer key is `(id(func.func), tuple(sorted(func.keywords.items())))` (include the values, not just names). Keyword values are JAX arrays here, which aren't hashable by default — use `id(v)` for array values to preserve the same-object shortcut.

PR #320, opened 2026-04-10
WIP: Parallel lowering via subprocesses with HLO transfer

hmgaudecker:
## Summary

Exploratory work on parallelizing JAX lowering (tracing) across subprocesses.

**Approach:** Each subprocess rebuilds the Model from cloudpickled Regimes, lowers its assigned functions, serializes the StableHLO bytes + metadata, and sends them back. The main process reconstructs `MeshComputation` from HLO bytes in ~50ms (no re-tracing) and compiles.

**Verified:**
- HLO bytes + metadata roundtrip works correctly (49ms reconstruct + compile, correct results)
- `jax.jit`-wrapped functions can't be cloudpickled, but user Regimes can — so each subprocess rebuilds the Model independently
- `jax_enable_x64` must be set before unpickling JAX arrays in subprocess

**Why parked:**
- Lowering on GPU is already fast (~5-15s per function, not the 4 min seen on CPU). Sequential AOT (lower + thread-pool compile) takes ~4 min total for 37 functions — acceptable.
- This approach adds significant complexity: cloudpickle, spawn, `MeshComputation` reconstruction via JAX private APIs (`_lowering`, `_name`, `_donated_invars`, `_platforms`, `_compiler_options_kvs`, `make_ir_context`)
- JAX private APIs may break across versions
- Remaining issues: dtype mismatches with x64 mode, test failures with `io_callback` in reconstructed executables

**When to revisit:** If lowering times grow significantly (e.g., with larger models or more target regimes in the Q_and_F continuation value loop), this approach could provide ~N-fold speedup for the lowering phase.

### DIFF
--- a/docs/user_guide/installation.md
+++ b/docs/user_guide/installation.md
@@ -110,7 +110,9 @@ importing pylcm:
 ```python
 import os
 
-os.environ["JAX_COMPILATION_CACHE_DIR"] = "/scratch/$USER/.cache/jax"
+os.environ["JAX_COMPILATION_CACHE_DIR"] = os.path.expandvars(
+    "/scratch/$USER/.cache/jax"
+)
 
 import lcm
 ```
--- a/src/lcm/model.py
+++ b/src/lcm/model.py
@@ -1,7 +1,5 @@
 """Collection of classes that are used by the user to define the model and grids."""
 
-import logging
-import os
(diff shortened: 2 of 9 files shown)


hmgaudecker:
Won't need this.

PR #308, opened 2026-04-09
Auto-convert pd.Series in fixed_params

hmgaudecker:
## Summary

- `fixed_params` now auto-converts `pd.Series` to JAX arrays, matching the behavior of runtime `params`
- Removes `Model` dependency from `pandas_utils.py` and eliminates the callback pattern -- `model_processing.py` calls `convert_series_in_params` and `_validate_param_types` directly with decomposed parameters (`regimes`, `ages`, `regime_names_to_ids`)
- Previously, passing a `pd.Series` in `fixed_params` caused `TypeError: ... is not a valid JAX type` during tracing

## Test plan

- [x] New test: Series as fixed_param solves and simulates
- [x] New test: parity between Series as fixed_param vs runtime param (identical results)
- [x] New test: mixed Series + scalar fixed_params
- [x] New test: Series fixed_param indexed by derived categorical with `derived_categoricals` on `Model.__init__`
- [x] Full test suite (824 passed)

Generated with [Claude Code](https://claude.ai/code)

PR #321, opened 2026-04-10
Fix outcome axis for cross-grid stochastic transitions

hmgaudecker:
## Summary

- For per-target stochastic transitions crossing grid sizes (e.g. 3-state health → 2-state health), `_build_outcome_mapping` now uses the **target** regime's grid for the outcome axis instead of the source regime's grid
- Fixes shape mismatch (`ValueError: Axis must be specified when shapes of a and weights differ`) in `Q_and_F.py:217` during solve

## Root cause

`_build_outcome_mapping` extracted the state name from the qualified function name (`next_health__post65` → `health`) but looked it up in the source regime's grids. For cross-grid transitions, the source grid has more categories than the target, producing an array with the wrong last dimension.

## Test plan

- [x] New test `test_convert_series_cross_grid_transition` — 3-state source → 2-state target, verifies `(n_ages, 3, 2)` shape
- [x] All 824 existing tests pass (`pytest -n 7`)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -223,7 +223,7 @@ Model(
 ### Derived Categoricals
 
 When parameters are indexed by a DAG function output (not a model state/action), declare
-`derived_categoricals={"name": DiscreteGrid(...)}` on the `Regime` that uses it. For
+`derived_categoricals={"name": CategoryClass}` on the `Regime` that uses it. For
 convenience, model-level `derived_categoricals` on `Model(...)` are broadcast to all
 regimes. Functions used as derived categoricals must return **integer** types, not
 booleans — JAX cannot use booleans as array indices inside JIT. Use `jnp.int32(...)` to
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -222,10 +222,12 @@ Model(
 
 ### Derived Categoricals
 
-When `solve()` / `simulate()` parameters are indexed by a DAG function output (not a
-model state/action), pass `derived_categoricals={"name": DiscreteGrid(...)}`. Functions
-used as derived categoricals must return **integer** types, not booleans — JAX cannot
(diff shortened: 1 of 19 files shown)


hmgaudecker:
### Code review

Found 3 issues (only changes relative to PR #308):

1. **`_Rid` naming inconsistency.** New test defines `class _Rid` (lowercase d) but all other test-local regime ID classes in the file use `_RId` (capital D) — see lines 527, 1435, 1525, 1607, 1698. CLAUDE.md says "Consistent naming across a file. When multiple functions in the same file use the same concept, use the same parameter name everywhere." (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/ee49b734bc11c239b2f311a4749f6f79a97650f4/tests/test_pandas_utils.py#L447-L449

2. **`_build_outcome_mapping` docstring not updated.** The docstring lists two cases (state transitions and regime transitions) but the PR adds a third case (per-target transitions using target regime's grid). The inline comment is helpful but the function-level docstring should also describe this behavior. (Score: 50)

https://github.com/OpenSourceEconomics/pylcm/blob/ee49b734bc11c239b2f311a4749f6f79a97650f4/src/lcm/pandas_utils.py#L656-L668

3. **ty failure: `arr.shape` on `int | float | Array` union.** Test accesses `.shape` on a value typed as `int | float | Array`. Needs type narrowing (e.g., `assert isinstance(arr, Array)`) or `# ty: ignore[unresolved-attribute]`. (Score: 75)

https://github.com/OpenSourceEconomics/pylcm/blob/ee49b734bc11c239b2f311a4749f6f79a97650f4/tests/test_pandas_utils.py#L1768-L1770

mj023 (approved):
Looks correct, nothing from my side.

PR #322, opened 2026-04-12
Fix AOT deduplication for fixed_params models

hmgaudecker:
## Summary

- When `fixed_params` wraps `max_Q_over_a` functions with `functools.partial`, AOT deduplication via `id(func)` treats each partial as unique — even when they wrap the same underlying JIT function with the same kwargs
- Add `_func_dedup_key()` helper that deduplicates by `(id(func.func), sorted_kwarg_names)` for partials, `id(func)` for plain callables
- Two new tests: solve parity (V arrays identical with/without fixed_params) and dedup count (same number of unique functions)

## Test plan

- [x] `test_solve_fixed_params_aot_parity` — V arrays match
- [x] `test_aot_dedup_with_fixed_params` — dedup count matches baseline
- [x] Full test suite: 833 passed, 5 skipped, 2 xfailed
- [x] `prek` clean
- [x] `ty` clean

### DIFF
--- a/src/lcm/solution/solve_brute.py
+++ b/src/lcm/solution/solve_brute.py
@@ -1,7 +1,8 @@
+import functools
 import logging
 import os
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Hashable
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from types import MappingProxyType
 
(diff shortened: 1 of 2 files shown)


hmgaudecker:
Moving this into #319 instead

PR #324, opened 2026-04-18
Let H consume DAG function outputs

hmgaudecker:
## Summary

Let pylcm's default Bellman aggregator (and any user-supplied H) consume DAG function outputs as H inputs, not only user-supplied scalars (through `params` / `fixed_params`). This unlocks the "per-type discount factor" pattern used in aca-dev — the discount factor is a DAG function indexed by a state (e.g. `pref_type`, `discount_type`), and `_default_H(utility, E_next_V, discount_factor)` resolves it automatically.

Before this PR, a `discount_factor` regime function was silently invisible to H: `H_kwargs` only included scalar values from user params, so the solve failed with `TypeError: _default_H() missing 1 required positional argument`.

**Core change (`src/lcm/regime_building/Q_and_F.py`):**
- Extend both `get_Q_and_F` and `get_compute_intermediates` so that every `_H_accepted_params` name that is also in `regime.functions` (other than `utility`, `feasibility`, or `H`) is computed as a DAG output at call time and merged into `H_kwargs`.
- Factor the build-time DAG compilation into `_get_h_dag_fn`, which returns `None` when H does not need DAG outputs — the common case — so the scalar-H path stays fully unchanged.
- The two pools (user params / DAG outputs) are disjoint by construction: `create_regime_params_template` excludes function names from every non-H function's params dict, so no precedence rule is needed.

**Reference test (`tests/solution/test_custom_aggregator.py`):**
- Reuse `_make_model` via a new `with_pref_type` switch that adds a three-category `pref_type` state and wires `discount_factor(pref_type, discount_factor_by_type)` as a DAG function.
- `test_dag_output_feeds_default_h_monotone_in_discount_factor` asserts V is monotone in the per-type β, confirming the DAG-output path works end-to-end with the default H.

**Apply to the Mahler & Yum example:**
- `DiscountType` categorical (small/large).
- `discount_type` as a fixed state on both alive and dead regimes (`state_transitions["discount_type"] = None`).
- New DAG function `discount_factor(discount_type, discount_factor_by_type)` on `ALIVE_REGIME.functions`.
- `create_inputs` now returns `(params, initial_states)` — dropped the third tuple element; exposes the two-valued β array via `params["discount_factor"]["discount_factor_by_type"]`. Callers (benchmark, regression generator/test, docs example) simplify to a single dict and a single simulate call.
- `utility` accepts an unused `discount_type` arg to satisfy pylcm's state-usage check (same pattern as the reference test).
- Regenerated `tests/data/regression_tests/f64/mahler_yum_simulation.pkl` to reflect the new simulation output (single solve; `discount_type` column present; both types in one stream).

Also ships a minor cleanup: drops a duplicated `_build_nan_model` + `test_nan_diagnostics_end_to_end` block left over from an earlier cascade merge.

## Test plan

- [x] `tests/solution/test_custom_aggregator.py::test_dag_output_feeds_default_h_monotone_in_discount_factor` — new coverage for the DAG-output → H path.
- [x] `tests/test_regression_test.py::test_regression_mahler_yum` — passes on the regenerated fixture (GPU + f64).
- [x] Full test suite: 842 passed, 5 skipped (CPU env).
- [x] `ty` clean, `prek` clean.


hmgaudecker:
Closing to reopen as a clean single-commit stack; replaced by the new PR in the same slot. No content change.

hmgaudecker:
Reopened as #327 (clean single-commit stack).

PR #329, opened 2026-04-19
Lift discrete fixed states as partition dimensions

hmgaudecker:
## Summary

`state_transitions[name] = None` on a `DiscreteGrid` state has always meant "this value never changes." Today pylcm implements that by vmap'ing an auto-generated identity transition alongside every other state, paying memory linearly in the fixed state's cardinality. Because the Bellman backward induction never couples values across such a dimension, each fixed state partitions the state space into independent sub-problems: pylcm compiles one reduced sub-model and runs it once per point in the product of all partition grids.

This PR adds that machinery end-to-end. User-facing API is unchanged — `None` transitions still declare a fixed state; internally pylcm lifts them out of `variable_info` / state-action space / Q_and_F, iterates the partition product in `Model.solve` / `Model.simulate`, and restores partition axes at the API boundary. **Compilation happens once per `Model.solve` call; the compiled kernel is reused across all partition points.**

## What changed

- **`src/lcm/regime_building/partitions.py`** (new). Shared helpers: `detect_model_partitions`, `lift_partitions_from_regime`, `iterate_partition_points`, `inject_partition_scalars`, `stack_partition_V_arrays`, `group_subjects_by_partition`, `slice_V_at_partition_point`, `slice_initial_conditions`. All handle the empty-grid case transparently (one iteration, no stacking, no grouping).
- **`src/lcm/regime_building/processing.py`**. `process_regimes` detects model-level partitions up-front and passes reduced regimes to every downstream builder. `InternalRegime` gains a `partitions` field; downstream machinery never sees partition states as vmap'd axes.
- **`src/lcm/model.py`**. `Model._partition_grid` aggregates partitions across regimes (validates category agreement). `Model.solve` wraps backward induction in a partition loop, compiling once up-front and dispatching the compiled kernel per point; sub-V-arrays stack along trailing axes. `Model.simulate` groups subjects by their partition-value tuple, slices V-arrays per group, and concatenates `PeriodRegimeSimulationData` with globally-correct `subject_ids`.
- **`src/lcm/solution/solve_brute.py`**. `solve()` splits into `compile_solve` + `run_compiled_solve` so `Model.solve` can reuse the compiled kernel across partition points.
- **`src/lcm/simulation/{simulate,result,initial_conditions}.py`**. `subject_ids` becomes an explicit field on `PeriodRegimeSimulationData`; `simulate()` accepts it as a kwarg so per-partition dispatch groups stay aligned after concatenation. `to_dataframe` emits partition names as state columns with correct per-regime categorical mapping. Partition names are required in `initial_conditions`.

## Qualification rules for partition lifting

A discrete state qualifies as a partition iff:

1. The grid is a `DiscreteGrid`.
2. Every non-terminal regime where the state appears declares `None`.
3. The state appears in at least one non-terminal regime (target-only states populated by per-target transitions at the boundary are excluded).
4. Categories agree across regimes.

Terminal regimes (which must have empty `state_transitions` by validation) count the absence of an entry as equivalent to `None`. Continuous fixed states (`None` on a continuous grid) fall through to an identity transition — partitioning a continuous value is meaningless.

## Mahler-Yum impact

Four fixed states (`productivity`, `education`, `discount_type`, `health_type`) auto-lift on this branch, giving a 32-point partition product. One JIT compile across all partition points (shape invariant across points).

## Test plan

- [x] `pixi run --environment tests-cpu tests -n 7` — **841 pass, 5 skipped** (including 13 new partition tests).
- [x] `pixi run ty` — clean.
- [x] `pixi install -e tests-cuda12 --locked` clean (CI lockfile parity).
- [ ] Mahler-Yum GPU benchmark end-to-end on cuda13.

## Not in scope

- Removing `_IdentityTransition` entirely — continuous fixed states still need it. Narrower cleanup is a follow-up.
- Memory-optimal stacking via `jax.lax.scan` into a pre-allocated output buffer (current implementation holds all sub-V-arrays during stack). Peak could drop another ~30 %.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


hmgaudecker:
Closed in favor of #331, which lifts partition axes as a JAX-visible vmap dimension (single compile, single dispatch) instead of the per-point Python loop this PR introduced.

PR #330, opened 2026-04-19
Hoist compile_simulate out of Model.simulate partition loop

hmgaudecker:
## Summary

Mirrors the compile-once pattern #329 introduced for `Model.solve`: split `simulate()` into `compile_simulate` + `run_compiled_simulate`, and hoist the `compile_simulate` call out of `Model.simulate`'s partition loop.

**Today's win is modest** — the per-regime kernels (`argmax_and_max_Q_over_a`, `calculate_next_states`, `calculate_next_regime_membership`) already live on `InternalRegime` from model initialization, so `compile_simulate` is a reference bundle, not a fresh compile pass. What this PR buys is:

1. **API symmetry with solve.** Readers no longer have to chase two different patterns.
2. **A pure per-partition-point entry point.** `run_compiled_simulate` takes exactly the args that vary per partition (params, initial conditions, V-slice, seed, subject_ids) — and nothing else. Partition dispatch is one call into this function.
3. **A stable seam for multi-GPU.** `CompiledSimulate.devices: tuple[jax.Device, ...] | None = None` is declared now and unused. A future PR can populate it from `compile_simulate(..., devices=...)` to pin compiled kernels to specific devices, then swap `Model.simulate`'s serial `for` loop for `jax.pmap(run_compiled_simulate)` or an explicit thread-pool doing `jax.device_put` per point — without touching the inner loop body.

## What changed

- **`src/lcm/simulation/simulate.py`**: add `CompiledSimulate` frozen dataclass, `compile_simulate(...)` factory, `run_compiled_simulate(...)` (the former `simulate()` body minus the structure bundle). `simulate(...)` stays as a thin wrapper so `tests/simulation/*` and any external callers keep working.
- **`src/lcm/model.py`**: `Model.simulate` calls `compile_simulate` once above the partition loop, `run_compiled_simulate` inside. No change to stacking, merging, error handling, debug snapshotting, or partition detection.
- **`tests/regime_building/test_partitions.py`**: add `test_simulate_compiles_once_for_multi_point_partition` mirroring the existing solve equivalent; monkeypatches `compile_simulate` and asserts call_count == 1 over a 3-point partition.

## Multi-GPU forethought (explicit non-goal for this PR)

Preconditions the follow-up will lean on, documented here so they do not drift:

1. `run_compiled_simulate` is **pure** — no reads from `self`, no iteration-order-dependent side effects. A future `jax.pmap(run_compiled_simulate, in_axes=(None, 0, 0, 0, 0, 0))` wrap over partition points is a structural no-op for ragged-free groups.
2. `CompiledSimulate.devices` is the declared home for per-device kernel placement.
3. The partition loop body is already one function call per point — swap for `jax.lax.map` (serial, traced) or an explicit thread-pool that `device_put`s inputs, same signature.
4. `_merge_sub_simulation_results` uses `jnp.concatenate`, which works across devices transparently.

Not in this PR:

- Ragged subject counts per partition (pmap needs uniform leading axis; follow-up will pad groups to max size using the existing `subjects_in_regime` masking machinery in `_simulate_regime_in_period`).
- Per-device kernel placement.
- V-slice device placement.

## Test plan

- [x] `pixi run ty` — clean.
- [x] `pixi run --environment tests-cpu tests -n 7` — 865 passed, 5 skipped (one new test above the pre-PR 864 on this stack).
- [x] `pixi install -e tests-cuda12 --locked` — clean.
- [ ] GPU test run via CI.
- [ ] Mahler-Yum GPU benchmark — no regression expected (kernels unchanged).

Stacked on #329.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


hmgaudecker:
### Code review

Found 4 issues:

1. **`CompiledSimulate` is missing inline field docstrings** (AGENTS.md says "Inline field docstrings (PEP 257) for dataclass attributes. Place a `\"\"\"...\"\"\"` on the line after each field instead of listing fields in an `Attributes:` section in the class docstring.") The analogous `CompiledSolve` carries inline docstrings on both fields; `CompiledSimulate` has none on any of its six.

https://github.com/OpenSourceEconomics/pylcm/blob/9cc2c6fe59e94fbcf2516233e08383cede526c84/src/lcm/simulation/simulate.py#L46-L65

2. **`simulation_output_dtypes` typed `Mapping` on a frozen dataclass** (AGENTS.md says "Prefer immutable data structures throughout … Use `MappingProxyType` instead of `dict`.") All other mapping-valued fields on internal/immutable types in pylcm use `MappingProxyType`; storing a plain `Mapping` on a `frozen=True` dataclass is a leaky immutability guarantee.

https://github.com/OpenSourceEconomics/pylcm/blob/9cc2c6fe59e94fbcf2516233e08383cede526c84/src/lcm/simulation/simulate.py#L61-L64

3. **First `monkeypatch.setattr` in the new test is dead** — `Model.simulate` uses `from lcm.simulation.simulate import compile_simulate` (model.py:45), so the live call-site binding is `lcm.model.compile_simulate`. The first patch (`simulate_module.compile_simulate`) does not redirect the call in `Model.simulate`; only the second patch does. The test still passes because of the second patch, but the first line plus its explanatory comment are misleading.

https://github.com/OpenSourceEconomics/pylcm/blob/9cc2c6fe59e94fbcf2516233e08383cede526c84/tests/regime_building/test_partitions.py#L493-L499

4. **`CompiledSimulate.devices` is dead code today** (AGENTS.md says "Don't add features, refactor, or introduce abstractions beyond what the task requires. … Don't design for hypothetical future requirements.") The field is declared unused and has no `CompiledSolve` analogue; the PR's own body calls it "a seam for a future multi-GPU PR." The scaffolding adds surface area and an implicit API promise that a later PR would be free to revisit on its own terms.

https://github.com/OpenSourceEconomics/pylcm/blob/9cc2c6fe59e94fbcf2516233e08383cede526c84/src/lcm/simulation/simulate.py#L55-L65

hmgaudecker:
Closed in favor of #331, which collapses the partition sweep (both solve and simulate) into a single vmap'd dispatch inside the compiled kernel.

PR #331, opened 2026-04-20
Lift discrete fixed states as a partition axis vmap'd inside the kernel

hmgaudecker:
## Summary

- Declaring a discrete state with `state_transitions[name] = None` lifts it out of the regime's state-action space and treats it as a **partition dimension** — swept by a single `jax.vmap` over the leading axis of the params pytree.
- `Model.solve` now stacks partition scalars (`stack_partition_scalars`), compiles once, dispatches once, and reshapes the flat leading axis into `partition_shape + V.shape[1:]`. `Model.simulate` gets the same compile-once split; subject-group raggedness still iterates in Python.
- `productmap` enforces a canonical axis order (partitions outermost, batched axes next, vmap'd axes innermost) so future `shard_map` / `pmap` wiring is a one-line swap at the `compile_solve` wrap site.
- V-arrays carry partition axes as the **leading** axes (flipped from the trailing convention in the old post-dispatch helper). `_reshape_leading_partition_axis` and `slice_V_at_partition_point` own this convention.

## What's in, what's not

- In: partition auto-detection, compile/run split for solve and simulate, canonical productmap ordering, per-partition-point diagnostics, row-major reshape invariant assertion.
- Not in: `shard_map` wiring for multi-GPU (the structural preconditions are now true — next PR wires it), simulate-side vmap over subject groups.

## Replaces

- #329 (partition lifting, superseded)
- #330 (compile_simulate hoist, superseded)

## Test plan

- [x] `pixi run ty` — clean
- [x] `pixi run --environment tests-cpu tests -n 7` — 869 passed, 5 skipped
- [x] `pixi install -e tests-cuda12 --locked` — clean
- [x] GPU Mahler-Yum benchmark on runner — target `≤ 10 s` execute, `≤ 300 M` GPU peak (matches Max Jahn's `batch_size=1` reference of 8.35 s / 276 M)


mj023 (commented):
I think the dispatch is mostly a doubling of the functionality of the `batch_size` parameter, especially when using `vmap` on the partitions. There are some instances, where I can imagine this might save a little bit of memory, but overall it adds a quite complex new mechanic for what I think might be very little benefit.

The multi-device support will also be much easier to add, by distributing one of the grids over the devices and then have JAX figure out how the computation should be done.

Nonetheless this PR showed that there are some issues with the current implementation. The `batch_size` parameter does not work correctly when the grids are not ordered such that batched grids come first. This has to be fixed by making `productmap` reorder the grids according to batch size. Furthermore the solutions from previous periods occupy GPU memory, even when they are not needed anymore. These should be moved to the CPU between periods.

hmgaudecker:
I was thinking the long-run benefit would be that we wouldn't need inter-gpu communication when lifting as partitions. But maybe that is not quite worth the hassle here. What do you think?

mj023:
I am hoping that the compiler would be able to recognize that we do not need communication between GPU's, as the result for each type will already reside on that GPU. Also the communication might not actually matter, as the bandwidth between GPU's is so large and computation and communcation can overlap.

But it might not be so smart to move everything to the CPU when using multi-GPU, maybe the last period should just remain on the GPU.

hmgaudecker:
> I am hoping that the compiler would be able to recognize that we do not need communication between GPU's, as the result for each type will already reside on that GPU. Also the communication might not actually matter, as the bandwidth between GPU's is so large and computation and communcation can overlap.

We would still need to split the simulations then, right? Otherwise, we would need results from all types at that point?

> But it might not be so smart to move everything to the CPU when using multi-GPU, maybe the last period should just remain on the GPU.

Yeah, let's maybe postpone that until we have serious experience with large models. My hunch is that batching should help enough to keep the memory footprint low enough.

mj023:
> > I am hoping that the compiler would be able to recognize that we do not need communication between GPU's, as the result for each type will already reside on that GPU. Also the communication might not actually matter, as the bandwidth between GPU's is so large and computation and communcation can overlap.
> 
> We would still need to split the simulations then, right? Otherwise, we would need results from all types at that point?
> 

I would generally split the simulation differently, just batch over the subjects, as there is no interaction between them. But again, I don't think it will matter, its just not a lot of data thats being transferred.

hmgaudecker:
Yeah, I guess those are the long term burns from HPC clusters with poor bandwidth between nodes from 15 years ago...

So you are saying memory usage of the `V` arrays during simulations will never be a binding constraint? Might well be, but maybe we should make a back of the envelope calculation for that before I shut up for good 😊

hmgaudecker:
Back-of-the-envelope on V memory during simulate, taking aca-baseline at the **production** `GRID_CONFIG` (not the tiny benchmark one):

| Quantity | Value |
|---|---|
| state grid in V (consumption is an action, eliminated by max) | assets 24 × aime 12 × wage_res 5 × hcc_persistent 3 × hcc_transitory 5 = **21,600 cells/regime** |
| regimes | 18 |
| periods (51→95) | 45 |
| dtype | f32 (4 B) |

- per-period V over all 18 regimes: **1.5 MiB**
- full 45-period backward-induction stack: **70 MiB**
- forward simulate only needs V_t at period t → working set ≤ 1 period ≈ 1.5 MiB

Scaling factors before V even rounds to 10 % of an A100-40GB (f32, full stack as conservative bound):

| factor on V | full-stack V | fraction of 40 GB |
|---|---|---|
| baseline | 70 MiB | 0.18 % |
| × 4 type partition | 280 MiB | 0.7 % |
| × 4 types × all 5 grids 2× (= 32×) | 9 GiB | 22 % |
| × 4 types × all 5 grids 3× (= 243×) | 68 GiB | overflow |

So you'd need to either (a) blow up at least three continuous grids by 2-3× **and** add a 4-way type axis, or (b) push state-space dimensionality up by a whole continuous variable, before V storage on a single A100/H100 even reaches 10 % of HBM.

PCIe side, for the \"do we ship V to CPU during simulate?\" sub-question: 70 MiB over PCIe 4.0 x16 (~25 GiB/s effective) is **~3 ms** one-way. Negligible compared to the 200 ms-class precautionary-savings simulate, and the simulate's other tensors (per-subject tape, action grid for vmap, intermediate Q) saturate HBM long before V does.

Bottom line: GPU-resident V is the right default; V-array memory is not the binding constraint on simulation.

mj023:
I think in theory there are the discrete states missing, so we have:

assets 24 × aime 12 × wage_res 5 × hcc_persistent 3 × hcc_transitory 5 x 2 LagLabor x 3 SpousalInc x  3 HealthwithD x  3 PrefType x 2 ClaimedSS = 0.0093312 GB/Regime

=> x 18 Regimes x 45 Periods = 7.5 GB overall

But for example the GPU's in one Marvin Node are all connected via NVLink which has maybe like 25GB/s to each other GPU (not very clear what kind of setup they have) and also 25GB/s to other nodes. So it would still take under 0.5s to distribute the whole VF Array to all GPU's. Small compared to the 45sec Runtime and the example overestimates the array size most likely.

hmgaudecker:
Thanks! Case closed.

PR #332, opened 2026-04-21
Let H consume DAG function outputs, but don't exploit in benchmarks

hmgaudecker:
## Summary

- Extend pylcm so the Bellman aggregator `H` can consume outputs of regime
  DAG functions (new helper `src/lcm/regime_building/h_dag.py`; `Q_and_F.py`
  and `model_processing.py` updated; covered by
  `tests/solution/test_custom_aggregator.py`).
- Leaves the mahler/yum example untouched relative to `main` — the loop
  over discount-factor types stays outside pylcm, so this PR is scoped to
  the core feature only.
- Pixi lock refreshed.

Opening as a draft to trigger the benchmark workflow.

## Test plan

- [x] `pixi run -e tests-cpu pytest -n 7` (full suite)
- [x] `pixi run -e tests-cpu pytest tests/solution/test_custom_aggregator.py tests/test_regression_test.py::test_regression_mahler_yum -n 7`
- [ ] Benchmarks (via CI on this draft PR)


hmgaudecker:
Closing in favour of #333 for my own stupidity.

PR #328, opened 2026-04-19
Add asv benchmark for aca baseline model

hmgaudecker:
## Summary

- **`benchmarks/bench_aca_baseline.py`**: asv bench timing the aca baseline model with `BENCHMARK_GRID_CONFIG` (tiny grids, full 18-regime structure). The expensive parts of aca-baseline's cost structure (compile pipeline over 19 regimes, DAG resolution, pref_type batching) are preserved while per-call numerical work is shrunk to fit in an asv invocation.
- New **`benchmarks`** pixi feature pulls `aca-model` from its public git URL, pinned to a specific commit so updates are deliberate. Composed with `cuda12` into the **`benchmarks-cuda12`** env.
- **CI workflows** (`.github/workflows/benchmark-main.yml`, `.github/workflows/benchmark-pr.yml`) switch from `tests-cuda12` to `benchmarks-cuda12` so the workflows have `aca-model` available.

## Test plan

- [x] `pixi run -e benchmarks-cuda12 asv run --quick --bench AcaBaseline` on a GPU host
- [x] `benchmark-pr` workflow runs on this PR (self-hosted GPU runner)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #334, opened 2026-04-24
solve_brute: defer NaN/stats diagnostics to post-solve (async)

hmgaudecker:
## Summary

- Replace the per-(regime, period) synchronous diagnostics trio — `log_nan_in_V` + `log_V_stats` + `validate_V` — inside `solve_brute.solve`'s hot loop with an accumulator pattern that launches only device-side reductions and flushes exactly once to host after the full backward induction.
- Result: **zero** device→host syncs inside the loop (down from ~n_regimes × n_periods per solve), two host transfers post-loop (`jnp.any(isnan)` + `jnp.any(isinf)`, plus min/max/mean if debug-logging).

## Why

`log_nan_in_V` fires on every `solve()`, every period, every regime — no log-level gate. Its `if jnp.any(jnp.isnan(V_arr)) or jnp.any(jnp.isinf(V_arr))` body forces Python to materialise the JAX scalar as a bool, which means the reduction kernel has to run *and* its result has to round-trip to host before the loop can proceed. On aca-baseline-sized models that's a few hundred stalls per solve; in an MSM loop that re-solves on every optimiser iteration the overhead compounds.

The intermediate `isnan(V_arr)` bool buffer also sits at full-V shape during the reduction — the allocator has to find room for it on top of `V_arr` itself. On tight-memory machines that's been the difference between a successful solve and an OOM.

## Semantics (preserved)

- **NaN** → raises `InvalidValueFunctionError` via `validate_V`, same as before. `validate_V` still runs `compute_intermediates` on the offending slice to produce the detailed diagnostic note.
- **Inf** → `logger.warning(...)` per offending (regime, period), matching the old `log_nan_in_V` warning for the inf-without-nan case.
- **Debug stats** → emitted as before; only the call site moves from per-period to post-solve.

## Scope

- Orthogonal to the `DispatchStrategy` / partition-lifting work on PR #331. This extracts just the async-diagnostics pattern (same pair of commits the harmonisation of #331 had isolated already: `9301e29` + the `_raise_at` tweak in `4f537fd`), rebased onto #328 without the partition plumbing.
- `log_V_stats` is removed from `src/lcm/utils/logging.py` — its only caller was this `solve_brute.solve`. `log_nan_in_V` is kept because `simulate.py` still uses it (converting simulate to the same pattern is a future change).

## Test plan

- [x] `pixi run ty` clean.
- [x] `pixi run -e tests-cpu tests` — 848 pass, 5 skipped.
- [x] `prek run --all-files` clean.
- [x] Benchmark-pr workflow on PR CI — expect the Mahler-Yum / Precautionary runtime numbers within noise of #328's baseline; aca-baseline should benefit most.


hmgaudecker (commented):
Autoreview.

mj023 (approved):
Looks good. Maybe we should also consider an option to just disable all diagnostics?

hmgaudecker:
> Looks good. Maybe we should also consider an option to just disable all diagnostics?

Good catch! I thought we did that -- now it is happening:

- `log_level="off"` now skips every per-row reduction, the row-metadata accumulation, and the flush + NaN fail-fast entirely. 
- `"warning"` / `"progress"` keeps NaN/Inf detection;
- `"debug"` adds min/max/mean.

PR #335, opened 2026-04-24
Reorder dimensions in productmap

mj023:
Reorder the dimensions in `productmap` so that the batched dimensions are always applied last. Otherwise the batching will have no effect on the peak memory usage.


mj023:
@hmgaudecker I'm having a lot of trouble applying `prouctmap` to functions with `positional-only` arguments. I don't see any reason to ever use `positional-only` arguments in a function that is supposed to be productmapped, therefore I would be in favor of just disallowing this case. The reason for this was mostly that `vmap` needs positional arguments, but `map` can now handle everything. What do you think?

hmgaudecker:
Wherever we can get rid of positional arguments, we should!!!

hmgaudecker:
Btw, just thinking of this now: In #331, I tried to make the axes ordering fixed, taking into account anything like batching. I think that would probably be something worth porting here.

mj023:
> Btw, just thinking of this now: In #331, I tried to make the axes ordering fixed, taking into account anything like batching. I think that would probably be something worth porting here.

You mean fixed as in that the order of the axes in the state space of a regime does not depend on how the user orders the states in the dict that he provides?

hmgaudecker:
Precisely. User passes a dict (officially unordered, anyhow) as part of the `Regime`, which is what it is. We fix it in `InternalRegime` to whatever is most useful from a computational perspective and that is the single source of truth for everything.

mj023:
You are right, this is the better way. I had now encountered the problem that AOT Compilation did not know about the reordering. Just changing it globally is much easier.

mj023:
For some reason it is important that we first have the discrete states and then the continuous states, otherwise I see a ~10% increase in execution time for the MY model.

I think the JAX update might have also made this change unnecessary, because the memory reduction is now also in PR #334, even though the batched axis is not in front. It can probably reorder the axes itself now. Nonetheless I think it's better to fix it this way, not sure what the impact on other models is.

src/lcm/regime_building/variable_info.py
```diff
@@ -34,15 +35,29 @@ def get_variable_info(regime: Regime) -> pd.DataFrame:
...
+            regime.states[x].batch_size if regime.states[x].batch_size != 0 else jnp.inf
+        ),
+    )
+    action_order = info.query("is_action").index.tolist()
+
+    if set(state_order_disc + state_order_cont + action_order) != set(info.index):
         raise ValueError("Order and index do not match.")
 
-    return info.loc[order]
+    return info.loc[state_order_disc + state_order_cont + action_order]
```
hmgaudecker:
```suggestion
    return info.loc[[*state_order_disc, *state_order_cont, *action_order]]
```
Slight preference for this form.

src/lcm/utils/dispatchers.py
```diff
@@ -234,6 +235,13 @@ def _base_productmap_batched(
 
     """
     parameters = inspect.signature(func).parameters
+    for name, param in parameters.items():
+        if param.kind == inspect.Parameter.POSITIONAL_ONLY:
+            raise FunctionDispatchError(
+                f"Positional only prameters are not allowed in dispatched functions."
+                f"The parmeter '{name}' to the function {func.__name__} "
```
hmgaudecker:
```suggestion
                f"The parameter '{name}' to the function {func.__name__} "
```

src/lcm_examples/precautionary_savings.py
hmgaudecker, line 1: We can just revert this, no?

tests/test_dispatchers.py
```diff
@@ -12,30 +12,14 @@
 from lcm.utils.functools import allow_args
 
 
-def f(a, /, *, b, c):
-    """Tests that dispatchers can handle positional-only and keyword-only arguments.
+def f(a, *, b, c):
+    """Tests that dispatchers can handle keyword-only arguments.
```
hmgaudecker:
```suggestion
    """Tests that dispatchers can handle standard arguments and keyword-only arguments.
```

tests/test_dispatchers.py
```diff
@@ -12,30 +12,14 @@
...
 
     a is positional-only, b and c are keyword-only
```
hmgaudecker:
```suggestion
    a is positional-or-keyword, b and c are keyword-only
```

tests/test_dispatchers.py
hmgaudecker, line 1: Should add a test that positional-only is refused with a meaningful message.

hmgaudecker:
### Code review

Reporting all findings; first 4 are new, the rest are concrete suggestions for the existing inline remarks.

**New issues**

1. Second typo plus missing space in the new dispatch error. In addition to `parmeter` (already flagged), `prameters` on the previous line is also misspelled, and the two adjacent f-strings concatenate without whitespace, producing `"...dispatched functions.The parmeter..."`.

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/src/lcm/utils/dispatchers.py#L238-L245

Suggested replacement (covers both typos plus the spacing):
```python
raise FunctionDispatchError(
    "Positional-only parameters are not allowed in dispatched functions. "
    f"The parameter '{name}' to the function {func.__name__} "
    "is POSITIONAL_ONLY."
)
```

2. `_base_productmap_batched` docstring still says `Cannot have keyword-only arguments`, but this PR inverts the restriction (keyword-only is fine; positional-only is now forbidden).

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/src/lcm/utils/dispatchers.py#L227-L230

Suggested:
```
func: The function to be dispatched. Cannot have positional-only parameters.
```

3. `jnp.inf` is used purely as a Python `sorted()` key, and `import jax.numpy as jnp` was added solely for that. `math.inf` (or `float("inf")`) keeps the module JAX-free for what is plain Python data manipulation.

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/src/lcm/regime_building/variable_info.py#L40-L55

4. Orphan fixtures `setup_productmap_g` / `expected_productmap_g`. The `g` function and tests using these fixtures were removed, but the fixtures remain. `expected_productmap_g` was even rewritten to call `allow_args(f)(*helper)` with a 4-element `helper` and `.reshape(10, 7, 5, 4)` — `f` only has 3 parameters, so this would raise `TypeError` if anything ever referenced it. Delete both fixtures.

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/tests/test_dispatchers.py#L43-L66

**Suggestions for the existing inline remarks**

a. `[*state_order_disc, *state_order_cont, *action_order]` form — agreed; it also makes the matching `set(...)` check on the line above more consistent if you want to apply it there too.

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/src/lcm/regime_building/variable_info.py#L56-L61

b. `dispatchers.py` typo — see issue #1 above for the full corrected snippet covering both `prameters` and `parmeter`.

c. `precautionary_savings.py` "We can just revert this, no?" — yes. With `get_variable_info` now imposing the canonical order globally, source-order in the user-facing dict no longer affects the computation, so reverting the swap keeps the example readable.

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/src/lcm_examples/precautionary_savings.py#L127-L135

d, e. `tests/test_dispatchers.py` docstring — apply both your inline edits in one shot:

https://github.com/OpenSourceEconomics/pylcm/blob/501387bd2aef0f853620dcbf247f9ff6be0ba763/tests/test_dispatchers.py#L14-L21

```python
def f(a, *, b, c):
    """Tests that dispatchers can handle standard arguments and keyword-only arguments.

    a is positional-or-keyword, b and c are keyword-only
    """
    return jnp.sin(a) + jnp.cos(b) + jnp.tan(c)
```

f. Test for the new `FunctionDispatchError` path — minimal version, drop it next to `test_productmap_with_all_arguments_mapped`:

```python
from lcm.exceptions import FunctionDispatchError


def test_productmap_rejects_positional_only():
    def h(a, /, *, b):
        return a + b

    with pytest.raises(FunctionDispatchError, match="POSITIONAL_ONLY"):
        productmap(func=h, variables=("a", "b"), batch_sizes={"a": 0, "b": 0})
```

While you are in this file, also clean up the orphan fixtures from issue #4.

hmgaudecker (approved):
Very nice -- just made a few cosmetic changes, nothing substantial!

src/lcm_examples/precautionary_savings.py
hmgaudecker: @mj023, this is the only thing that is open -- just in case there  is too much info on the page :upside_down_face:

PR #337, opened 2026-04-28
Apply RegimeName + introduce 6 sibling typing aliases (closes #323)

### LINKED ISSUE #323: Apply RegimeName (and consider StateName/ActionName/...) consistently across the codebase
## Motivation

`src/lcm/typing.py` defines `type RegimeName = str` and uses it in several
interface mappings (e.g. `internal_regimes: MappingProxyType[RegimeName,
InternalRegime]`). The alias is a pure semantic label — it is erased at
runtime — but it makes signatures self-documenting and prevents passing
e.g. a state name where a regime name is expected.

Today the alias is applied inconsistently: some modules (e.g.
`interfaces.py`, `utils/error_handling.py`) use it faithfully, while
others (especially under `simulation/`, `solution/`, `params/`,
`pandas_utils.py`) still type regime-name parameters, keys and return
values as bare `str`. #316 tightened a handful of them in
`regime_building/processing.py`; the rest of the codebase needs the same
treatment.

At the same time, we should decide whether to introduce matching aliases
for other semantic string categories that currently float around as
`str`.

### Part 1 — apply `RegimeName` everywhere it is meant

Audit every `str` annotation in the codebase and replace with
`RegimeName` when the value is (or must be) a regime name. Typical
signs:
- Variable/parameter names like `regime_name`, `regime_names`, `name`
  while iterating `regimes`.
- Keys of mappings typed against `internal_regimes`, `regimes_to_ids`,
  `regime_to_V_arr`, `state_action_spaces`, etc.
- Return values of helpers that produce regime names (sets, lists,
  tuples).

Files likely to need changes (non-exhaustive):
- `src/lcm/simulation/simulate.py`, `initial_conditions.py`, `result.py`
- `src/lcm/solution/solve_brute.py`
- `src/lcm/pandas_utils.py`
- `src/lcm/model.py`, `src/lcm/model_processing.py`
- `src/lcm/params/**`
- tests under `tests/`

### Part 2 — consider new aliases for sibling concepts

Evaluate whether to add similar semantic aliases. Candidates and
their current callsites:

- `StateName` — used in `variable_info.query("is_state")`, `states_per_regime`
  values, `state_names` tuples on `StateActionSpace`, `V` interpolation
  info, etc.
- `ActionName` — dual to `StateName`, appears on `StateActionSpace`,
  `action_names`, per-regime `actions` mappings.
- `StateOrActionName` — used in flat-grid mappings (`flat_grids`,
  `all_grids[regime]` values), `state_and_discrete_action_names`.
- `ShockName` — state names that are shock grids; subset of `StateName`
  so may not pay for itself.
- `FunctionName` — names of user functions (`"utility"`, `"H"`, custom
  helpers). Maybe overkill.
- `TransitionFunctionName` — `next_<state>`, `weight_next_<state>`, etc.
  Already partially structured via `next_`-prefix conventions.

For each: decide if the alias improves readability enough to justify
the churn. A single `type StateOrActionName = str` + callsite renames
may buy most of the value with little cost. Document the decision in
`.ai-instructions/` so future additions stay consistent.

## Non-goals

- Switching any of these to `NewType` (which is runtime-distinct and
  would require explicit casts at every boundary). The existing
  `type RegimeName = str` is a pure alias; we stick with that model.
- Introducing structural/value validation (that already lives in
  `_validate_*` helpers).

## Acceptance criteria

- `pixi run ty` still passes.
- Every `regime_name: str` / `RegimeName`-keyed mapping typed `dict[str,
  ...]` / `Mapping[str, ...]` in the source tree is reviewed and either
  tightened or annotated with a comment explaining why `str` is correct.
- A short note in `.ai-instructions/` (or `CLAUDE.md` via `AGENTS.md`)
  listing the accepted aliases and when to use them.

## Related

- #316 — introduced `_get_reachable_targets` return type `set[RegimeName]`
  and tightened `processing.py` regime-name annotations; this issue
  generalises that work.

## Known instances (starter list)

From #316-cascade exploration, these bare-`str` regime-name usages in the
source tree should be tightened to `RegimeName` and/or renamed from
short-form loop variables (`name`, `n`, `r`, `rn`) to `regime_name`:

**`src/lcm/regime_building/processing.py` (on `main`):**
- L89: `name: set(regime.states.keys()) for name, regime in regimes.items()`
- L93: `for name, regime in regimes.items()`
- L101: `{n: get_variable_info(r) for n, r in regimes.items()}`
- L103: `all_grids = MappingProxyType({n: get_grids(r) for n, r in regimes.items()})`
- L108: `{n: create_v_interpolation_info(r) for n, r in regimes.items()}`
- L115: `for n in regimes`
- L119: `{n: ages.get_periods_where(r.active) for n, r in regimes.items()}`
- L123: `for name, regime in regimes.items()`

**`src/lcm/solution/solve_brute.py` (on `main`):**
- L41: `{name: jnp.empty(0) for name in internal_regimes}`
- L58: `for name, internal_regime in active_regimes.items()`

**`src/lcm/simulation/transitions.py` (on `main`):**
- L229: `regime_ids = jnp.array([regime_names_to_ids[name] for name in regime_names])`

Beyond these already-identified locations, also audit the wider files
listed under "Scope — Part 1" above.

hmgaudecker:
## Summary
- Catalog the seven accepted domain string aliases (`RegimeName`, `StateName`, `ActionName`, `StateOrActionName`, `ShockName`, `FunctionName`, `TransitionFunctionName`) in `AGENTS.md` (commit 1).
- Apply `RegimeName` across every signature in the source tree and rename short loop variables (`n`, `r`, `name`) to `regime_name` (commit 2 — the bulk-of-the-codebase commit).
- Introduce each remaining alias one commit at a time (commits 3–8), tightening only the slots where the meaning is unambiguous.

All eight commits are pure aliasing — `type X = str` is erased by ty, so behaviour is unchanged. Each commit is self-consistent (`pixi run ty` clean, full pytest suite green, prek green).

Branched off `reorder-batched-dims` (#335), since this builds on top of those renames.

Closes #323.

## Test plan
- [x] CI green on Linux + macOS
- [x] `pixi run ty` clean (verified locally for every commit)
- [x] `pixi run -e tests-cpu pytest -n 7` clean (848 passed, 5 skipped — verified locally for every commit)
- [x] `prek run --all-files` clean (verified locally for every commit)


src/lcm/regime_building/diagnostics.py
```diff
@@ -38,11 +42,11 @@ def _build_compute_intermediates_per_period(
     functions: FunctionsMapping,
     constraints: FunctionsMapping,
     transitions: TransitionFunctionsMapping,
-    stochastic_transition_names: frozenset[str],
+    stochastic_transition_names: frozenset[TransitionFunctionName],
     compute_regime_transition_probs: RegimeTransitionFunction,
     regime_to_v_interpolation_info: MappingProxyType[RegimeName, VInterpolationInfo],
     state_action_space: StateActionSpace,
-    grids: MappingProxyType[str, Grid],
+    grids: MappingProxyType[StateOrActionName, Grid],
```
mj023, line 49: Do we need `StateOrActionName`? Can we not just do `StateName | ActionName`, or does this give a problem with ty?

mj023 (approved):
The places where the new typings have been added all look correct to me.

src/lcm/regime_building/diagnostics.py
hmgaudecker: No, not strictly required -- just readability since we need it so often.

hmgaudecker:
### Code review

Found 1 issue:

1. `_validate_per_target_dict` updated `state_name: str` → `state_name: StateName` in this PR but left `targets: Mapping[str, object]` even though its keys are target regime names. The matching slot in `_add_raw_transition` and `collect_state_transitions` was tightened to `RegimeName` in commit `1fd8026`, so this is the same scope. AGENTS.md says "`RegimeName` — keys of `regimes`, `internal_regimes`, ..." applies here too.

https://github.com/OpenSourceEconomics/pylcm/blob/c46fd6068fd7358cd877475c2a89bb5aa52e7c78/src/lcm/regime_building/validation.py#L257-L262

hmgaudecker:
### Code review (additional issues, below the 80 confidence threshold)

5 more issues found across the parallel review agents that the threshold filter dropped:

1. `Regime.state_transitions` per-target dict value type still has bare `str` for the inner regime-name keys (parallel slot in `_add_raw_transition` was tightened to `RegimeName` in `1fd8026`).

   https://github.com/OpenSourceEconomics/pylcm/blob/c46fd6068fd7358cd877475c2a89bb5aa52e7c78/src/lcm/regime.py#L139-L143

2. `_extend_transitions_for_simulation` return type is `dict[str, Callable[..., Array]]`; keys are transition function names. The function's `stochastic_transition_names` parameter was tightened to `frozenset[TransitionFunctionName]` in this PR, so the return type is the matching miss.

   https://github.com/OpenSourceEconomics/pylcm/blob/c46fd6068fd7358cd877475c2a89bb5aa52e7c78/src/lcm/regime_building/next_state.py#L143-L148

3. `get_grids` docstring says "Create a **dictionary** of grid objects ..." / "Immutable **dictionary** containing all variables ..." while the annotation is `MappingProxyType[StateOrActionName, Grid]`. AGENTS.md: "`MappingProxyType[...]` → 'Immutable mapping of ...'. Never write 'Dict' when the annotation is `Mapping` or `MappingProxyType`."

   https://github.com/OpenSourceEconomics/pylcm/blob/c46fd6068fd7358cd877475c2a89bb5aa52e7c78/src/lcm/regime_building/variable_info.py#L66-L80

4. `build_regime_transition_probs_functions` has `grids: MappingProxyType[StateOrActionName, Grid]` but the docstring still says "grids: Immutable mapping of grid names to grid objects." The whole point of `StateOrActionName` is to make this slot self-documenting; "grid names" is now stale.

   https://github.com/OpenSourceEconomics/pylcm/blob/c46fd6068fd7358cd877475c2a89bb5aa52e7c78/src/lcm/regime_building/processing.py#L1175-L1192

5. (Disputable, score 25) Four single-or-positional-arg helpers touched by this PR were not made keyword-only — `_classify_transitions`, `_wrap_transitions`, `_add_raw_transition`, `_validate_per_target_dict`. The `b02752a` commit covered helpers with leading positional-or-keyword args before a `*,` separator, so these arguably aren't in scope; flagging only because the PR's "make helpers fully keyword-only" intent was a judgment call worth confirming.

PR #338, opened 2026-04-29
Runtime-supplied points on action grids; tighten grid + regime validators

hmgaudecker:
## Summary

Three threads bundled into one PR:

1. **Feature.** `IrregSpacedGrid` runtime-supplied points (previously state-only) extended to continuous **action** grids. Action grids declared as `IrregSpacedGrid(n_points=N)` now contribute `{action_name: {"points": "Float1D"}}` to the regime params template; `InternalRegime.state_action_space(regime_params=...)` substitutes runtime points into `continuous_actions` at solve **and** simulate time. Motivation: lets a model tie its `consumption` (or other continuous-action) grid bounds to a per-iteration parameter without rebuilding the model each estimation step.

2. **Two silent-failure fixes that surfaced while completing the feature.**
   - **Simulate was using the NaN placeholder for runtime action grids.** `create_regime_state_action_space` (forward simulation) bypassed the substitution that solve uses, so `state_action_space.continuous_actions` carried the compile-time NaN placeholder. That fed into `argmax_and_max_Q_over_a` and `_lookup_values_from_indices`, optimal actions came back NaN, the source regime's `next_state` propagated NaN to every target regime's namespaced state, and `validate_V` raised on the first downstream regime whose utility depended on those states. Now routes through the same substitution path solve uses; restores `_validate_all_states_present` on the per-subject states overlay.
   - **`IrregSpacedGrid.to_jax()` silently returned a placeholder.** Previously runtime grids returned `jnp.full(N, NaN)` from `to_jax()`, so any caller forgetting to call `state_action_space(regime_params=...)` got an undetectably broken grid. Now raises with a message pointing the caller at the substituted-grid path.

3. **Adjacent validator hardening (also surfaced while completing the feature).**
   - **Function-output / discrete-grid-indexed-input clash detector.** A regime function `f` that takes a discrete grid `g` (state, action, or derived categorical) as input has a per-cell scalar output, and a consumer that does `f[g]` raises `IndexError: Too many indices: array is 0-dimensional, but 1 were indexed` at trace time — well after the user could see what they wrote. New AST validator catches the pattern at `Regime.__post_init__`. The clashing function in the original spec produced this footgun and survived all existing validators.
   - The validator only fires when the producing function actually takes the discrete grid as input — verified by exercising both shapes with the validator disabled (`func_output[grid]` is correct code when the producer doesn't take `grid`; the validator now correctly exempts that shape).
   - **Log-spaced grids rejected non-positive start.** `LogSpacedGrid(start=0, …)` (or negative) silently produced NaN/-inf via `to_jax()` because the base validator only checked `start < stop`. Now refused at construction. Tightened two related silent-failure modes for free: `_validate_continuous_grid` rejects non-finite `start`/`stop` (the `start >= stop` check is False for NaN, so a NaN bound previously slipped through), and `_validate_irreg_spaced_grid` rejects non-finite points (the ascending-order test uses `>=`, also False for NaN, so an all-NaN runtime-supplied points array previously passed the order check silently).

## Changes

- `src/lcm/grids/continuous.py` — `LogSpacedGrid.__post_init__` enforces `start > 0`; `_validate_continuous_grid` rejects non-finite bounds via `jnp.isfinite`; `_validate_irreg_spaced_grid` rejects non-finite points; `IrregSpacedGrid.to_jax()` raises with a message pointing at `state_action_space(regime_params=...).states[name]` / `.continuous_actions[name]`.
- `src/lcm/params/regime_template.py` — walk `regime.actions` alongside `regime.states` for runtime-grid params; shared `_fail_if_runtime_grid_shadows_function` helper.
- `src/lcm/interfaces.py` — `InternalRegime.state_action_space(regime_params=...)` builds both state and continuous-action replacements in one pass and returns via `_base_state_action_space.replace(...)`.
- `src/lcm/state_action_space.py` — `_grid_to_jax_or_placeholder` private helper centralises the NaN-placeholder convention; deep-module ordering (public function on top, helpers below).
- `src/lcm/regime_building/validation.py` — new `_validate_function_output_grid_indexing` AST validator covering states, actions, and derived categoricals; only fires when the producing function takes the discrete grid as input.
- `src/lcm/simulation/transitions.py` + `src/lcm/simulation/simulate.py` — `create_regime_state_action_space` routes through `internal_regime.state_action_space(regime_params=...)` and overlays per-subject states with `_validate_all_states_present`.
- `src/lcm/pandas_utils.py` — `_is_runtime_grid_param` also recognises action grids.
- `tests/test_runtime_params.py` — TDD tests for runtime action grids: template entry, solve works, runtime ≡ fixed equivalence, varying runtime points changes V, simulate path returns finite V.
- `tests/test_function_output_grid_indexing.py` — TDD tests for the validator across all three discrete-grid sources, plus a regression test that the array-valued-producer + state-indexed-consumer shape (correct code) is **not** flagged.
- `tests/test_grids.py` — TDD tests for the new log-grid / non-finite guards.
- `tests/test_single_feasible_action.py` — regression tests pinning down `map_coordinates` NaN-on-non-finite-coordinate, `get_irreg_coordinate` divide-by-zero on duplicate points, runtime-state-grid `validate_initial_conditions` placeholder feasibility, and CRRA-bequest-with-discrete-state-indexing under `jnp.where`.

## Test plan

- [x] `pixi run -e tests-cpu tests` (880 passed, 5 skipped)
- [x] `pixi run ty`
- [x] `prek run --all-files`
- [x] aca-model long-running end-to-end test (`test_benchmark_model_simulates_end_to_end`) — full 18-regime baseline solve+simulate with runtime-supplied consumption points
- [x] benchmark CI on a clean GPU: aca-baseline +7% peak GPU mem, +10% exec time, +1% compile — all within run-to-run noise

### DIFF
--- a/src/lcm/regime_building/validation.py
+++ b/src/lcm/regime_building/validation.py
@@ -145,14 +145,12 @@ def validate_logical_consistency(regime: Regime) -> None:
 def _validate_function_output_grid_indexing(regime: Regime) -> list[str]:
     """Detect the regime-function-output / discrete-grid-indexed-input name clash.
 
-    A regime function whose output is then re-indexed by a discrete grid (state,
-    action, or derived categorical) inside another consumer (function,
-    constraint, or transition) is a silent footgun: pylcm broadcasts function
-    outputs to per-cell scalars before consumption, so the indexing produces
-    NaN at runtime instead of the intended scalar.
-
-    The safe pattern is to take the discrete grid as input on the producing
-    function and return the scalar directly.
+    The unsafe pattern is: a regime function `f` takes a discrete grid `g`
+    (state, action, or derived categorical) as an input — so `f`'s output
+    is a per-cell scalar — and a consumer then indexes `f[g]`. The
+    consumer is indexing a 0-d array by a scalar integer, which raises
+    `IndexError` at trace time. The fix is to drop the redundant `[g]`
+    in the consumer (or refactor `f` not to take `g`).
     """
     function_output_names = set(regime.functions)
     discrete_grid_names = (
@@ -166,6 +164,19 @@ def _validate_function_output_grid_indexing(regime: Regime) -> list[str]:
     )
     if not function_output_names or not discrete_grid_names:
         return []
+
+    # Only treat `func_output[grid]` as unsafe when the producing function
+    # *also* takes `grid` as an input — that is the case where the output
+    # is per-cell scalar and the consumer's indexing is wrong. If the
+    # producing function does not take `grid`, its output shape is
+    # whatever it computed (typically an array indexable by `grid`) and
+    # the consumer pattern is correct.
+    function_inputs: dict[str, set[str]] = {}
+    for name, func in regime.functions.items():
+        try:
+            function_inputs[name] = set(inspect.signature(func).parameters)
+        except ValueError, TypeError:
+            function_inputs[name] = set()
 
     consumers: list[tuple[str, Callable]] = []
     consumers.extend(regime.functions.items())
@@ -181,14 +192,17 @@ def _validate_function_output_grid_indexing(regime: Regime) -> list[str]:
             discrete_grid_names=discrete_grid_names,
         )
         for func_output_name, grid_name in clashes:
+            if grid_name not in function_inputs.get(func_output_name, set()):
+                continue
             errors.append(
                 f"Consumer '{consumer_name}' indexes regime function output "
                 f"'{func_output_name}' by discrete grid '{grid_name}' "
-                f"(`{func_output_name}[{grid_name}]`). pylcm broadcasts "
-                f"function outputs to per-cell scalars before consumption, so "
-                f"this indexing silently produces NaN. Refactor "
-                f"'{func_output_name}' to take '{grid_name}' as input and "
-                f"return the scalar directly."
+                f"(`{func_output_name}[{grid_name}]`), but '{func_output_name}' "
+                f"already takes '{grid_name}' as input — its output is a "
+                f"per-cell scalar, so the indexing raises IndexError at trace "
+                f"time. Drop the redundant `[{grid_name}]` in '{consumer_name}', "
+                f"or refactor '{func_output_name}' not to take '{grid_name}' "
+                f"as input."
             )
     return errors
 
--- a/src/lcm/regime_building/validation.py
+++ b/src/lcm/regime_building/validation.py
@@ -128,7 +128,7 @@ def validate_logical_consistency(regime: Regime) -> None:
 
     error_messages.extend(_validate_active(regime.active))
     error_messages.extend(_validate_state_transitions(regime))
-    error_messages.extend(_validate_function_output_state_indexing(regime))
+    error_messages.extend(_validate_function_output_grid_indexing(regime))
 
     states_and_actions_overlap = set(regime.states) & set(regime.actions)
     if states_and_actions_overlap:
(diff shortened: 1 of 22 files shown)


hmgaudecker:
### Code review

Found 6 issues (reporting all, no score filter applied per request).

1. Decorative section-separator comment blocks in a new test file (AGENTS.md says "Never add decorative section-separator comments like `# ---...---`. Code structure should be self-evident from function names and ordering.") Three blocks of dashed banners were added.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_single_feasible_action.py#L246-L251

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_single_feasible_action.py#L397-L408

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_single_feasible_action.py#L448-L457

2. Test helpers added in this PR have parameters with no type annotations (AGENTS.md says "Type Hints — Always use type hints in all function signatures. This is mandatory."). `_make_action_grid_model(*, consumption_grid)` and `_make_action_grid_model_with_stateful_dead(*, consumption_grid)` lack annotations on `consumption_grid`; `_crra_bequest` and `_alive_utility` similarly omit annotations on `pref_type`, `consumption_weight`, `coefficient_rra`.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_runtime_params.py#L153-L175

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_runtime_params.py#L243-L286

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/tests/test_single_feasible_action.py#L264-L294

3. Simulate's new `create_regime_state_action_space` no longer runs `_validate_all_states_present` (bug due to bypass: the previous path through `create_state_action_space(states=...)` validated that the per-subject `states` mapping covered every required state name; the new path calls `base.replace(states=...)` directly and skips that check). If a future caller forgets a state, the failure deferred to a less helpful site.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/src/lcm/simulation/transitions.py#L52-L61

vs the dropped validation in:

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/src/lcm/state_action_space.py#L55-L62

4. New private helper has no docstring (AGENTS.md says "Docstrings — Use Google convention. Imperative mood in summary lines."). Sibling helpers in the same file (`_add_runtime_grid_params`, `_collect_all_functions_for_template`, `_validate_no_shadowing`) all carry docstrings; this one was missed.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/src/lcm/params/regime_template.py#L107-L121

5. Public `create_regime_params_template` docstring still says template entries match "the state name", but the PR extends the template to runtime-supplied **action** grids (the private helper `_add_runtime_grid_params` has the updated docstring). The public-facing description is now stale.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/src/lcm/params/regime_template.py#L31-L43

6. `_ShockGrid` runtime-param substitution is gated by `in_states and isinstance(spec, _ShockGrid)` — silently a no-op if a `_ShockGrid` ever appears as a continuous action (bug due to asymmetry between the new `IrregSpacedGrid` branch which handles both `in_states` and `in_continuous_actions`). In practice `_ShockGrid` is state-only by design (intrinsic transitions, AGENTS.md forbids them as actions), so the path is unreachable today; flagging for completeness.

https://github.com/OpenSourceEconomics/pylcm/blob/db6214f275d3863fc11a79395ec6b9a0e300db79/src/lcm/interfaces.py#L266-L296

hmgaudecker (commented):
Autoreview.

@timmens @mj023: I am sorry this became larger once more -- I needed runtime-supplied points on an action grid, and this surfaced lots of other things which seemed in scope initially. Hope it's still reasonable, but if you want me to tear the PR apart, lmk!

hmgaudecker:
Note: aca-baseline performance regression is because the model [switched](https://github.com/OpenSourceEconomics/aca-model/pull/8) to using an `IrregSpacedGrid` with runtime-supplied points so XLA can't propagate the action values through algebraic simplifications

src/lcm/regime_building/validation.py
```diff
@@ -138,6 +142,103 @@ def validate_logical_consistency(regime: Regime) -> None:
         raise RegimeInitializationError(msg)
 
 
+def _validate_function_output_grid_indexing(regime: Regime) -> list[str]:
```
mj023, line 145: I would prefer to not do this check. I think it's quite intuitive that indexing like this is wrong and checking all user functions for mistakes like this is too much. I understand that it can be hard to debug just from the error message, but I think the examples are enough to show how to correctly use additional functions.

mj023 (commented):
Everything else looks correct.

src/lcm/regime_building/validation.py
hmgaudecker: It's always a fine line between trying to error early and expecting a certain level of understanding from the user. In this case, it actually happened to me when I got a little lost during refactoring and forgot an instance. 

It would be great if you could spell out what precisely you are worried about? E.g., runtime, fragility of the approach, ...?

timmens (approved):

src/lcm/regime_building/validation.py
mj023: I think checking for this is quite complicated and there are many ways to hide the pattern from the check and still get essentially the same error. Users are also supposed to mostly treat their functions like they were using scalars, so it should be clear that this won't work. I just want to avoid that we start checking the user supplied functions for all sorts of coding errors, as the same justification can be made for many other cases, like for example a user having a function `utility( consumption, ..)` and thinking they can index into the consumption variable `consumption[0]` because they provided consumption as a grid.

If you both want this check then I'm fine with it, but we should really try not to add too many of these, as they add complexity and are not very reliable.

hmgaudecker: I agree on the general sentiment. At the same time, I don't think this will hurt much and do at least some good (it certainly would have saved me a couple of hours had it been around last week). 

So let's keep it in but remove it should it ever cause trouble!

PR #339, opened 2026-05-01
solve_brute: lazy diagnostics with per-period fail-fast on NaN

hmgaudecker:
## Summary

Three coupled fixes to the solve-time diagnostics path. All three were necessary to keep the production aca-baseline solve healthy on a 16 GB V100 *and* surface NaN failures usefully.

### 1. Stream NaN/Inf reductions instead of stacking-and-flushing (`2cc46ff`)

PR #334 introduced a deferred-diagnostics accumulator: every `(regime, period)` appends `jnp.any(jnp.isnan(V_arr))` / `jnp.any(jnp.isinf(V_arr))` to a Python list, the lists are stacked, and the stacks are `.tolist()`-ed to host at end of solve. The win was zero per-period host transfers — important for MSM-style estimation loops that re-solve hundreds of times.

The cost: on a 16 GB V100 at production aca-baseline grid sizes, the stacked reduction graph keeps every period's `isnan(V_arr)` / `isinf(V_arr)` intermediate alive simultaneously. The post-loop `.tolist()` asks XLA to compile the fan-in, which then requests ~7.3 GiB on top of the already-resident solution V arrays and OOMs. Symptom: `solve()` reports every age as "finished in ~14 ms" (dispatch-async times, not actual compute), then `JaxRuntimeError: RESOURCE_EXHAUSTED` at the first `.tolist()`.

Fix: replace the list-append with a running scalar `OR` and add a per-period `block_until_ready()` so each period's reduction kernel finishes (and its intermediate is freed) before the next period dispatches. End of solve transfers two bools to host; on a healthy solve we return without materialising any per-row state. Failure paths walk `diagnostic_rows` and `.item()`-localise the offender.

### 2. Stop pinning per-period V templates in `diagnostic_rows` (`365da07`)

The streaming reduction landed first addressed only the per-period reduction buffers; row-level retention was the larger leak. Each `_DiagnosticRow` was holding the active-period `state_action_space`, the rolling `next_regime_to_V_arr`, the regime's flat params, and a `compute_intermediates` closure (which itself captured the state_action_space). At production grid sizes — 50+ periods × ~6 active regimes — the accumulated references pinned every period's full-shape V mapping in device memory, OOMing the V100 mid-loop on `block_until_ready` (the next allocation that has nowhere to go).

Fix: strip `_DiagnosticRow` to the three Python scalars actually needed for failure-path localisation (`regime_name`, `period`, `age`) and reconstruct the heavy bits from `solution`, `internal_regimes`, and `internal_params` inside `_raise_at`. The reconstruction mirrors the loop's roll-forward semantics: for each regime, take the smallest later period in `solution` where the regime was active, falling back to a zeros template — the same value the rolling `next_regime_to_V_arr` slot held during the live dispatch.

### 3. Fail-fast on NaN per period; rewrite stale diagnostic hint (`bf1cdf4`)

Two finishing touches that surfaced once the streaming reduction was working:

- **Fail-fast at age boundary.** Adds a per-period `running_any_nan.item()` host transfer right after the existing `block_until_ready`. On `True`, the loop breaks out and the existing post-loop emitter raises immediately. Cost: one scalar bool transfer per period — negligible next to `max_Q_over_a`. Without this, backward induction would finish the entire age range (potentially ~2h on production grids) before raising at the first-NaN row, leaving the user staring at an idle-looking solve. Inf stays non-fatal; the post-loop warning still fires for any period that flagged it.
- **Drop the misleading "re-solve with debug logging" suggestion** from `validate_V`. The diagnostic `[NOTE]` is added inline by `_enrich_with_diagnostics` whenever `compute_intermediates` is wired up — i.e. on the default path — so suggesting a re-solve to "produce" diagnostics is wrong: they were already produced. Replace with a pointer to the `[NOTE]` for the per-axis breakdown plus a mention of `log_path=...` for snapshot persistence (the only thing debug-mode actually adds beyond the inline diagnostic).

## Why this preserves PR #334's win

`block_until_ready()` is a device-only sync. No host transfer, no PCIe round-trip — it only waits for the kernel to finish. PR #334 removed `.item()` / bool-check syncs that *did* round-trip; that knob stays off. In practice the small `jnp.any` reduction has finished by the time `max_Q_over_a` (~14 ms/period) returns, so the call is near-free. The fail-fast `.item()` (added in commit 3) only fires when diagnostics are enabled (`log_level >= "warning"`); `log_level="off"` skips even that, preserving the tight estimation-loop contract.

## Changes

- `src/lcm/solution/solve_brute.py`
  - Replace `diagnostic_any_nan` / `diagnostic_any_inf` lists with `running_any_nan` / `running_any_inf` scalars folded via `|` per `(regime, period)`.
  - Per-period `block_until_ready()` on the running scalars (and on the last-appended `mean` when stats are enabled).
  - Per-period `running_any_nan.item()` for fail-fast on NaN; loop breaks on `True`.
  - End-of-loop `_emit_post_loop_diagnostics` orchestrator: `.item()` on each running scalar; only on `True` does it walk `diagnostic_rows` and `.item()`-localise the offending `(regime, age)` for `_raise_at` / `logger.warning`.
  - `_DiagnosticRow` reduced to `(regime_name, period, age)` Python scalars; `_raise_at` reconstructs `state_action_space` / `next_regime_to_V_arr` / `compute_intermediates` from `solution` + `internal_regimes` + `internal_params`.
  - Drop `_StackedReductions`, `_emit_deferred_diagnostics`, the per-row-flag-list variants of `_raise_if_nan` / `_warn_if_inf`. Replace with `_raise_first_nan_row`, `_warn_inf_rows`, `_log_per_period_stats`.
- `src/lcm/utils/error_handling.py`
  - Drop the misleading "re-solve with debug" suggestion from the `InvalidValueFunctionError` message.
- `tests/solution/test_diagnostics.py` (new)
  - Happy-path solve at `log_level="warning"` returns finite V.
  - NaN-bearing solve raises `InvalidValueFunctionError` with the offending regime in the message.
  - `log_level="off"` returns and emits zero `WARNING+` records even when the params would NaN.
  - `log_level="debug"` emits ≥1 per-`(regime, period)` `V min=… max=… mean=…` log line.

## Test plan

- [x] `pixi run -e tests-cpu tests` (885 passed, 5 skipped)
- [x] `prek run --all-files`
- [x] On a V100 box: production `task_simulate_aca` runs without the OOM at end of solve.
- [x] On a V100 box: a NaN-injecting params variant now raises within ~10 min of the offending age, not after the full ~2h backward induction.
- [ ] ASV benchmark on a clean GPU: confirm Mahler-Yum / Precautionary Savings within run-to-run noise of #338's baseline. The aca-baseline benchmark may move slightly because of the per-period `block_until_ready`; expect single-digit % overhead at most.

## Followup (separate change, after this lands)

In `aca-estimation/src/aca_estimation/task_simulate_baseline.py` and `task_simulate_aca.py`: drop the `log_level="off"` workaround that was added when the original OOM blocked production runs.

### DIFF
--- a/src/lcm/solution/solve_brute.py
+++ b/src/lcm/solution/solve_brute.py
@@ -73,15 +73,29 @@ def solve(
 
     solution: dict[int, MappingProxyType[RegimeName, FloatND]] = {}
 
-    # Async diagnostics accumulators: every `jnp.any(isnan)`,
-    # `jnp.any(isinf)` (and the debug min/max/mean trio) lives here as
-    # a device-side scalar during the hot loop. No host sync happens
-    # until the single flush in `_emit_deferred_diagnostics` post-loop.
-    # This replaces the pre-existing synchronous `log_nan_in_V` +
-    # `log_V_stats` + `validate_V` triple, which forced one host
-    # transfer per (regime, period) — ~n_regimes * n_periods stalls
-    # per solve, a meaningful throughput tax in MSM-style loops.
-    # Both gates fall out of the public log level: `"off"` ⇒ nothing,
+    # Async diagnostics accumulators: per-period `jnp.any(isnan)` /
+    # `jnp.any(isinf)` (and the debug min/max/mean trio) live here as
+    # device-side scalars during the hot loop. The two NaN/Inf flags
+    # are folded into single running scalars; the per-period min/max/
+    # mean trio is appended to a list (only emitted at debug, where
+    # we genuinely want every number on host).
+    #
+    # Per-period `block_until_ready()` after the running update forces
+    # the device kernel to finish before the next period dispatches.
+    # This frees the per-period `isnan(V_arr)` / `isinf(V_arr)`
+    # intermediate buffers (~2 MB each at production grid sizes) so
+    # they don't stack up. `block_until_ready` is a *device-only* sync
+    # — no host transfer, no PCIe round-trip — so it doesn't
+    # re-introduce the per-period host stalls that #334 removed; if
+    # `max_Q_over_a` (the dominant per-period kernel) is in flight,
+    # the call returns immediately when the small reduction is done.
+    #
+    # One host transfer per stat at end of solve (`.item()` on the
+    # running scalars) decides whether to enter the failure-path
+    # localisation. On a healthy solve no per-row materialisation
+    # happens.
+    #
+    # Gate falls out of the public log level: `"off"` ⇒ nothing,
     # `"warning"` / `"progress"` ⇒ NaN/Inf only, `"debug"` ⇒ adds the
     # min/max/mean trio. `"off"` skips even the NaN fail-fast — that
     # is the documented contract of `"off"` (suppress all output) and
@@ -92,8 +106,8 @@ def solve(
     diagnostic_min: list[FloatND] = []
     diagnostic_max: list[FloatND] = []
     diagnostic_mean: list[FloatND] = []
-    diagnostic_any_nan: list[FloatND] = []
-    diagnostic_any_inf: list[FloatND] = []
+    running_any_nan: FloatND = jnp.zeros((), dtype=bool)
+    running_any_inf: FloatND = jnp.zeros((), dtype=bool)
 
     logger.info("Starting solution")
     total_start = time.monotonic()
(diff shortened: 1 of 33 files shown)


hmgaudecker (commented):
Autoreview

tests/test_nan_diagnostics.py
```diff
@@ -146,6 +146,28 @@ def borrowing_constraint(
...
+    """
+    from lcm.solution.solve_brute import _DiagnosticRow  # noqa: PLC0415
+
+    expected = {"regime_name", "period", "age"}
+    actual = set(_DiagnosticRow.__dataclass_fields__)
+    assert actual == expected, (
+        f"_DiagnosticRow must hold only {expected}; got {actual}. Adding "
+        "device-backed fields here pins per-period V templates in device "
+        "memory and re-introduces the OOM during long backward inductions."
+    )
```
timmens: I feel like this test can be deleted?

This seems like Claude Code testing that it correctly removed something from somewhere. Very _claudish_ and not useful IMO. But maybe I am missing something.

src/lcm/solution/solve_brute.py
```diff
@@ -390,11 +418,17 @@ def _get_regime_V_shapes(
...
+    every (regime, period) row stays at a few bytes. The expensive bits
+    (state-action space, next-period V mapping, params, the
+    `compute_intermediates` closure) are reconstructed lazily on the
+    failure path from `internal_regimes`, `internal_params`, and the
+    partial `solution` that has been built up to that point.
+
+    The earlier design captured those device-backed objects directly on
+    each row, which pinned every period's V template in device memory
+    until the post-loop flush — at production grid sizes that hits OOM
+    well before the loop completes.
```
timmens: This docstring, and many other texts sound very _claudish_ (Claude Code generated). If I would write the docstring for a human developer, I would write it differently.

While the info in this docstring is important; I am not sure what the effect will be once all updated docstrings will read like this.

However, potentially this is exactly what an agent needs to read in the future. But I am not sure. Not saying that this needs to be changed, just raising this issue.

timmens (approved):

tests/test_nan_diagnostics.py
hmgaudecker: Deleted!

src/lcm/solution/solve_brute.py
hmgaudecker: Fixed, and added instructions to AGENTS.md etc. (accidentally in #342).

PR #340, opened 2026-05-01
Model.n_subjects: AOT-compile simulate, lock integer dtype to int32

hmgaudecker:
## Goal

Do the same for simulate as we did for solve before: AOT-compile. Main reason is to fail early and get cleaner benchmarks (sim-jit was included in runtime previously).

This showed subtle differences re what int types were used during compilation and runs. Explicit casts inside the hot loop turned out to be necessary, but very costly. For speed / mem efficiency / peace of mind, convert everything to `jnp.int32` at the boundary (no use cases will ever require ints > 2_147_483_647).

## Summary

Two coupled changes that together let `simulate(...)` ship a single AOT-compiled program per unique callable and batch shape:

### 1. `Model(n_subjects: int | None = None)` — AOT-compile simulate

- When `n_subjects` is set, the first matching `simulate(...)` call AOT-compiles every unique simulate function (`argmax_and_max_Q_over_a` per period, `next_state` per regime, `compute_regime_transition_probs` per regime) for batch shape `n_subjects` in parallel via `ThreadPoolExecutor`, mirroring solve's existing AOT path in `solve_brute._compile_all_functions`.
- Subsequent calls with the same size hit the cache; calls with a mismatching size warn once per size and fall back to the runtime-traced path.
- Side benefit: normalises `period_to_regime_to_V_arr` at the entry of `simulate(...)` so every period dispatches with the same pytree (active-regime padding with zeros). Without this the last period's empty `next_regime_to_V_arr` breaks both the AOT pytree signature and JAX's own JIT cache reuse.
- `n_subjects=None` (the default) preserves the previous lazy behaviour.

### 2. Lock integer dtype to `int32` end-to-end (formerly #341, squash-merged as `6c610d1`)

The lazy JIT cache silently retraces per dtype, so `int32` (no x64) vs `int64` (x64) variants of the same regime compiled into different specialisations. AOT compile via `jax.jit(...).lower(**args).compile()` ships a single signature and broke at runtime with `int32[N] vs int64[N]` mismatches. Fix:

- Tighten `Int1D`/`IntND`/`DiscreteState`/`DiscreteAction`/`ScalarInt` in `src/lcm/typing.py` from `Int` to `Int32` — ~113 internal usages inherit the dtype constraint without further edits.
- Pin `DiscreteGrid.to_jax()` to `int32` regardless of `jax_enable_x64`.
- `build_initial_states` casts discrete states to grid dtype (one-shot fix for the period-0 vs period-1+ dispatch dtype split).
- Boundary casts at every site whose integer dtype actually depends on `jax_enable_x64`:
  - `regime_building/argmax.py:46, 68` — `argmax` and scalar fallback (real `int64→int32` narrowing under x64).
  - `simulation/initial_conditions.py:365, 409` and `utils/error_handling.py:376` — `where` index extractions in error/validation paths.
- Lock the AOT signature inputs:
  - `pandas_utils.py:155` — regime-id ingestion cast to `int32` to match `DiscreteGrid.to_jax()`.
  - `simulation/simulate.py:102` — `subject_regime_ids` sentinel buffer pinned to `int32` regardless of input regime dtype.

Note: `jnp.searchsorted` already returns `int32` even under x64, and `unravel_index` outputs are immediately consumed by index ops that accept any int dtype — so no inline casts are needed inside hot interpolation/lookup kernels (an earlier draft added them, which broke XLA fusion and roughly doubled GPU peak memory on aca-baseline).

## Test plan

- [x] `pixi run -e tests-cpu pytest tests/simulation/test_simulate_aot.py` — 7 new tests cover validation, status-quo regression, AOT-compile-once, mismatch warn, warn-once-per-size.
- [x] `pixi run -e tests-cpu pytest tests/ -n 7` — 895 passed, 5 skipped (full pylcm suite green; +3 dtype-invariant tests in `tests/test_int_dtype_invariants.py`).
- [x] aca-dev simulate-AOT smoke — `simulate OK, n=20` with all int initial-condition entries reporting `int32`.

Stacked on #339.

### DIFF
--- a/src/lcm/simulation/initial_conditions.py
+++ b/src/lcm/simulation/initial_conditions.py
@@ -7,7 +7,7 @@ Consolidates initial condition construction (`build_initial_states`) and validat
 
 from collections.abc import Callable, Mapping, Sequence
 from types import MappingProxyType
-from typing import Never
+from typing import Never, cast
 
 import jax
 import numpy as np
@@ -25,6 +25,7 @@ from lcm.interfaces import InternalRegime
 from lcm.regime_building.Q_and_F import _get_feasibility
 from lcm.typing import (
     ActionName,
+    FlatRegimeParams,
     InternalParams,
     RegimeName,
     RegimeNamesToIds,
@@ -579,11 +580,21 @@ def _check_regime_feasibility(  # noqa: C901
     if not action_names:
         return None
 
-    grids = MappingProxyType(
-        {name: spec.to_jax() for name, spec in internal_regime.grids.items()}
+    # Build the state-action space with runtime-supplied grid points
+    # substituted. The base grid's `to_jax()` raises for runtime-supplied
+    # `IrregSpacedGrid`s declared with `pass_points_at_runtime=True`, so the
+    # validator must read points from `state_action_space(regime_params=...)`.
+    state_action_space = internal_regime.state_action_space(
+        regime_params=cast("FlatRegimeParams", MappingProxyType(dict(regime_params))),
     )
-
-    flat_actions = _build_flat_action_grid(action_names=action_names, grids=grids)
+    action_grids: dict[str, Array] = {
+        **state_action_space.discrete_actions,
+        **state_action_space.continuous_actions,
+    }
+    flat_actions = _build_flat_action_grid(
+        action_names=action_names,
+        grids=MappingProxyType(action_grids),
+    )
 
     filtered_params = {k: v for k, v in regime_params.items() if k in accepted}
     state_names = list(internal_regime.variable_info.query("is_state").index)
--- a/src/lcm/simulation/initial_conditions.py
+++ b/src/lcm/simulation/initial_conditions.py
@@ -64,10 +64,24 @@ def build_initial_states(
     for regime_name, internal_regime in internal_regimes.items():
         for state_name in _get_regime_state_names(internal_regime):
             key = f"{regime_name}__{state_name}"
-            if state_name in initial_states:
+            grid = internal_regime.grids[state_name]
+            if isinstance(grid, DiscreteGrid):
+                # Match the grid's index dtype so the state is index-stable
+                # across the simulate loop. Without this, period-0 dispatch
+                # carries the user-supplied dtype (often int32) but post-
+                # transition states are promoted to the grid dtype (int64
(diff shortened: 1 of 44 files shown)


hmgaudecker:
Note:


| Benchmark | Statistic | before | after | Ratio | Alert |
|---|---|---|---|---|---|
| aca-baseline | peak GPU mem | 671 MB | 2.73 GB | 4.08 | ❌ |

happens because of changes to the benchmark itself, see [clean test](https://github.com/OpenSourceEconomics/pylcm/pull/344#issuecomment-4387573256).

src/lcm/simulation/initial_conditions.py
```diff
@@ -653,15 +663,107 @@ def _check_combo(action_kw: dict[str, Array]) -> Array:
     if not infeasible_indices:
         return None
 
+    per_constraint_admits_any = _per_constraint_feasibility(
```
hmgaudecker, line 666: This helps a lot for debugging invalid initial conditions. Sorry it landed here, making the PR bigger...

hmgaudecker (commented):
Autoreview.

hmgaudecker:
Note:


| Benchmark | Statistic | before | after | Ratio | Alert |
|---|---|---|---|---|---|
|  | peak CPU mem | 8.39 GB | 10.36 GB | 1.23 | ❌ |


If this ever were the bottleneck, reducing `max_compilation_workers` would get this down at a one-off cost.

Plus it seems to be [gone in #345](https://github.com/OpenSourceEconomics/pylcm/pull/345#issuecomment-4393051624).

timmens (approved):
This looks good to me!

If we want to enforce that we only ever use jnp.int32, we could add a pre-commit hook or simple test that checks that int64 does not occur in src/ (except for processing.py).

hmgaudecker:
> If we want to enforce that we only ever use jnp.int32, we could add a pre-commit hook or simple test that checks that int64 does not occur in src/ (except for processing.py).

Nice suggestion! 

I think we're mostly good with ty checking for (`ScalarInt` / `IntND` / `BoolND`) plus the `dtype` barrier in `processing.py`. A simple grep wouldn't add much to that, if anything. Another `ast.walk` would probably be overkill here. 

Also, I have promoted the dtype truncation warning to an error now, so this should surface quickly in any case.

PR #342, opened 2026-05-04
regime_template: exempt next_<state> names from fixed_param extraction

hmgaudecker:
## Summary

A state transition that consumes another state transition's output (e.g.
`next_wealth(next_aime, ...)`) was rejected at solve time with:

```
InvalidParamsError: Missing required parameter:
'<regime>__next_wealth__next_aime'
```

The cause is in `create_regime_params_template`: it classifies `next_<state>`
references inside transition signatures as regime-level fixed_params and
`_rename_params_to_qnames` qualifies them, so `_resolve_fixed_params`
demands them at solve time.

But `get_next_state_function_for_solution` already merges all transitions
and DAG functions into a single dict before calling `concatenate_functions`,
so dags resolves the chain at evaluation time. The block is purely the
params-template step.

This PR extends the exempt set with `{f"next_{name}" for name in regime.states}`
and lets dags do its job.

## Motivation

aca-model needs this for the French & Jones 2011 pension imputation
correction: a per-target `next_assets` factory that reads `next_aime` to
compute next-period imputed pension wealth. Without this fix, the natural
DAG composition (`next_assets` consuming `next_aime`) is unreachable
through the public Regime API.

## Tests

- Unit test in `tests/regime_building/test_create_regime_params_template.py`
  asserting `next_aime` is not surfaced as a fixed_param when another
  transition consumes it.
- End-to-end test in `tests/test_chained_state_transitions.py` that builds
  a tiny model where `next_wealth` consumes `next_aime`, runs `solve()`,
  and verifies no NaN values in V.

Verified locally: 882 passed, 5 skipped, both new tests pass; without the
fix the integration test fails with the documented error message.

## Test plan
- [x] `pixi run -e tests-cpu pytest tests -n 7` (882 passed, 5 skipped)
- [x] `pixi run -e type-checking ty` (all checks pass)
- [x] `prek run --all-files` (all hooks pass)
- [x] CI green

### DIFF
--- a/src/lcm/params/regime_template.py
+++ b/src/lcm/params/regime_template.py
@@ -32,7 +32,7 @@ def create_regime_params_template(
 
     Grids with runtime-supplied values (IrregSpacedGrid without points,
     `_ShockGrid` without full shock_params) add entries to the template under
-    pseudo-function keys matching the state name.
+    pseudo-function keys matching the state or action name.
 
     Args:
         regime: The regime as provided by the user.
@@ -112,6 +112,23 @@ def _fail_if_runtime_grid_shadows_function(
     name: str,
     kind: str,
 ) -> None:
+    """Raise if a runtime grid name collides with an existing function name.
+
+    Runtime-supplied state and action grids contribute pseudo-function entries
+    to the params template (keyed by the state or action name). Letting such a
+    pseudo-function entry shadow a real regime function would silently break
+    parameter resolution, so we reject it at template-construction time.
+
+    Args:
+        function_params: Template entries collected so far, keyed by
+            (pseudo-)function name.
+        name: State or action name being added.
+        kind: `"state"` or `"action"`, surfaced in the error message.
+
+    Raises:
+        InvalidNameError: If `name` already exists in `function_params`.
+
+    """
     if name in function_params:
         raise InvalidNameError(
             f"IrregSpacedGrid {kind} '{name}' (with runtime-supplied "
--- a/src/lcm/params/regime_template.py
+++ b/src/lcm/params/regime_template.py
@@ -70,27 +70,53 @@ def create_regime_params_template(
 
     _validate_no_shadowing(function_params, regime)
 
-    for state_name, grid in regime.states.items():
-        if isinstance(grid, IrregSpacedGrid) and grid.pass_points_at_runtime:
-            if state_name in function_params:
-                raise InvalidNameError(
-                    f"IrregSpacedGrid state '{state_name}' (with runtime-supplied "
-                    f"points) conflicts with a function of the same name in the regime."
-                )
-            function_params[state_name] = {"points": "Float1D"}
-        elif isinstance(grid, _ShockGrid) and grid.params_to_pass_at_runtime:
-            if state_name in function_params:
-                raise InvalidNameError(
-                    f"_ShockGrid state '{state_name}' (with runtime-supplied params) "
-                    f"conflicts with a function of the same name in the regime."
-                )
-            function_params[state_name] = dict.fromkeys(
-                grid.params_to_pass_at_runtime, "float"
-            )
+    _add_runtime_grid_params(function_params, regime)
 
     return MappingProxyType(
         {k: MappingProxyType(v) for k, v in function_params.items()}
     )
+
+
+def _add_runtime_grid_params(
+    function_params: dict[FunctionName, dict[str, str]],
+    regime: Regime,
+) -> None:
+    """Add runtime-supplied state/action grid params to the template in place."""
+    for state_name, grid in regime.states.items():
+        if isinstance(grid, IrregSpacedGrid) and grid.pass_points_at_runtime:
+            _fail_if_runtime_grid_shadows_function(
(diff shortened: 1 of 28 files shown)


hmgaudecker (commented):
Autoreview.

src/lcm/regime_building/next_state.py
```diff
@@ -137,64 +148,101 @@ def get_next_stochastic_weights_function(
...
+    def combined(*args: Array, **kwargs: Array) -> dict[str, Array]:
+        if args:
+            kwargs = {**dict(zip(all_args, args, strict=False)), **kwargs}
+        out: dict[str, Array] = {}
+        for target, func in per_target_funcs.items():
+            target_kwargs = {arg: kwargs[arg] for arg in target_args[target]}
+            target_out = func(**target_kwargs)
+            for next_state_name, value in target_out.items():
+                out[qname_from_tree_path((target, next_state_name))] = value
+        return out
```
mj023: I think maybe this could be done with `concatenate_functions` inside `get_next_state_function_for_simulation` to be a bit cleaner.

```python
concatenate_functions(
            functions=per_target_funcs,
            targets=list(per_target_funcs.keys()),
            return_type="dict",
            enforce_signature=False,
            set_annotations=True,
)

```

and then just flattening the resulting dictionary in `lcm.simulation.transitions._update_states_for_subjects`  if needed.

mj023 (approved):
I like it, even though it was possible to work around this, there was no reason to restrict the user in this way.

Maybe we could at some point try to get rid of the somewhat strange conversion to qualified names, by making some changes to `simulation.py`.

src/lcm/params/regime_template.py
```diff
@@ -23,8 +23,15 @@ def create_regime_params_template(
...
-    special variables (period, age, E_next_V).
+    function arguments that are not states, actions, other regime functions,
+    `next_<state>` transition outputs, or special variables (period, age,
+    E_next_V).
+
+    The `next_<state>` exemption lets a state transition consume the output of
+    another state transition: dags resolves the chain at evaluation time
+    (`get_next_state_function_for_solution` merges all transitions and DAG
+    functions into a single dict before calling `concatenate_functions`), so
+    these names must not surface as user-facing fixed_params.
```
timmens: This sounds very _claudish_ to me, and I don't understand why we would write such a big paragraph now, but not for `regime.functions`.

- Why is it now an "exemption"?
- Shouldn't `next_state_names` also be part of `H_variables`, since `"E_next_V"` depends on them?

tests/test_chained_state_transitions.py
timmens, line 1: This entire file reads so _claudish_. It seems like Claude is just checking that its implementation works and writing down what the prior state was. An example:

```python
def test_solve_resolves_chain_via_dags() -> None:
    """`solve()` runs and dags wires `next_aime → next_wealth` correctly.

    Before the fix, `_resolve_fixed_params` raised
    `InvalidParamsError: Missing required parameter:
    'active__next_wealth__next_aime'` because `create_regime_params_template`
    classified `next_aime` (a `next_<state>` reference inside another
    transition's signature) as a regime-level fixed_param.
    """
    ...
            assert not jnp.any(jnp.isnan(V_arr))
```

How does this docstring help me (or an agent) in 9 months, and what does this assert not any nan actually check?

---

I feel like we could get better tests and test-docstrings if we would follow a strict AI-assisted TDD approach?

src/lcm/regime_building/next_state.py
```diff
@@ -78,30 +91,28 @@ def get_next_state_function_for_simulation(
...
-    # stochastic next states functions and their weights.
-    extended_transitions = _extend_transitions_for_simulation(
-        all_grids=all_grids,
-        flat_transitions=flat_transitions,
-        variable_info=variable_info,
-        stochastic_transition_names=stochastic_transition_names,
-    )
-    functions_to_concatenate = extended_transitions | dict(functions)
+    per_target_funcs: dict[RegimeName, Callable[..., dict[str, Array]]] = {}
+    for target, target_trans in transitions.items():
```
timmens: Why don't we write it out fully: `target_transitions`?

Definitely in the signature of `_extend_target_transitions_for_simulation`.

timmens (approved):
Very nice! Good that we allow for this now!

src/lcm/regime_building/next_state.py
hmgaudecker: Done, thanks!

src/lcm/regime_building/next_state.py
hmgaudecker: Very good. Done.

tests/test_chained_state_transitions.py
hmgaudecker: 1. Fair enough, fixed.
2. Most of the time, I did that in instructions, but ofc forgot frequently. Added to `AGENTS.md` and `.ai-instructions`.

src/lcm/params/regime_template.py
hmgaudecker: Thanks. Got rid of (a) misleading terminology and (b) `H_variables` entirely -- one assignment to `variables` left.

PR #346, opened 2026-05-08
Add Paralellization across multiple devices during Solve

### LINKED ISSUE #175: ENH: Support multi-GPU computation
For SOTA models most PyLCM users solve and simulate the model on a server, which often has multiple GPUs available.

If multiple GPUs are available, PyLCM should be able to leverage this.

## Where it could help

**Multi-regime case:**

In the multi-regime case, in a given period we compute the value function for each regime. These computations are independent and embarrassingly parallel and could be done on separate hardware.

**Split-up state-space:**

For very large state-spaces that cannot be reduced further using regimes, we could try to divide the state-space into as many pieces as there is hardware available.

mj023:
This PR is a continuation of #147. It uses JAX Auto Parallelization inside jitted functions to make it possible to split the state space across multiple devices and then let every device solve it's part independently.

### Distribution strategy

The grids get a new argument `distributed` that the user can use to specify which grids should be considered for the distribution across devices. If only one grid is marked for distribution, then the length of the grid needs to be a multiple of the available devices, if multiple grids are marked, the product of the lengths needs to be exactly the number of available devices (might be possible to relax this requirement). The grids then need to be moved to the right devices after they have been initialized with runtime-supplied points and shocks. The resulting `VF_arr` will then automatically be split on the devices where each value has been calculated. 

### Distributed Simulation

Because it was easy to implement, the PR also includes parallelization of the simulation. This happens along the axis of subjects, the special state space for the simulation only needs to be distributed to the available devices, for this the number of subjects needs to be a multiple of the number of available devices.

### TODO

- [x] Move grids to right devices after state action space creation
- [x] Fix AOT-Compilation
- [x] Add Distributed Simulation
- [ ] Test communication overhead


hmgaudecker:
Pushed three commits to `distributed` addressing the code-review findings. Happy to revert any individual commit if you'd rather take a different angle.

**1. `c6d5078` — Fix sharding correctness in solve and forbid distributed actions**

- Introduces `_RegimeSharding` and a single `_build_regime_sharding` helper in [interfaces.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/interfaces.py#L319-L463) used by both `_distribute_states_to_devices` and the new `_get_regime_V_shapes_and_shardings` in [solve_brute.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/solution/solve_brute.py#L383-L443). State arrays and V-arrays now share the same mesh and the same device set, fixing the producer/consumer mismatch.
- Policy is now **exact match, no truncation**: the single-grid path requires `n_points % n_devices == 0` (as before), and the multi-grid path requires `prod(grid_sizes) == n_devices` rather than `≤` — silent truncation was a footgun and made geometry mismatches surface late.
- `solve()` now iterates a `dict[RegimeName, _RegimeVTopology]` so non-distributed regimes coexist with distributed ones without the `KeyError` from the previous dict-truthiness check.
- `_reconstruct_next_regime_to_V_arr` applies the same sharding on the diagnostic path so reconstructed zero templates match the live ones.
- New `_validate_distributed_grids` in [regime_building/validation.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/regime_building/validation.py#L145-L163) rejects `distributed=True` on action grids at `Regime` construction with a hint to move the flag to the corresponding state.
- New tests in [test_distributed.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/tests/test_distributed.py) for partial distribution (mixed regimes), the renamed multi-grid \"product must equal devices\" error, and the action-grid validator. The error message in `test_solution_error_if_grid_product_exceeds_devices` is updated to match the new wording.

**2. `0e46529` — Hoist non-parameter field set on shock grids**

`_NON_PARAM_FIELDS: ClassVar[frozenset[str]]` on [`_ShockGrid`](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/shocks/_base.py#L46-L52) keeps `distributed` (and any future `ContinuousGrid` field) out of `params`. Tauchen / Normal / LogNormal overrides extend it via `cls._NON_PARAM_FIELDS | {...}`. New parametrised test in [test_shock_grids.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/tests/test_shock_grids.py#L293-L313) pins the invariant for Normal / LogNormal / Uniform / Tauchen.

**3. `984fac0` — Fix `DiscreteGrid.distributed` docstring and tighten the abstract one**

The `DiscreteGrid.distributed` property had a copy-pasted docstring describing `batch_size` ([discrete.py](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/grids/discrete.py#L57-L60)). Updated, and tightened the [abstract base](https://github.com/OpenSourceEconomics/pylcm/blob/984fac01a1b7f71871a8edb29d5c827672d8efc0/src/lcm/grids/base.py#L19-L27).

Verification: `pixi run --environment tests-cpu tests -n 4` → 962 passed, 10 skipped. `pixi run ty` → clean. `prek run --all-files` → clean.

hmgaudecker (approved):
This is very nice!!! Looking forward to the speed bump!

Before merging, please double check the changes I had Claude make, see [comment above](https://github.com/OpenSourceEconomics/pylcm/pull/346#issuecomment-4443421235).

Thanks a ton!

PR #350, opened 2026-05-11
@categorical: require ScalarInt field annotations + runtime values

### LINKED ISSUE #349: @categorical should require ScalarInt field annotations
## Problem

Leftover from #340: `@categorical(ordered=...)` decorates a class with
integer category codes and annotates each field as `int`. That
annotation is a lie for JAX-side flow: codes appear as `jnp.int32`
everywhere (pref_type, regime ID, transitions, grid indices). The
`int` lie forces every JAX consumer to either wrap with
`jnp.int32(...)` or rely on implicit broadcasting.

Python `int` is only useful at the pandas / Python-interop boundary
(e.g. `to_categorical`, DataFrame column codes). That belongs in a
separate API, not in the decorator that defines the JAX-side category
space.

## Proposal

Require `ScalarInt` (the existing pylcm alias in `lcm.typing`) on
`@categorical`-decorated fields. Surface the violation at decoration
time with a clear error pointing to the offending field if the
annotation is `int` (or anything else that isn't `ScalarInt`).

## Out of scope

- Python/pandas-side conversion helpers (e.g. `to_categorical`) — if
  needed, design as a separate API that returns Python `int`.
- Instance-level access patterns.

hmgaudecker:
## Summary

- Tightens `@categorical` so the annotation contract, runtime values, and validator all agree on `ScalarInt`. Fields annotated otherwise raise the new `CategoricalDefinitionError` at decoration time, and class- / instance-level attribute access (`LaborSupply.work`, `LaborSupply().work`) now return 0-d `jnp.int32` scalars instead of Python ints.
- Adds `lcm.invert_regime_ids` (in `lcm.utils.containers`) so the two pylcm dict-inversion sites can coerce `ScalarInt` keys to Python `int` without spelling out the comprehension. `DiscreteGrid.codes` likewise coerces at the boundary where hashable tuples are needed; the JAX-side surface (`to_jax()`) is unchanged.
- Sweeps every `@categorical`-decorated class across `src/lcm_examples/`, `tests/`, and the user-guide notebooks/docs to `ScalarInt`. The `Effort` example in `mahler_yum_2024/_model.py` switches to a `make_dataclass + type.__setattr__` shim so its 40 `ScalarInt` defaults bypass `dataclass`'s mutable-default check.

Closes #349.

## Test plan

- [x] `pixi run -e tests-cpu tests` — 957 passed, 5 skipped.
- [x] Downstream cascade in aca-model verified locally (216 aca-model tests + 407-test workspace suite green with this pylcm branch as the submodule pointer); separate aca-model PR will follow.


benchmarks/bench_aca_baseline.py
```diff
@@ -71,7 +71,7 @@ def _build() -> tuple[object, object, object]:
         n_subjects=_N_SUBJECTS,
         pref_type_grid=DiscreteGrid(BenchmarkPrefType),
     )
-    _, model_params = get_benchmark_params(model=model)
+    _, _, model_params = get_benchmark_params(model=model)
```
hmgaudecker:
```suggestion
    model_params = get_benchmark_params(model=model)[2]
```

src/lcm/grids/categorical.py
```diff
@@ -70,13 +104,13 @@ def validate_category_class(category_class: type) -> list[str]:
     This validates that:
     - The class is a dataclass
     - It has at least one field
-    - All field values are int
+    - All field values are `ScalarInt` (0-d `jnp.int32` scalars)
```
hmgaudecker:
```suggestion
    - All field values are `ScalarInt`s
```

src/lcm/grids/categorical.py
```diff
@@ -70,13 +104,13 @@ def validate_category_class(category_class: type) -> list[str]:
...
     - All field values are unique
     - Field values are consecutive integers starting from 0
 
     Args:
         category_class: The category class to validate. Must be a dataclass with fields
-            that have unique int values.
+            whose values are unique `ScalarInt` scalars.
```
hmgaudecker:
```suggestion
            whose values are unique `ScalarInt`s.
```

hmgaudecker (commented):
Autoreview.

hmgaudecker:
This completes #340. 

Lots of changes (sorry, @mj023 @timmens), but almost all of them just the expected `int` -> `ScalarInt`, so just have a quick glance.

Just a workaround necessary for RegimeIds since they are mostly used in pure Python, but not worth an exception IMO.

src/lcm/utils/logging.py
```diff
@@ -113,26 +113,26 @@ def log_regime_transitions(
     logger: logging.Logger,
     prev_regime_ids: Int1D,
     new_regime_ids: Int1D,
-    ids_to_names: Mapping[int, str],
+    regime_ids_to_names: Mapping[int, str],
```
timmens: We have `RegimeNamesToId` type, should we have a `RegimeIdsToNames` type as well?

timmens (approved):
Looks good to me!

PR #351, opened 2026-05-11
IrregSpacedGrid: extra_param_names for user-side runtime scalars

### LINKED ISSUE #348: fixed_params should accept keys required for runtime-supplied grids
## Problem

`broadcast_to_template` (and the validator chain that follows) rejects
any `fixed_params` key that no DAG function lists in its signature.
That assumption breaks for runtime-supplied grids: a parameter
referenced by an `IrregSpacedGrid(pass_points_at_runtime=True)` (or
similar grid-spec consumer) is needed at grid-build time, never by a
DAG function — yet it still needs to flow through `fixed_params` so
the per-iteration value can size the grid.

The user-visible friction: callers either funnel the param through
`model.simulate(params=...)` (where it competes with estimation
parameters and arguably doesn't belong) or fabricate a no-op DAG
function that "consumes" it just to satisfy the validator.

Useful when some components of a grid change across optimizer
iterations but others are constant.

## Proposal

Relax the validator so a `fixed_params` key referenced by any
runtime-supplied grid spec is accepted without a DAG-input footprint.
The grid spec already knows which params it needs; surface that to
`broadcast_to_template`.

hmgaudecker:
## Summary

- Adds `IrregSpacedGrid(extra_param_names=("foo", "bar"))` so scalar params consumed by user-side code that builds the runtime points (e.g., an upper bound that changes per optimizer iteration) can flow through `fixed_params` / per-call `params` without tripping `broadcast_to_template`'s unknown-keys check.
- Each name surfaces as a `ScalarFloat` slot in the grid's pseudo-function template entry alongside `points`. Pylcm itself never reads these values — they're carried through so user-side injection code can pick them up via `model.fixed_params[name]` (or the resolved internal params) before computing the points.
- Rejected on fixed-points grids — a baked-in grid has no user-side runtime computation to feed.

Closes #348. Stacks on #350 (uses the `feat/categorical-scalarint` branch as base).

## Test plan

- [x] `pixi run -e tests-cpu tests` — 961 passed (4 new tests in `test_runtime_params.py`).
- [x] Downstream cascade verified in aca-model (216 + workspace 407 green; aca-model PR cascading off PR #10 follows).


hmgaudecker:
Closing without merging.

The motivating use case (aca-model's `consumption_dollars` grid pinned to `max_consumption_dollars`) doesn't actually need the scalar to live in pylcm's params machinery — no DAG function consumes it; only the user-side injector that constructs the gridpoints reads it.

The user-side fix is to pass that scalar as an explicit Python argument to the injector, sourced from wherever it naturally lives (an aca-data pickle, a config module, whatever). pylcm's `fixed_params` then doesn't have to open a per-key hole in `broadcast_to_template`'s unknown-keys check.

Runtime points themselves (`IrregSpacedGrid(n_points=N)` with `params[regime][slot]["points"]` injection) already work as of #338 — that's still the load-bearing mechanism. Just don't carry the construction inputs through `fixed_params`.

Closing #348 along with this PR.

PR #353, opened 2026-05-13
Rewrite qname introspection + structured stochastic factories (#343 phase 2)

hmgaudecker:
Stacked on #352.

## Summary

Phase 2 of [#343](https://github.com/OpenSourceEconomics/pylcm/issues/343): drop the flat-iterate + `tree_path_from_qname` decode pattern from pylcm code. The wrapper-arg `__`-joined strings still exist inside `dags`'s qname encoding (that's the public DAG-arg convention `dags` ships), but pylcm code no longer produces or parses them outside the introduction boundary.

**Three introspection sites — direct nested iteration.** `simulation/transitions.py:111`, `simulation/compile.py:381`, `simulation/result.py:436` iterate `internal_regime.simulate_functions.transitions: Mapping[RegimeName, Mapping[TransitionFunctionName, ...]]` directly instead of flattening and re-parsing the qname. The names that flow into `generate_simulation_keys` are built locally via `qname_from_tree_path((target, transition_name))`.

**Per-leaf stochastic factories — structured args.** `regime_building/next_state.py`'s `_create_discrete_stochastic_next_func` and `_create_continuous_stochastic_next_func` take `target` and `next_state_name` as separate keyword args; the wrapper-kwarg qname is assembled inside the factory. `_create_continuous_stochastic_next_func` additionally takes nested `all_grids` and looks up the shock grid via `all_grids[target][prev_state_name]`, dropping the `name.split("next_")[1]` / `name.replace("next_", "")` parse used to recover components from a pre-joined qname.

**Producer-side f-string concat removed.** `processing.py:599-611`, `next_state.py:143` (`get_next_stochastic_weights_function`), and `Q_and_F.py:516` (`get_next_stochastic_weights_function`'s `_outer` arg list) now go through `qname_from_tree_path` instead of `f"{regime}__next_{shock}"` / `f"weight_{regime_name}__{key}"`.

**DAG topology unchanged.** Per-(target, shock) wrappers stay independent nodes; `dags.concatenate_functions` keeps its pruning behaviour. The flat-name surface that `dags` itself sees is identical to before.

**Out of scope** (per the plan):
- Param-template qname machinery in `params/processing.py` and `model_processing.py:318`.
- DAG function-name qnames inside `regime_building/processing.py:534, 587` — these belong to the broader function-namespace concern, not the regime/state container concern this PR targets.
- The regime-name `__` restriction in `model_processing.py:176-180` — structurally pinned by `dags.tree` elsewhere in pylcm.

## Test plan
[...]

Issue #176, opened 2025-11-25
ENH: Improve type annotations and checking


timmens:
Meta issue to collect ideas to improve type annotations and type checking:

- [ ] Use pandera to get more informative type-hints for pandas dataframes

   *We use pandas dataframe internally (e.g., `variable_info`). It is very hard to check that a column exists or what dtype it has. We should discuss whether we should improve the type annotations for these objects, or replace the dataframes entirely by objects that are easier to type check and give a more descriptive structure.*

- [ ] Use pydantic / beartype (or other tools) to validate user-input.

   *We currently do a very basic validation of user input in each `Model`/`Regime` class. This simply duplicates behavior of better existing validation tools.*

PR #354, opened 2026-05-13
Replace variable_info DataFrame with typed mapping (#176 part 1)

hmgaudecker:
Stacked on #353.

## Summary

Closes part 1 of #176 (the DataFrame-typing half of the issue).

`InternalRegime.variable_info` flips from `pd.DataFrame` with five
boolean columns (`is_state`, `is_action`, `is_continuous`,
`is_discrete`, `is_shock`) to
`MappingProxyType[StateOrActionName, VariableInfo]`, where
`VariableInfo` is a frozen dataclass with three fields:

```python
@dataclass(frozen=True)
class VariableInfo:
    kind: Literal["state", "action"]
    topology: Literal["continuous", "discrete"]
    is_shock: bool
```

The redundant `is_state`/`is_action` and `is_continuous`/`is_discrete`
pairs collapse into `Literal`-typed fields. ty now sees every
attribute access — typos at `vi[name].kind == "stae"` become an
error at the call site instead of a `KeyError` three layers deep.

`VariableInfo` + the `VariableInfoMapping` alias live in
`interfaces.py` next to the other internal data types
(`InternalRegime`, `StateActionSpace`, `PeriodRegimeSimulationData`).
`get_variable_info` is rewritten in pure Python — no pandas
intermediate.

**Shocks keep `topology="discrete"`.** Existing semantics had
`is_continuous = ContinuousGrid and not _ShockGrid`, so shocks were
already tagged discrete despite being mathematically continuous —
shock grids approximate a continuous random variable but are processed
by discrete-sweep machinery. Preserving this avoids silently
reshuffling iteration order in solve/simulate. The docstring on the
`topology` field documents the computational-vs-mathematical
distinction.

**No `__post_init__` validator.** The (kind, topology, is_shock)
triple has illegal combinations the type system can't catch (e.g.
`is_shock=True` with `kind="action"`), but `get_variable_info` is the
only constructor and derives all three from one source. Per CLAUDE.md,
internal data structures don't need defensive validation outside
system boundaries.

**15 consumer sites swept.** Every `variable_info.query("is_state")`
becomes `[k for k, v in vi.items() if v.kind == "state"]` and
similar. The pandas import drops from `interfaces.py`,
`processing.py`, `next_state.py`, and `state_action_space.py`.

**Out of scope** (part 2 of #176): perimeter input validation via
beartype on `Model.__init__` / `Regime.__init__`. Tracked separately.

## Test plan
- [x] `pixi run -e tests-cpu pytest -n 4` — 957 passed, 5 skipped
- [x] `pixi run ty` — clean


hmgaudecker:
I find the code **much** more readable with this, hope you'll agree!

mj023 (approved):
Yes, I think this is a very good change, having this one data structure as a `DataFrame` was always a bit awkward.

PR #355, opened 2026-05-13
Adopt beartype runtime type checking across `lcm`

hmgaudecker:
Closes #176.

Runtime type checking via `beartype` across the entire `lcm` package,
plus the annotation cleanup, type-discipline work, and bug fixes the
checking surfaced. Originally split as #355 / #356 / #357 (perimeter +
scoped claw + annotation cleanup) — collapsed into one PR because the
follow-on work was a single arc.

### Package-wide claw with explicit user-boundary decorators

`src/lcm/__init__.py` registers a single
`beartype_package("lcm", INTERNAL_CONF)` before any submodule loads.
Every module loads with AST-rewritten runtime type checks; violations
default to `BeartypeCallHintViolation` (an internal pylcm bug).

User-facing constructors stack their own `@beartype(conf=PROJECT_CONF)`
on top: `Regime` / `MarkovTransition` / `Model`, every grid and shock
class, `AgeGrid`, `DiscreteGrid`, the `categorical(...)` factory, the
`as_leaf` factory. The explicit decorator wins, so the documented
project exception (`GridInitializationError`, `ModelInitializationError`,
`InvalidParamsError`, `RegimeInitializationError`,
`CategoricalDefinitionError`) keeps surfacing at the user boundary.

### Type-system: user-input vs. canonical

The boundary-vs-internal split is reified in `lcm.typing`:

- `_UserParamsLeaf` / `_ParamsLeaf`, with `UserMappingLeaf` /
  `MappingLeaf` and `UserSequenceLeaf` / `SequenceLeaf` as the
  corresponding leaf containers. The wider user variant accepts Python
  scalars, numpy arrays, `pd.Series`; the narrow canonical variant
  carries only canonical-dtype JAX arrays.
  `cast_params_to_canonical_dtypes` is typed `UserParams → Params`.
- `initial_conditions: Mapping[StateName | Literal["regime"], Array | np.ndarray]`
  at boundary entries (`Model.simulate`, `canonicalize_initial_conditions`);
  `Mapping[..., FloatND | IntND]` downstream.
- State/action-keyed mappings (`state_action_values`, grid mappings,
  `action_grids`) use the `StateOrActionName` / `ActionName` aliases.

### Canonicalization at every user-supplied boundary

`_ShockGrid.params` returns `MappingProxyType[str, ScalarFloat | ScalarInt]`
— Python scalars on dataclass fields are cast to 0-d JAX scalars on
access. `_params_to_jax` and the per-call casts in
[...]

PR #356, opened 2026-05-13
Activate scoped beartype claw on lcm.grids/shocks/params + drift fixes

hmgaudecker:
Stacked on #355.

## Summary

Activates beartype's AST-rewriting claw on `lcm.grids`, `lcm.shocks`,
and `lcm.params` and fixes every annotation lie those three subpackages
were carrying. Perimeter `@beartype` / `@beartype_init` decorators on
user-facing constructors (from #355) are unchanged.

### What's live

The claw registrations sit at the top of `lcm/__init__.py`, **before**
the bulk-import of submodules, so the AST-rewriting import hook gets
installed before any matching submodule loads. Per-subpackage
`BeartypeConf` maps type violations to the project exception most
natural to that scope:

- `lcm.grids` / `lcm.shocks` → `GridInitializationError`
- `lcm.params` → `InvalidParamsError`

A probe call into a deliberately-bad-typed helper (e.g.
`get_linspace_coordinate(value="not a float", ...)`) raises
`GridInitializationError` end-to-end.

The earlier `tests/conftest.py` registration ran *after* `from
lcm._beartype_conf import ...` triggered `lcm/__init__.py`, which
eagerly imports every targeted submodule — the claw was a no-op on
every prior run. Moving the registration into `lcm/__init__.py`
itself fixed that.

### Drift fixed alongside activation

- **`Piece` cumulative `n_points`**: explicit `dtype=jnp.int32` on the
  `_piece_n_points` array and `sum(dtype=jnp.int32)` so the property
  honours its `ScalarInt` annotation under `jax_enable_x64=True`
  (`jnp.sum` otherwise promotes to int64).
- **`_ShockGrid.compute_gridpoints` / `compute_transition_probs` /
  `Tauchen._innovation_variance`**: kwargs widened from `float |
  FloatND` to `float | FloatND | IntND`. Runtime params like `n_std=2`
  arrive as 0-d int arrays after pylcm's canonical-dtype cast.
- **`get_coordinate` on every continuous grid + `get_irreg_coordinate`**:
  accept Python `float` / `int` alongside `ScalarFloat | FloatND`.
- **`process_params(params_template=...)`** widened to
  `Mapping[RegimeName, Mapping[str, object]]` since real callers and
  test fixtures both use plain dicts at the inner levels.
- **`_ParamsLeaf`** adds `np.ndarray` so numpy-array params satisfy
  the perimeter check before pylcm casts them.
- **`create_regime_params_template(regime: ...)` and helpers**,
  **`create_params_template(internal_regimes: ...)`** typed as `Any`
  so duck-typed test mocks (`RegimeMock`, `MockRegime`) satisfy the
  signature — the functions themselves only touch a small structural
  interface.
- **PRNG keys**: every `draw_shock`'s `key:` parameter uses
  `KeyArray` (new alias in `lcm.typing`); JAX PRNG keys have
  `dtype=key<fry>`, not float, so they don't match `FloatND`.
- **`MappingProxyType → Mapping`** at internal entry points where
  callers pass plain `dict` (e.g. `solve_brute.solve`, perimeter of
  `process_params`).
- **JAX scalar/array dtype drift** in `grids/coordinates.py`:
  `get_linspace_coordinate` / `get_logspace_coordinate` widen
  `start: ScalarFloat` / `stop: ScalarFloat` to `ScalarFloat |
  FloatND` (recursive log → linear path passes 0-d arrays).
- **`Never` → `NoReturn`** in `simulation/initial_conditions.py`
  (`typing.Never` is unsupported by beartype 0.22).
- **`persistence.py`** defines `_ModelOrNone` / `_SimulationResultOrNone`
  as `TYPE_CHECKING`-conditional aliases that resolve to `Any` at
  runtime — the bare-union string forward reference `'Model | None'`
  was unparseable by beartype's forward-ref machinery, and importing
  `Model` at the top causes a circular import.
- **Two corrections found in passing**:
  - `dtypes.py:canonical_float_dtype` return type was `jnp.dtype` but
    actually returns `_ScalarMeta` (a class). Changed to `type`.
  - `utils/containers.py:get_field_names_and_values` parameter was
    `dc: type` but the function passes through `dataclasses.fields()`,
    which accepts both classes and instances. Changed to `dc: object`.

### Naming

`v_array` → `V_arr` rename on `_RegimeSharding.V_arr_sharding` and
`_build_zero_V_arr` to match the rest of the solve path (which
already uses `V_arr` / `next_regime_to_V_arr` throughout). Came in
through the merge from `distributed` (#346).

### Test plan
- [x] `pixi run -e tests-cpu pytest -n 4` — 979 passed, 10 skipped
- [x] `pixi run ty` — clean
- [x] Probe call confirms the claw is genuinely live: a wrong-typed
      `get_linspace_coordinate(value="not a float", ...)` raises
      `GridInitializationError` from the claw-installed wrapper, not
      `TypeError` from the bare Python operator inside the function.
- [x] aca-dev workspace tests (414 passed) green against this branch's
      submodule pointer.

### Performance

Full pylcm test suite: ~88-90 s with claw live vs ~86-87 s without —
~3% overhead at suite scale. Inside `jax.jit`, beartype's checks run
during tracing only, so steady-state execution is unaffected.

### Out of scope

Extending the claw to `lcm.regime_building` / `lcm.solution` /
`lcm.simulation` is blocked on a structural conflict:
`lcm.regime_building.processing._wrap_regime_transition_probs` uses
`functools.wraps(user_func)` + dags' `with_signature` to propagate the
user function's annotations onto an internal wrapper. The claw then
decorates that wrapper with `@beartype`, and beartype enforces user
annotations like `age: float` / `period: int` against JIT tracers
inside `jax.jit`. Either annotation stripping at the wrap site, or a
beartype hint-override mapping `float`/`int` to `Array | float | int`,
is needed before that scope can go live.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -527,11 +527,16 @@ Code structure should be self-evident from function names and ordering.
 - **Helper function names follow `{verb}_{qualifier}_noun` patterns.** E.g.,
   `get_irreg_coordinate`, `find_irreg_coordinate`, `get_linspace_coordinate` — not
   `get_coordinate_irreg`.
-- **Use `@overload` when a function accepts both scalar and array inputs.** When a
-  function works with both `ScalarFloat` and `Array`, add overload declarations so the
-  type checker can track `(ScalarFloat) -> ScalarFloat` and `(Array) -> Array`
-  separately. Concrete subclass methods need their own overloads too (not just the
-  abstract base).
+- **Pick the single narrowest jaxtyping alias — never scalar/array `@overload` pairs,
+  never `ScalarX | XND` unions.** ty erases jaxtyping shape annotations: `ScalarFloat`,
+  `Float1D`, and `FloatND` all reveal as `Array`, so scalar/array `@overload` pairs and
+  `ScalarFloat | FloatND`-style unions add zero static precision — they are pure noise.
+  At runtime, beartype treats a 0-d float array as satisfying both `ScalarFloat` and
+  `FloatND`, so `ScalarFloat ⊆ FloatND` and the union is redundant. Annotate each slot
+  with the one alias that matches its genuine rank: `ScalarFloat`/`ScalarInt` for
+  fixed-0-d, `Float1D`/`Int1D` for fixed-1-d, `FloatND`/`IntND` for genuinely rank-
+  polymorphic. Never use a bare `Array` annotation — always reach for the narrowest
+  `lcm.typing` alias.
 - **`func` for callable abbreviations** — use `func`, `func_name`, `func_params` (never
   `fn`). Full word `function(s)` in dataclass field names and public method names.
 - **Singular `state_names` / `action_names`** — not `states_names` / `actions_names`.
--- a/src/lcm/__init__.py
+++ b/src/lcm/__init__.py
@@ -28,10 +28,26 @@ import jax
 with contextlib.suppress(ImportError):
     import pdbp  # noqa: F401
 
-from lcm import shocks
-from lcm._version import __version__
-from lcm.ages import AgeGrid
-from lcm.grids import (
+# Install beartype's AST-rewriting claw on `lcm.grids`, `lcm.shocks`,
+# and `lcm.params` before any submodule of those packages is imported.
+# The claw transforms each matching module's AST at first import to
+# insert runtime type checks; if it isn't registered before the import
+# happens, the affected module loads uninstrumented and `sys.modules`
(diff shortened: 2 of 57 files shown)


hmgaudecker:
Collapsed into #355 — the #355/#356/#357 stack is now a single PR. `feat/beartype-perimeter` was fast-forwarded to this branch's tip, so GitHub flags this as merged; all commits are preserved intact. Review continues on #355.

PR #358, opened 2026-05-17
Harmonize UserX / X pattern across the code base and related renames

hmgaudecker:
## Summary

In most cases, we have converged to using something like `UserParams` for what the user supplies and `Params` for our canonical, internal version. This PR converts the instances that used a different pattern. Plus some related renames that came up in the process.

### 1. `Regime`: keep the user-facing class name, alias everywhere else

Users still type `Regime`. Two distinct classes named `Regime` co-exist in different modules:

| Form | Module | Used by |
| --- | --- | --- |
| User-facing `Regime` | `lcm.user_regime` (re-exported as `lcm.Regime`) | Anyone defining a model |
| Canonical `Regime` | `lcm.interfaces` (not re-exported) | Internal solver / simulator |

**Rule:** bare `Regime` only ever names the user form in `lcm/__init__` (the re-export) and in `src/lcm_examples/` (the showcase package read as user code). Every other file in `src/` and `tests/` aliases the user form locally:

```python
from lcm.user_regime import Regime as UserRegime
```

When a file also needs the canonical form, it stays as the bare name from `lcm.interfaces`. Everywhere else, the module path tells the reader which is which.

The pre-rename `lcm/regime.py` is moved to `lcm/user_regime.py` to remove the read-time ambiguity of two `Regime`-named classes living in `regime.py` and `interfaces.py`.

### Model attribute split

- `model.regimes: MappingProxyType[RegimeName, Regime]` — canonical form, what solve/simulate operate on.
- `model.user_regimes: MappingProxyType[RegimeName, UserRegime]` — boundary snapshot kept for inspection / round-trip.

### Model constructor keyword

`Model.__init__` keeps its existing public keyword — users still write `Model(regimes={...})`. The user-form mapping is stored as `self.user_regimes` after merging in `derived_categoricals`; the processed canonical form is `self.regimes`. The public/internal naming split lives at the field boundary, not at the constructor signature.

### 2. Function protocol: `InternalUserFunction` → `EconFunction` (+ split)

Role-based name — these are all things that are the core economic model as opposed to the DP backend code. Sidesteps the dense `*Function` Protocol family (`RegimeTransitionFunction`, `QAndFFunction`, `MaxQOverAFunction`, …) where a bare `Function` would read as too generic. `UserFunction` (boundary form) is unchanged.

After review, `EconFunction` is further split by role:

- `EconFunction` (`FloatND | IntND`) — utility, `H`, helper/DAG functions. The *economic content* of the model.
- `ConstraintFunction` (`BoolND`) — feasibility predicates stored on `Regime.constraints`.
- `TransitionFunction` (`FloatND | IntND`) — state / regime / shock transition functions stored under `Regime.transitions` and `_ProcessCoreResult.next_regime_func`.

`ConstraintFunctionsMapping` joins `EconFunctionsMapping` / `TransitionFunctionsMapping` as a typed alias. Note: jaxtyping shape aliases erase to bare `Array` for `ty`, so the three protocols collapse to identical signatures at static-check time — the split serves documentation and beartype runtime checks, not ty's narrowing.

### 3. `InternalParams` → `FlatParams`

The terminal-form params type was named for who it's NOT (the user) rather than what it IS (a flat mapping from regime to flat regime params). `FlatParams` describes the structure directly and lines up with the existing `FlatRegimeParams` leaf type.

Followup: the matching local-variable name `fixed_internal` is renamed to `fixed_flat_params` in `_partial_fixed_params_into_regimes` and `_remove_fixed_params_from_template`. The former additionally renames its `regimes` parameter to `raw_regimes` to reflect the precondition that it takes regimes *before* fixed params have been partialled in.

### Internal sweep

- All function parameters and local variables typed as `UserRegime` are now named `user_regime` / `user_regimes`.
- All function parameters and local variables typed as canonical `Regime` keep the unqualified `regime` / `regimes`.
- The transitional `user_regimes = regimes` alias line inside `process_regimes` is dropped — the parameter is renamed instead.

### Mock standardization in tests

The test suite previously mixed `XMock` (suffix) and `MockX` (prefix) for mock classes. Standardized on prefix style throughout, matching the Python ecosystem convention (`MagicMock`, etc.):

- `RegimeMock` → `MockRegime` (module: `tests/regime_mock.py` → `tests/mock_regime.py`)
- `CanonicalRegimeMock` → `MockRegime` (canonical-form subclass in `test_solve_brute.py`)
- `SolveFunctionsMock` → `MockSolveFunctions`

### dags pin

Now that the upstream PR is merged, the pin tracks `main` instead of a feature-branch SHA. Will become a `dags>=0.6` requirement once that release is cut.

### Out of scope

- `UserFacingParamsTemplate` / `ParamsTemplate` — left as is because *we* create the former *for* the user, so different thing.
- `MarkovTransition`, `Model`, `AgeGrid`, grids, shocks — single-form, no rename.

## Test plan

- [x] 992 pylcm tests pass (CPU)
- [x] `ty` clean
- [x] `prek run --all-files` clean
- [x] aca-model cascade + downstream tests
- [x] CI matrix (incl. GPU 32-bit)
- [x] benchmark-pr


docs/user_guide/defining_models.md
```diff
@@ -105,8 +105,7 @@ The `Model` constructor validates:
 After construction, the model exposes several useful attributes:
 
 ```python
-model.regimes  # immutable mapping of user Regime objects
```
hmgaudecker, line 108: Should now document the canonical regime.

pyproject.toml
hmgaudecker: Point to main

src/lcm/model_processing.py
```diff
@@ -24,117 +24,115 @@
...
-        Tuple of internal_regimes and params_template with fixed params
+        Tuple of regimes and params_template with fixed params
         partialled in.
 
     """
-    raw_internal_regimes = process_regimes(
+    raw_regimes = process_regimes(
         ages=ages,
-        regimes=regimes,
+        regimes=user_regimes,
```
hmgaudecker:
```suggestion
        user_regimes=user_regimes,
```

src/lcm/model_processing.py
```diff
@@ -24,117 +24,115 @@
...
     fixed_internal = _resolve_fixed_params(
         fixed_params=dict(fixed_params), template=raw_params_template
     )
     if has_series(fixed_internal):
         fixed_internal = convert_series_in_params(
-            internal_params=fixed_internal,
+            flat_params=fixed_internal,
             ages=ages,
-            regimes=regimes,
+            regimes=user_regimes,
```
hmgaudecker:
```suggestion
            user_regimes=user_regimes,
```
Will check for all such instances.

Though in this one we may actually eat to use`raw_regimes` as name and argument.

src/lcm/model_processing.py
```diff
@@ -24,117 +24,115 @@
...
             regime_names_to_ids=regime_names_to_ids,
         )
     fixed_internal = cast_params_to_canonical_dtypes(fixed_internal)
     _validate_param_types(fixed_internal)
 
     return (
         _partial_fixed_params_into_regimes(
-            internal_regimes=raw_internal_regimes,
+            regimes=raw_regimes,
```
hmgaudecker:
```suggestion
            raw_regimes=raw_regimes,
```

src/lcm/params/regime_template.py
```diff
@@ -15,6 +14,7 @@
     TransitionFunctionName,
     UserFunction,
 )
+from lcm.user_regime import Regime
```
hmgaudecker:
```suggestion
from lcm.user_regime import Regime as UserRegime
```

Internally, we should only ever use that form — only the user-facing interface should use the standalone `Regime`.

src/lcm/typing.py
```diff
@@ -49,10 +49,10 @@
 type RegimeNamesToIds = MappingProxyType[RegimeName, ScalarInt]
 type RegimeIdsToNames = MappingProxyType[int, RegimeName]
 
-type FunctionsMapping = MappingProxyType[FunctionName, InternalUserFunction]
+type FunctionsMapping = MappingProxyType[FunctionName, EconFunction]
```
hmgaudecker:
```suggestion
type EconFunctionsMapping = MappingProxyType[FunctionName, EconFunction]
```
or widen the type hint if that is the correct thing to do.

tests/regime_building/test_regime_processing.py
```diff
@@ -5,10 +5,11 @@
 import pytest
 from numpy.testing import assert_array_equal
 
-from lcm import Regime, categorical
+from lcm import Regime as UserRegime
```
hmgaudecker: Import from `user_regime`

tests/solution/test_solve_brute.py
```diff
@@ -23,10 +23,10 @@ class SolveFunctionsMock:
     compute_intermediates: dict = dataclasses.field(default_factory=dict)
 
 
-class InternalRegimeMock(InternalRegime):
-    """Mock InternalRegime with only the attributes required by solve().
+class CanonicalRegimeMock(Regime):
```
hmgaudecker:
```suggestion
class MockRegime(Regime):
```
We were using the same elsewhere, right?

tests/test_next_state.py
```diff
@@ -21,14 +21,14 @@ def test_get_next_state_function_with_solve_target():
     regime_names_to_ids = MappingProxyType(
         {name: jnp.int32(idx) for idx, name in enumerate(regimes.keys())}
     )
-    internal_regimes = process_regimes(
+    regimes = process_regimes(
         regimes=regimes,
```
hmgaudecker:
```suggestion
         user_regimes=user_regimes,
```

hmgaudecker (commented):
Autoreview.

timmens (approved):
Looks good to me. Two comments, but no showstoppers:

1. I don't like that users have to write `Model(user_regimes=....)` instead of `Model(regimes=...)`. But its also not a big deal.
2. I could also see the term `ModelFunction` as an alternative to `EconFunction`, with the advantage of being a bit more general.

hmgaudecker:
> 1. I don't like that users have to write `Model(user_regimes=....)` instead of `Model(regimes=...)`. But its also not a big deal.

Good catch! My last rename was too greedy, reverted.

> 2. I could also see the term `ModelFunction` as an alternative to `EconFunction`, with the advantage of being a bit more general.

I am afraid "Model" is way too overloaded already in this context. But I took this as an opportunity to distinguish between:

- `EconFunction`
- `ConstraintFunction`
- `TransitionFunction`

now. So we really should have mostly core econ stuff in the first group + helpers (which don't deserve their own name / name component IMO).

PR #359, opened 2026-05-18
Validation cleanup — delete error_handling.py and validation.py

hmgaudecker:
## Summary

Our validation strategy was not clear. Most of the time we had it coupled in the modules where the stuff was happening, but we also had two umbrella modules whose names misled about what lived inside (`lcm/utils/error_handling.py` and `lcm/regime_building/validation.py`). This PR absorbs each validator into the module it actually serves. Furthermore, it streamlines validation by removing redundancies and fixing a couple of inconsistencies.

### File-level outcome

| File | Action |
|---|---|
| `lcm/utils/error_handling.py` | **deleted** |
| `lcm/regime_building/validation.py` | **deleted** |
| `lcm/regime_building/transitions.py` | **new** (`collect_state_transitions` + 2 helpers) |
| `lcm/_transition_checks.py`              | **new** (regime-prob family, top-level private; pre-flight transition-probability validation) |
| `lcm/solution/validate_V.py`             | **new** (per-period value-function NaN check + diagnostic-attachment helpers) |
| `lcm/utils/ast_inspection.py` | **new** (4 AST helpers) |
| `lcm/user_regime.py` | absorbs 8 regime-spec validators + slimmed `validate_transition_probs` |
| `lcm/pandas_utils.py` | drops the import of AST helpers (now imports from `utils/ast_inspection`) |
| `lcm/regime_building/diagnostics.py` | unchanged |

*Note:*  `lcm/user_regime.py` blows up to 744 LOC and the public `validate_transition_probs` sits below the eight private `Regime.__post_init__` validators — a temporary violation of the "public API first" layout rule. This is purely transient if as PR #360 removes `validate_transition_probs` (-175 LOC). If we decide that PR #361 has the right architecture *(it splits this file into `api/regime.py` (class + public function) and `lcm/_regime/` (validators, transition-probs helpers)*, that would reduce further.

### Behaviour changes

- **`validate_transition_probs` is now state-probs-only.** The old function was overloaded: with `state_name` it validated a state-transition array, without it (`state_name=None`) it validated a *regime*-transition array. The regime-probs overload is redundant with `validate_regime_transitions_all_periods`, which runs unconditionally on every `model.solve()` / `model.simulate()` and additionally checks inactive-regime probability and reachability. The slimmed version requires `state_name`. **The overload's `TypeError("does not have a stochastic regime transition")` guard is removed along with it** — that error surface no longer has a reachable code path; misuse is caught instead by `validate_regime_transitions_all_periods` at solve/simulate time.
- **`validate_regime_transition_probs` is now private** (`_validate_regime_transition_probs`). Tests adjust their imports; no other src caller exists.
- **One pre-existing fix folded in:** `_validate_no_reachable_incomplete_targets` (relocated into `_transition_checks.py`) had an override that, when a target regime was entirely absent from the source's `state_transitions`, listed *all* of the target's states as "missing" — including non-stochastic ones that need no explicit entry. The override is dropped; the error now lists only the stochastic transitions actually required.
- Three tests in `tests/test_pandas_utils.py` were dropped: `_valid` and `_wrong_shape` are subsumed by `validate_regime_transitions_all_periods`; `_not_markov_raises` asserted the `TypeError` guard that is removed above, so it goes with the overload. Docs updated accordingly. Everything else is pure mechanical absorption + import-path updates.


hmgaudecker (commented):
Autoreview.

mj023 (approved):

PR #360, opened 2026-05-18
Auto-validate state transitions; make log_level the runtime-validation policy

hmgaudecker:
**Stacked on top of #359.** Merge that one first.

# Summary

pylcm did almost all validation checks internally, except for exposing `validate_transition_probs` and leaving it to the user to validate *state* transition probs, whereas we had an internal mechanism for *regime* transition probs. Furthermore, which validation was done was dependent on a rather obscure / ad-hoc mix of settings.

This PR unifies validation and control over it by adding runtime-validation of state transition probabilities and controlling all costly validation via `log_level` (`off < warning < progress < debug`) and `log_path`:

| `log_level`  | `log_path` | Runtime validation        | Console output                  | Snapshots to disk                                         |
| ------------ | ---------- | ------------------------- | ------------------------------- | --------------------------------------------------------- |
| `"off"`      | (ignored)  | not run                   | silent                          | none                                                      |
| `"warning"`  | `None`     | runs → failures **warn**  | warnings                        | none                                                      |
| `"warning"`  | set        | runs → failures **warn**  | warnings                        | one per warned failure, capped at `log_keep_n_latest`     |
| `"progress"` | `None`     | runs → failures **warn**  | warnings + timing               | none                                                      |
| `"progress"` | set        | runs → failures **warn**  | warnings + timing               | one per warned failure, capped at `log_keep_n_latest`     |
| `"debug"`    | `None`     | runs → failures **raise** | warnings + timing + V_arr stats | none                                                      |
| `"debug"`    | set        | runs → failures **raise** | warnings + timing + V_arr stats | one per solve and on raise, capped at `log_keep_n_latest` |

`validate_transition_probs` is removed from the public API.

`log_level` governs only the *costly runtime* numerical validation — the transition-probability sweeps, the initial-conditions check, and the NaN/Inf check on the value-function arrays. The cheap construction-time sanity checks (regime/model structure, grid definitions, function signatures, the `probs_array` subscript-order check) always run when a `Regime` or `Model` is built, regardless of `log_level`.

## `log_level` is a required argument

`solve()` and `simulate()` no longer default `log_level` — the caller picks it deliberately. Start every project at `"debug"` (fail early, gather full diagnostics) and ease to `"warning"` / `"off"` only once the model is trusted and the run needs the speed or the non-raising behaviour. A loose default would hide that `"debug"` exists; a `"debug"` default would make pylcm look slow.

`simulate()`'s `check_initial_conditions` flag is removed too: initial-conditions validation now follows the same `log_level` policy as every other runtime check.

## Runtime validation

`log_level` is the single knob. The logger it produces carries the policy — there is no separate `validation_mode` value threaded around the engine. It governs four checks, run before backward induction:

- **State transition probabilities** — every `MarkovTransition` state transition is swept over the regime's grid; outcome-axis size, `[0, 1]` range, and sum-to-1 are checked.
- **Regime transition probabilities** — finiteness, `[0, 1]` range, sum-to-1, no probability mass to inactive regimes, and no mass to targets with incomplete stochastic transitions.
- **Initial conditions** (`simulate()` only) — states on-grid, regime IDs valid, at least one feasible action combination per subject.
- **Value function** — NaN/Inf check on each period's `V_arr`, with the offending `(regime, period)` localised.

`"off"` skips all four; `"warning"` / `"progress"` log a warning and let the run continue (the returned solution may carry NaN); `"debug"` raises on the first failure.

## Construction-time guard

`batch_size > 0` paired with `distributed=True` on a single axis is rejected at grid init. Each Python-level batch is its own `jax.jit` dispatch in the solve loop, and on a distributed axis every dispatch carries a cross-device collective; batching therefore multiplies the per-period collective count by `ceil(n_per_device / batch_size)`, and for small `batch_size` the collective overhead per kernel dwarfs the compute per kernel — sharding becomes a regression rather than a speedup. `_fail_if_batch_size_combined_with_distributed` in `grids/base.py` fires from `_init_uniform_grid` (covers `Lin`/`LogSpacedGrid`), `IrregSpacedGrid.__init__`, and `DiscreteGrid.__init__`; piecewise grids inherit `batch_size=0, distributed=False` defaults from `ContinuousGrid` and need no change. Error message points at the right escape valves (more devices or another distributed axis) rather than restoring `batch_size`. Construction-time tests cover all four grid types.

## Test plan

- [x] `pixi run -e tests-cpu tests`
- [x] `pixi run ty`
- [x] `prek run --all-files`


hmgaudecker:
@mj023, in case you are wondering about 867a362, here are some timings. We always distribute the assets grid.

| GPUs | n_assets_bs | n_aime_bs | 3-reg per age | ratio  | 4-reg per age | ratio  |
|------|------------:|----------:|--------------:|-------:|--------------:|-------:|
|   4  |           1 |         0 |    1.2 s      |  7.7×  |     OOM(*)    |   —    |
|   4  |           1 |         1 |    2.4 s      | 15.5×  |     16.2 m    |  8.5×  |
|   1  |           1 |         1 |      –        |   –    |      2.1 m    |  1.10× |
|   1  |           0 |         1 |    155 ms     |   1×   |      1.9 m    |   1×   |
|   4  |           0 |         1 |     94 ms     |  0.61× |      0.41 m   |  0.22× |
|   4  |           0 |         0 |     68 ms     |  0.44× |      0.40 m   |  0.21× |

(*) died at solve running_any_nan.block_until_ready() (RESOURCE_EXHAUSTED: 81.53 GiB on device 0).


I certainly did not expect this...

hmgaudecker:
But ofc the great thing here is that we are supra-linear in the number of GPUs once configured correctly! :tada:

hmgaudecker:
The check just did its job:

```
● I see the root cause: _build_health_trans_cross only emits age 64 in its series;
  pylcm fills other ages with NaN, so the validator sees NaN rows at ages 51-63 where the
  source regime is active.

  Fix: broadcast the age-64 row across the source regimes' full pre-65 active range (51-64)
  so every row sums to 1. 
  Functional behavior unchanged (only age-64 values are ever
  invoked at runtime via the regime-transition gate).
```

*Well, turns out it was too greedy in that case, but still...* :laughing:

mj023 (approved):
Good changes. I can see how batched + distributed would be problematic, good catch! I am wondering if the axis order might also be important, e.g. if it is better to have the distributed axes on the outside in `productmap`.

PR #361, opened 2026-05-18
Split pylcm into a public `lcm/` package and a private `_lcm/` package

hmgaudecker:
## Summary

I found the source code increasingly hard to navigate and to predict what might be the public interface and what we should be free to change (case in point that came up in the process: `validate_transition_probs`, which was removed in #361). I do not want to imagine how that may look for a user who inspects the source code for guidance of what she may use). After a couple of iteration rounds, I think [pytask's strategy](https://github.com/pytask-dev/pytask/tree/main/src) is the right one: Shallow public modules and all the hard work in `_lcm`. See the [docs page](https://pylcm--361.org.readthedocs.build/en/361/architecture/) explaining the internal architecture.

After this PR, `pylcm`'s source is two packages with a hard public/private boundary:

- **`src/lcm/`** — the public surface. Everything a user constructs or consumes: the user-facing classes, the `@categorical` decorator, `as_leaf`, the public type aliases, and the exception classes. `lcm/__init__.py` re-exports the public symbols.
- **`src/_lcm/`** — the private implementation. The build pipeline, the canonical engine dataclasses, the JAX-traced solve / simulate machinery, validators, I/O plumbing, and the engine-side type aliases and protocols.

### Public surface — `src/lcm/`

`__init__.py`, `ages.py`, `categorical.py`, `exceptions.py`, `grids.py`, `model.py`, `params.py`, `persistence.py`, `processes.py`, `regime.py`, `result.py`, `transition.py`, `typing.py`. Users keep writing `from lcm import Model, Regime, ...` — the import path is unchanged.

### Private implementation — `src/_lcm/`

Every engine internal, plainly named (no leading underscore): `engine.py`, `model_processing.py`, `pandas_utils.py`, `state_action_space.py`, `variables.py`, `dtypes.py`, `transition_checks.py`, the bootstrap modules (`jaxtyping_patch.py`, `beartype_conf.py`, `config.py`), `typing.py`, and the `grids/`, `processes/`, `persistence/`, `regime/`, `regime_building/`, `solution/`, `simulation/`, `params/`, `utils/` subpackages.

### Typing split

- `lcm/typing.py` — model-authoring aliases (jaxtyping array shapes, `Period`, `Age`) and the `User*` boundary aliases. Imports nothing from `_lcm`.
- `_lcm/typing.py` — engine-side string labels, compound mapping aliases, canonical post-processing forms, and the structural `Protocol` classes.

### Exceptions

The `PyLCMError` subclasses stay public in `lcm/exceptions.py` and are re-exported from `lcm` — both `from lcm.exceptions import InvalidParamsError` and `except lcm.InvalidParamsError` work. `format_messages`, internal validation plumbing, lives in `_lcm/utils/error_messages.py`.

### Params

`lcm/params.py` exposes `as_leaf` and re-exports `MappingLeaf`, `UserMappingLeaf`, `SequenceLeaf`, `UserSequenceLeaf`. The leaf-class definitions and the engine params machinery live in `_lcm/params/`.

### Bootstrap

`_lcm/__init__.py` applies the jaxtyping `"..."`-sentinel patch. `lcm/__init__.py` imports `_lcm` first, then registers beartype's package claw on both `_lcm` and `lcm` before any submodule loads.

### Renames

- `interfaces.py` has always been confusing me as it sounds too much like API/UI IMO. Went for `engine.py` after a brainstorming session, but happy to adjust that.
- The "shocks" has become too narrow at least since we have been allowing for means and variances to differ from 0/1. I think the correct vocabulary for what those objects describe are (approximations to) *continuous stochastic processes*. The PR updates the docs to make clear that these objects bundle grids and transitions, e.g. [first sentence here](https://pylcm--361.org.readthedocs.build/en/361/continuous-stochastic-processes/). The engine vocabulary has thus become `process` in place of `shock`: the seven `*Process` classes (`UniformIIDProcess`, `NormalIIDProcess`, `LogNormalIIDProcess`, `NormalMixtureIIDProcess`, `TauchenAR1Process`, `RouwenhorstAR1Process`, `TauchenNormalMixtureAR1Process`), `VariableInfo.is_process`, `Variables.process_names`, and the `ProcessName` typing alias. Each `*Process` class bundles a discretization grid and a transition mechanism — instances go in `Regime(states={...})`. All of these are directly importable from `lcm` now.
- `Piece` has been renamed to `PiecewiseGridSegment` *(looking at the public surface told me this needed explanation)* and the corresponding keywords from `pieces` to `segments` (we were using that in prose, anyhow)

### Build

`hatch-vcs` writes the generated version file to `src/_lcm/version.py`.

## Migration guide for downstream code

- `from lcm import Model, Regime, AgeGrid, UniformIIDProcess, ...` — unchanged.
- `from lcm.params import MappingLeaf, as_leaf` — unchanged.
- `from lcm.typing import FloatND, ScalarInt, Period, Age, ...` — unchanged for the model-authoring aliases.
- `from lcm.exceptions import InvalidParamsError, ...` — unchanged.
- `from lcm._grids... import ...` → `from _lcm.grids... import ...`. The grid / process ABCs (`Grid`, `ContinuousGrid`, `_ContinuousStochasticProcess`) are private; import the public leaf classes from `lcm`.
- Engine internals — `lcm.engine`, `lcm.model_processing`, `lcm.regime_building`, `lcm.solution`, `lcm.simulation`, `lcm.variables`, `lcm.pandas_utils`, `lcm.state_action_space`, `lcm.dtypes`, `lcm.utils` — are now under `_lcm.*`.
- Explanation notebooks that demonstrate `_lcm` internals must `import lcm` before importing any `_lcm` submodule, so the package bootstrap completes first.

## Test plan

- [x] `pixi run -e tests-cpu tests -n 4` — 1004 passed, 10 skipped
- [x] `pixi run ty` — all checks pass
- [x] `prek run --all-files` — all hooks pass
- [x] `pixi run explanation-notebooks` — all notebooks execute
- [x] `pixi lock` — re-locked
- [x] **Verify #360 is merged before merging this PR**


src/lcm_examples/mahler_yum_2024/_model.py
```diff
@@ -15,14 +15,16 @@
 from jax import random
 from scipy.interpolate import interp1d
 
-import lcm
+from _lcm.utils.dispatchers import productmap
```
hmgaudecker, line 18: Let's not worry about this, refactor is on its way.

timmens (approved):
Very nice changes; approved from my side 🚀 

As you mentioned on Zulip, this PR cannot be properly reviewed, so I did not go through each change on GitHub. Hence, there may certainly be things that I missed; but of course they can be fixed easily with an agent afterwards.

I actually really like the changes. Given the state of the library and all of the features we (you and Max) have added over the last months, this refactoring was needed! I strongly believe this will make the codebase easier to work with for humans and agents, definitely for me.

---

Coding agents have gotten so good that if there is need, the architecture is thought through, and the names are well chose, the agent-generated code is usually pretty good. And in this case here, I think the new src-layout, and the renamings are a huge improvement. I was actually slightly confused in the beginning by the new nomenclature for the shocks, but processes is simply the better and more accurate name; well chosen! 🙂

mj023 (approved):
Good changes. Makes it much clearer to the user, where he can look for things to import, the old structure was becoming very complicated.

I'm also okay with the renamings, `Shocks` never really fit in my opinion, calling them stochastic processes is better, considering that they will often be used for income processes etc. 

Regarding commit 1e702fd86d1a78377f65b3974e5f51ce53c86271:
I was at first confused how this escaped my tests, but I just didn't realize that AOT-Compilation only activates when setting `n_subjects`. Thanks for fixing this.

PR #363, opened 2026-05-22
Rewrite the Mahler & Yum example in modern pylcm style

hmgaudecker:
## Summary

Replaces the precomputed-grid `lcm_examples/mahler_yum_2024/_model.py` with an
in-DAG model:

- **In-DAG computation** — the health-transition logit and income are computed
  inside the regime DAG, not precomputed into grids. Removes the import-time
  `tr2yp_grid` tensor and the `from _lcm.utils.dispatchers import productmap`
  private-API import.
- **Public `lcm` API only**, real `@categorical` for the 40-level `Effort`,
  labeled `pd.Series` grids, `fixed_params`/`START_PARAMS` split, and a
  `create_inputs` that returns a `(model_params, initial_conditions_DataFrame)`
  pair. Dead `ProductivityShock` categorical dropped.
- Package collapsed into `__init__.py` (the `_model.py` indirection is gone).
- Behavioral-moment tests are carried as `tests/test_mahler_yum_2024.py`; 
  the opaque pickle regression (`test_regression_mahler_yum`) and its fixtures
  are dropped. Benchmark, docs, and `generate_benchmark_data.py` updated
  to the new `create_inputs` signature.

## Notes for review

- **@mj023** — the model carries `utility_constant = 11.0` from
  the other repo's `START_PARAMS`. The previous standalone example used
  `13.108` (which looks like a paper estimate).
- The Mahler & Yum solve is GPU-scale and float32-unstable; the example tests
  are GPU-only and run at float64.

## Test plan

- [x] `test_mahler_yum_2024.py` — 29 passed on GPU at float64
- [x] `ty`, `prek` clean
- [ ] CI on the 364 branch


docs/examples/mahler_yum_2024.md
```diff
@@ -4,47 +4,37 @@ title: Mahler & Yum (2024)
...
 # Mahler & Yum (2024)
 
-Full replication of the lifecycle model from @mahler2024.
+Example implementation of the lifecycle model from @mahler2024.
 
-Two regimes (alive/dead), 8 states including health, education, productivity type, and
-health type. Three actions: labor supply, saving, and health effort. Features stochastic
-health and regime transitions, AR(1) productivity shocks, and discount-factor
+Two regimes (alive/dead), nine states including health, education, productivity type,
+and health type. Three actions: labor supply, saving, and health effort. Features
```
mj023: Should probably either only name all exogenous states (education, productivity type, health type, discount factor type) or just all states.

src/lcm_examples/mahler_yum_2024/data/init_distr_2b2t2h.txt
```diff
@@ -1,4 +1,3 @@
-Type    Health  Initial_Effort
```
mj023, line 1: I think we should keep the header, just to be 100% sure to avoid confusion in the future and because we don't really gain anything from removing them.

src/lcm_examples/mahler_yum_2024/data/surv_CL.txt
```diff
@@ -1,4 +1,3 @@
-Healthy     Unhealthy
```
mj023, line 1: Same here.

src/lcm_examples/mahler_yum_2024/data/surv_HS.txt
```diff
@@ -1,4 +1,3 @@
-Healthy     Unhealthy
```
mj023, line 1: Same here.

src/lcm_examples/mahler_yum_2024/__init__.py
```diff
@@ -1,30 +1,880 @@
...
+        "yt_sq": pd.Series({"low": -0.00250769285750586, "high": -0.00293713499239749}),
+        "wagep": pd.Series({"low": 0.17769766414897, "high": 0.144836058314823}),
+        "sigx": 0.0289408524185787,
+    },
+    "adjustment_cost": [0.000120437772838191, 0.14468204213946],
+    "discount_factor": pd.Series(
+        {"mean": 0.942749393405227, "std": 0.0283688760224992}
+    ),
+    "effort_elasticity": 1.11497911620865,
+    "utility_constant": 11.0,
```
mj023: The old parameter value was the estimate from the MY paper. Since their model contained an error, these parameters don't really fit the model specified in this file. I wanted to use their parameters to start my own estimation, but I knew that ~13 was probably too high so I changed it to 11 to get faster convergence and avoid a pretty bad local minima. 

I'm not sure if we should keep it at the 13.1079320277342 value, so that all parameters come from the paper.

mj023 (approved):
Looks correct overall, just some small changes. I am not sure with regards to the utility constant, but I would be in favor of keeping it at the value from the paper, since we are not doing estimation here.

PR #364, opened 2026-05-23
Make distributed V-arrays survive solve + simulate at production scale

hmgaudecker:
## Summary

Six coupled changes make `distributed=True` grids work end-to-end through solve and simulate at production scale (Marvin 4×A100 aca-baseline). Stacks on #361.

**Solve emits V already-sharded.** Each `max_Q_over_a` is lowered with the regime's declared V-array sharding as `out_shardings`. XLA produces V partitioned across the right devices in one go, matching what the next-period consumer's `next_regime_to_V_arr` slot was lowered against. Eliminates the runtime sharding-mismatch on grids that combine `distributed=True` and `batch_size > 0`. Regression: `test_solution_with_distributed_and_batched_grid`.

**NaN/Inf reductions stay sharded.** Module-level `v_array_has_nan` / `v_array_has_inf` are jit-decorated. The solve diagnostic accumulator, `validate_V`, `log_nan_in_V`, and the post-loop scan all route through them. GSPMD partitions the reduction per-device → all-reduce → mesh-replicated scalar; `isnan`/`isinf` fuses into the `any` reduction. The eager alternative `jnp.any(jnp.isnan(V_arr))` could fall through to gathering V onto the default device — exhausts GPU memory at production scale. Regression: two checks on 4 CPU devices that both helpers return a fully replicated 0-d scalar on a `NamedSharding`-sharded input.

**`log_level="off"` does zero per-(regime, period) V work.** The `validate_V` and `log_nan_in_V` calls in `_simulate_regime_in_period` and the period loop are gated behind `validation_enabled(logger)`. Helpers keep their "do the work; caller chose to call" signatures. `_simulate_regime_in_period` carries the logger so it gates `validate_V` before `calculate_next_states` consumes V. Regression: `test_simulate_with_log_level_off_skips_nan_value_function_check`.

**Stochastic-grid `batch_size` is honored.** `_get_stochastic_batch_sizes` reads `batch_size` from each stochastic variable's grid (both `DiscreteGrid` and `_ContinuousStochasticProcess` expose it) and passes the dict into the shock-integration productmap in `get_Q_and_F` / `get_compute_intermediates`. User-configured `batch_size>0` actually chunks the inner shock-integration loop. Regression: helper unit + solve `batch_size=0` vs `batch_size=1` identical V_arrs.

**`Model(subjects_batch_size=B)` chunks each device's subject shard.** Simulate-side `argmax_and_max_Q_over_a` builds a Q intermediate that broadcasts across `(n_subjects × actions × shocks × interpolation × next-regimes)` simultaneously — tens of TB per device at production grid sizes; XLA HLO rematerialization can't bring it inside HBM. New knob (default `0` = single-vmap path, current behaviour) routes through `chunked_map_1d` built on `jax.lax.map(..., batch_size=B)`: vmaps within a chunk, scans across chunks. JAX handles non-divisible per-device remainders. Per-iteration intermediate shrinks by `n_per_device / B`. Threaded `Model → process_regimes → _build_argmax_and_max_Q_over_a_per_period → simulation_spacemap`. Float-tolerance equivalent to the baseline but not bit-exact: XLA reduction order changes with chunk boundary. Regression: period-0 V_arr invariant across `subjects_batch_size=0` vs `=2`.

**`n_subjects` no longer needs to be a multiple of `n_devices`.** `pad_initial_conditions_for_devices` pads the leading axis up to the next multiple of `n_devices` by duplicating the last subject; pad rows simulate through their assigned regime, `trim_pad_from_raw_results` slices them off before `SimulationResult`. `generate_simulation_keys` takes `original_n_subjects` so per-subject PRNG keys split from the user's real count — pad slots replicate the last real key — keeping RNG draws for real subjects device-count-invariant (same outcomes on 1, 4, or 5 devices, modulo XLA reduction-order drift). The AOT cache keys on the padded shape (different user inputs that pad alike share one compiled program); the mismatch warning compares user-facing counts. The defensive `subject_array_sharding` divisibility check stays in place as a tripwire for code paths that bypass the padder.

## Test plan

- [x] `pixi run -e tests-cpu pytest tests -n 4` — full suite green
- [x] `pixi run -e tests-cpu pytest tests/test_distributed.py` — passes on 4 CPU devices
- [x] `pixi run -e type-checking ty` — clean
- [ ] Marvin sim run (aca-estimation, 4×A100, `log_level="debug"`, distributed assets) reaches estimation green


hmgaudecker (commented):
Autoreview.

src/_lcm/solution/solve_brute.py
```diff
@@ -338,7 +353,9 @@ def _compile_all_functions(
         logger.info("%d/%d  %s", i, n_unique, label)
         logger.info("  lowering ...")
         start = time.monotonic()
-        lowered[func_id] = jax.jit(func).lower(**lower_args)
+        lowered[func_id] = jax.jit(
```
hmgaudecker, line 360: I would have never, ever found this one myself. Manifested itself as an OOM on the 80GB GPU...

hmgaudecker:
Will start afresh from all that I learned here...

hmgaudecker:
All relevant elements are included in #368 and #370 now.

PR #365, opened 2026-05-24
Drop _jaxtyping_patch.py in favour of jaxtyping 0.3.10

### LINKED ISSUE #390: cloudpickle round-trip of `Foo[Array, "..."]` breaks the anonymous-variadic-dim sentinel
# cloudpickle round-trip of `Foo[Array, "..."]` breaks the anonymous-variadic-dim sentinel

## Summary

cloudpickling a jaxtyping type that uses the `"..."` axis (`Float[Array, "..."]` and friends) yields a type whose variadic-dim entry no longer matches the module-level `_array_types._anonymous_variadic_dim` sentinel. Any subsequent shape check on the unpickled type then trips the `assert type(variadic_dim) is _NamedVariadicDim` in `_array_types._check_dims` (`_array_types.py` L297) and raises `AssertionError`.

Plain `pickle` round-trips fine — the bug surfaces only under `cloudpickle`, which serializes the class structure (and the sentinel reference inside its `dims` tuple) by value.

## Reproducer

`jaxtyping==0.3.9`, `cloudpickle==3.1.2`, `jax==0.10.1`, Python 3.14.5.

```python
import cloudpickle
import jax.numpy as jnp
from jax import Array
from jaxtyping import Float, _array_types

sentinel = _array_types._anonymous_variadic_dim
T = Float[Array, "..."]
assert T.dims[T.index_variadic] is sentinel        # holds before pickle

loaded_T = cloudpickle.loads(cloudpickle.dumps(T))

loaded_sentinel = loaded_T.dims[loaded_T.index_variadic]
print("identity preserved:", loaded_sentinel is sentinel)      # False
print(isinstance(jnp.zeros((2, 3)), loaded_T))                  # AssertionError
```

Output:
```
identity preserved: False
AssertionError
```

(Plain `pickle.dumps(T)` / `pickle.loads(...)` works fine, because stdlib pickle resolves the type by qualified name on the receiving side and reads `_anonymous_variadic_dim` from the live module global.)

## Root cause

`_array_types.py` (L86):

```python
_anonymous_variadic_dim = object()
```

`object()` has no `__reduce__`, so cloudpickle's default reduction creates a fresh instance on unpickle. The new instance is unrelated to the live module global, breaking the `is`-identity check at L294-297:

```python
if variadic_dim is _anonymous_variadic_dim:
    ...
else:
    assert type(variadic_dim) is _NamedVariadicDim   # ← fails
```

The unpickled `variadic_dim` is a plain `object`, neither the live sentinel nor a `_NamedVariadicDim`.

## Proposed fix

Replace the bare `object()` sentinel with a `__reduce__`-backed singleton. On unpickle, `__reduce__` calls the singleton constructor, which returns the cached instance — identity holds.

```python
# _array_types.py, replacing the L86 definition

class _AnonymousVariadicDim:
    """Picklable singleton for the `"..."` axis marker."""

    _instance = None

    def __new__(cls):
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        return cls._instance

    def __reduce__(self):
        return (_AnonymousVariadicDim, ())


_anonymous_variadic_dim = _AnonymousVariadicDim()
```

With this in place, the reproducer above prints:

```
identity preserved: True
True
```

Existing call sites (`is _anonymous_variadic_dim` at L294, L351, L507) continue to work unchanged, because there's still exactly one live instance per process.

## Context

Found while debugging a beartype-claw'd JAX project that cloudpickles DAG-built functions for distribution to worker processes. Currently worked around by monkey-patching `_array_types._anonymous_variadic_dim` at import time: [`pylcm/src/_lcm/jaxtyping_patch.py`](https://github.com/OpenSourceEconomics/pylcm/blob/main/src/_lcm/jaxtyping_patch.py).

Happy to send a PR with the in-tree change above if useful!

[jaxtyping_issue_repro.py](https://github.com/user-attachments/files/28171679/jaxtyping_issue_repro.py)

hmgaudecker:
## Summary

Removes the local monkey-patch at `src/lcm/_jaxtyping_patch.py` now that jaxtyping 0.3.10 (released on PyPI) carries the upstream fix.

## Background

jaxtyping's `_array_types` module-level `object()` sentinels broke under cloudpickle round-trips because plain `object()` does not survive serialisation with identity intact. pylcm hits this through beartype-claw'd DAG-built cloudpickleable callables for the variadic-dim sentinel; the local patch worked around it.

jaxtyping 0.3.10 fixes all three affected sentinels (`_any_dtype`, `_anonymous_dim`, `_anonymous_variadic_dim`) at the source via `__reduce__`-backed singleton classes — patrick-kidger/jaxtyping#391, fixes patrick-kidger/jaxtyping#390.

## Changes

- `pyproject.toml`: `jaxtyping>=0.3.2 → jaxtyping>=0.3.10`.
- Deleted `src/lcm/_jaxtyping_patch.py`.
- Removed the import + surrounding comment from `src/lcm/__init__.py`.
- Regenerated `pixi.lock`.

## Test plan

- [x] `pixi run --environment tests-cpu tests -n 4` — **992 passed, 10 skipped** in 95s. The cloudpickle round-trip paths in the simulation tests cover what the local patch was protecting.

## Notes for downstream branches

`refactor/phase-2-api-reorganisation` (#361) renames `src/lcm/_jaxtyping_patch.py → src/_lcm/jaxtyping_patch.py` and updates the matching import. Once this PR lands, the rename collides with a delete on main; cascade resolution is to keep the delete and drop the rename + import update from the downstream branch.


hmgaudecker (commented):
Autoreview.

PR #368, opened 2026-05-26
Forward-sim memory rework: subject batching + discrete-only sharding + orbax save/load

hmgaudecker:
## Summary

Started as discrete-only sharding + orbax persistence and grew to carry the full
forward-simulation memory rework that lets the production model solve, simulate,
and save on memory-constrained GPUs. The headline new knob is
`simulate(subject_batch_size=...)`.

## Sharding (discrete-only)

- **Reject `distributed=True` on continuous grids** (`_fail_if_continuous_grid_distributed`).
  Sharding a continuous axis forces an `all-gather` of the full V-array per
  interpolation lookup; only discrete state grids shard, where cross-shard
  probability mass is contracted by a cheap `all-reduce`.
- **Solve drains every V-array shard** at the end of backward induction, so the
  simulate phase starts from concrete (still-sharded) arrays.
- **Non-device-multiple subject counts.** When any grid is distributed, the
  per-subject leading axis is padded up to the next multiple of the device count
  (pad rows duplicate the last subject and are trimmed before the result is
  returned), so `n_subjects` need not divide `n_devices`. Subject sharding is
  model-wide — every regime shares one sharding, since subjects propagate across
  regime transitions.

## `subject_batch_size` — forward-simulation memory

The V100 OOM was never the V-array; it was the **per-subject** axis (`n_subjects`,
not the grid): forward sim's `argmax_and_max_Q_over_a` and
`to_dataframe(additional_targets=...)` both build intermediates over the whole
population.

- New `simulate(subject_batch_size=...)` knob: an **outer subject-chunk loop**
  wraps the period loop; each chunk runs the full forward sim on its slice and the
  per-chunk results are concatenated on the subject axis. `None` = whole population
  in one pass.
- **Bit-identical across batch sizes.** RNG is slice-at-use: per-subject keys are
  generated for the full population and sliced by global index, so a subject's
  draws never depend on the chunk it lands in (verified for both
  `jax.random.choice` and `draw_shock`).
- **AOT compile for the chunk shape.** When `Model(n_subjects=...)` is set, the
  simulate functions are AOT-compiled for the chunk size; the ragged final chunk
  overlaps the population tail so every chunk matches the compiled program, and the
  overlap is trimmed on concatenation.
- Each chunk is offloaded to host as it completes (`raw_results` stay `jax.Array`,
  CPU-backed) when batching.
- `to_dataframe(additional_targets=...)` evaluates the target DAG in the same
  `subject_batch_size` chunks.

## `save` / `load`

`SimulationResult.save(directory=...)` writes:

- `V_arr/` and `arrays/` via **orbax** — each leaf is streamed device→host on its
  native sharding (no `all-gather`, no second device buffer; a direct save of a
  materialised array shows ~0 GiB peak-memory delta),
- `metadata.pkl` via cloudpickle (regimes, `flat_params`, ages, result metadata) —
  `flat_params` are small and unsharded, so they ride in the pickle rather than the
  orbax tree,
- `simulated_data.arrow` (feather of `to_dataframe`).

`force=True` on the orbax writes makes re-running into an existing output directory
overwrite it (pytask products, MSM loops). `load` restores every leaf onto the
sharding it was saved with.

## Supporting

- Validation self-gates on `log_level` — `"off"` does zero per-(regime, period)
  V work; NaN/Inf checks run at `"warning"`+.
- `to_dataframe` host transfer walks `period_dicts` period-by-period and pulls each
  leaf via shard iteration (sharded leaves transfer shard-by-shard, side-stepping
  the implicit `all-gather` a naive `np.asarray` on a sharded array would trigger).
- `PeriodRegimeSimulationData` is registered as a JAX pytree so the simulate-end
  drain actually completes the device compute.
- benchmarks now set `--xla_gpu_autotune_level=0`, which gives consistent gpu mem
  trackings (else they could be dominated by transient usage during compilation). 
  Very likely explains this pattern (lower mem usage and shorter compilation time
  at the expense of longer runtime):
  
  | Benchmark | Statistic | before | after | Ratio | Alert |
  |---|---|---|---|---|---|
  | aca-baseline | execution time | 15.095 s | 16.807 s | 1.11 | ❌ |
  |  | peak GPU mem | 630 MB | 581 MB | 0.92 |  |
  |  | compilation time | 284.41 s | 267.05 s | 0.94 |  |

  
  
## Verification

- pylcm CPU suite green; `tests/test_distributed.py` green on 4 CPU devices; `ty`
  and `prek` clean.
- **16 GiB V100 end-to-end**: solve → batched simulate → V-array + metadata save
  all complete with device memory steady at ~14.9 GiB, where the unbatched run
  OOM'd; `force=True` confirmed overwriting a stale checkpoint in place.


src/_lcm/grids/continuous.py
```diff
@@ -195,6 +199,9 @@ def _init_uniform_grid(
     `ScalarInt` types and only check value invariants (finiteness, ordering,
     positivity).
     """
+    _fail_if_continuous_grid_distributed(
```
hmgaudecker, line 202: I would leave this for now; we can always revert later on.

mj023: I'm fine with that, most models will have a discrete grid anyways.

src/_lcm/simulation/result_dataframe.py
```diff
@@ -122,16 +146,64 @@ def _extract_period_data(
...
+    result: dict[str, np.ndarray] = {}
+    for key in keys:
+        column = np.concatenate(host_chunks.pop(key))
+        result[key] = column[mask]
+        del column
+
+    return result
+
+
+def _to_host(value: FloatND | IntND | BoolND) -> np.ndarray:
```
mj023, line 192: Maybe you have already tried it, but how about first moving the Jax array to host with `jax.device_put()` and then converting it to a numpy array if needed?

src/lcm/result.py
```diff
@@ -205,3 +346,357 @@ def __repr__(self) -> str:
...
+    output_dir: Path,
+) -> None:
+    """Write a V_arr leaf to `output_dir`, dispatching on its sharding.
+
+    Multi-device sharded leaves go through orbax, which transfers each
+    physical shard independently and preserves the sharding spec for
+    `load`. Single-device leaves go through chunked `.npy` files
+    sliced along `spec.chunk_axis` with width `spec.chunk_size`, so
+    the per-chunk device materialisation matches the engine's
+    compute-time peak.
```
mj023: I don't really understand how this can be a problem. Whether I use batches for some dimensions or not, the resulting V_arr will always have the same size and it will have to reside on the single device at some point during the computation. Therefore I don't see how it can exeed this size during the save operation. Admittedly I was not able to look at anything myself, because of the Marvin downtime, so I am a bit out of the loop on what is actually happening in the distrtibuted code.

mj023 (commented):
I think it's good to use Orbax, it saves us some complexity for the saving and loading operations. But I think we should then maybe try to avoid introducing special cases for single devices and solely rely on Orbax.

src/_lcm/simulation/result_dataframe.py
hmgaudecker: AFAIU, that is basically the same operation, the difference being that we would end up with two arrays on host if we used `device_put()` + subsequent conversion. Since this lives only on the pandas path, we will need the NumPy intermediate in any case -- so no reason for a detour.

hmgaudecker (commented):
Autoreview.

src/lcm/result.py
hmgaudecker: Thanks, I had gone deep down into a rabbit hole there indeed!

For `V_arr` your reasoning holds exactly: it's grid-sized, fixed, and already resident on the device after solve, so writing it can't exceed what's already there. A direct orbax save of a materialised device array shows a peak-memory delta of ~0 GiB (no second buffer, no doubling). So the chunked-.npy + chunk_axis/chunk_size specs were guarding a non-problem and are deleted; the solution is now written with a plain `StandardCheckpointer().save(...).`

Turns out that the OOM was never in the save, at least not after [registering `PeriodRegimeSimulationData` as a pytree node](https://github.com/OpenSourceEconomics/pylcm/pull/368/changes#diff-1425eb15e0341de31d4a8365be8cc186d8adec1ae80ef65f6e94c09cd23b3176). It's on the per-subject axis (~250k subjects in the aca application, not the grid): forward simulation's argmax_and_max_Q_over_a and to_dataframe(additional_targets=...) both build intermediates over the whole population, and that is what overflowed the 16 GiB V100. The lever is subject_batch_size, which chunks those subject loops. On an 80 GB A100 the unbatched simulate working set is only ~20 GiB, so it's really a small-card concern.

src/_lcm/simulation/result_dataframe.py
mj023: I was just thinking that we could then avoid having to address each shard seperately, but you are right, the operation will be the same with more host memory usage.

src/_lcm/simulation/simulate.py
```diff
@@ -93,34 +109,160 @@ def simulate(
...
+            regimes=regimes,
+            regime_names_to_ids=regime_names_to_ids,
+            regime_ids_to_names=regime_ids_to_names,
+            period_to_regime_to_V_arr=period_to_regime_to_V_arr,
+            flat_params=flat_params,
+            ages=ages,
+            seed=seed,
+            logger=logger,
+        )
+        if n_overlap:
```
mj023: Having two different kinds of padding makes the code quite complicated. It would be great if we could combine both, by padding to a multiple of `batch_size` and `num_devices`. I know the user can then create huge paddings with for example `num_devices = 7` and `batch_size=100000`, but maybe we can give some kind of warning if the padding is >50%. Then he can select an appropriate batch size.

mj023 (approved):
Both features are very important and definitely needed. I am worried about the interaction of both, since we introduce two different padding strategies that we have to carry through the whole simulate loop. Furthermore I am not sure that if one uses both multiple GPU's and batches, that all GPU's will work at the same time and not just sequentially, depending on the batch size. On the other hand, why would anyone want to simulate this many subjects that you need multiple GPU's and also batches, maybe we should forbid the usage of both at the same time.

The Orbax part of the PR looks good!

src/_lcm/simulation/simulate.py
hmgaudecker: We forbid batching and sharding subjects at the same time now. If you have multiple GPUs but not enough so they can fit the number of subjects, we cannot help you.

hmgaudecker:
Note: I deleted a comment on a feature that I introduced at some point and took out again.

PR #369, opened 2026-05-29
Eager DAG-at-optimum intermediates: drain at simulate-time, serve to_dataframe from host

hmgaudecker:
Stacked on #368.

## Why

On a 16 GiB V100, `SimulationResult.save()` / `to_dataframe()` OOM'd at
`np.asarray` time with 10–11 GiB allocations even though every leaf is a
`(n_subjects,)` single-device array. Root cause: `PeriodRegimeSimulationData`
was not registered as a JAX pytree, so the simulate-end
`jax.block_until_ready(...)` drain was a silent no-op — the lazy compute graphs
for `additional_targets` only fired later, at `np.asarray`, where they needed a
multi-GiB XLA workspace.

## What changed

- **Register `PeriodRegimeSimulationData` as a pytree** so the simulate-end
  drain actually completes the device compute (907df02).
- **Evaluate every additional target eagerly at the optimum** — once per
  `(regime, period)`, `_evaluate_dag_at_optimum` composes the full target DAG
  (user functions + constraints, minus `H` and `weight_*` helpers), vmaps over
  subjects, and pulls each output straight to host. Results live in
  `PeriodRegimeSimulationData.intermediates` as plain `np.ndarray`, pinning no
  device memory across the loop.
- **`to_dataframe` serves `additional_targets` from `intermediates`** — no DAG
  recomputation, no `Regime` objects, no `flat_params` needed at dataframe time.
- **`save()` simplified**: drops the in-memory regimes unconditionally before
  `to_dataframe` (intermediates are self-contained); `arrays/` persists the
  intermediates so a loaded result serves any `additional_targets` identically
  to a fresh one.
- **Name resolution** moved from the deleted `additional_targets.py` to
  `targets.py` (`_resolve_targets`, `_collect_all_available_targets`,
  `_target_names_for_regime`).

The public contract is unchanged: `to_dataframe(additional_targets=[...] | "all")`
behaves exactly as before; the default stays lean (states + actions).

## Review follow-up (f9650f6)

Adversarial review surfaced: a save-time census that ignored host-numpy leaves
(now counted), a type-only intermediates test (now a concrete-value + shape
assertion), and stale `architecture.md` references to the deleted module (fixed).

## Verification

- pylcm suite: 1034 passed, 42 skipped
- `ty`: clean · `prek run --all-files`: clean


hmgaudecker:
This was a rabbit hole not worth pursuing.

PR #372, opened 2026-06-03
Emit solved V-arrays sharded via out_shardings (continuous-axis prerequisite)

hmgaudecker:
Stacked on #370.

Re-adds the `out_shardings` path dropped from #364: `solve` lowers each `max_Q_over_a` with the regime's declared V-array sharding as `out_shardings`, so the solved V is emitted on the layout its next-period `next_regime_to_V_arr` consumers were lowered against — no post-hoc `device_put` reshard.

For a sharded **discrete** axis, natural jit output inference already reproduces this layout (the existing `test_solution_running_on_multiple_cpus` covers it), so this is a no-op there. Its purpose is to enable a sharded **continuous** axis (e.g. `aime`): there the next-period interpolation read folds the axis into the output, so without explicit `out_shardings` the solved V returns on a layout its consumers were not lowered against. This is the prerequisite for lifting `_fail_if_continuous_grid_distributed` and sharding `aime` to scale past the 3-device `pref_type` ceiling.

Ported from #364's `3009542` (the `batch_size>0 + distributed=True` combo its original test used is now rejected at grid construction, so the existing distributed suite is the regression guard).

### DIFF
--- a/src/_lcm/solution/solve_brute.py
+++ b/src/_lcm/solution/solve_brute.py
@@ -73,12 +73,20 @@ def solve(
         }
     )
 
-    # AOT-compile all unique max_Q_over_a functions in parallel.
+    # AOT-compile all unique max_Q_over_a functions in parallel. The regime's
+    # declared V-array sharding rides along as `out_shardings` so the solved V
+    # is emitted on the layout its next-period consumers were lowered against.
     compiled_functions = _compile_all_functions(
         regimes=regimes,
         flat_params=flat_params,
         ages=ages,
         next_regime_to_V_arr=next_regime_to_V_arr,
+        regime_V_shardings=MappingProxyType(
+            {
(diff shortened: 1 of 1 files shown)


hmgaudecker:
Closing — continuous-axis sharding (the only thing this enables) doesn't pan out. The aime-4× experiment on Marvin confirmed the pathology `_fail_if_continuous_grid_distributed` documents: with out_shardings the solved V *does* carry the declared sharding (it runs on 4 GPUs), but every next-period interpolation still all-gathers the full V-array per device — XLA flags 1–1.9 TiB input/output arguments. Net: ~28.5 s per 6-regime age vs ~10 s for discrete pref_type-3× (all-reduce), i.e. ~2.8× slower. The guard stays; this prerequisite isn't worth landing.

PR #373, opened 2026-06-08
The regime grammar: phase broadcast, canonical target-granular forms, model-level slots

hmgaudecker:
The regime grammar, end to end — one review over the final state (#379 was merged into this branch; its commits are part of this diff). Three layers, each an instance of one house pattern.

**Coarse vs. granular.** A *granular* spelling names each target regime explicitly — `{"retired": law_a, "dead": law_b}` — so the engine knows exactly which target gets what. A *coarse* spelling is target-blind — a bare law, a bare `MarkovTransition` — and means "the same thing toward every (reachable) target". The canonical form the engine stores is always granular; coarse user spellings are sugar that broadcasts into it. This one rule governs state transitions, regime transitions, and (in #384) parameters.

## 1. Phase as a broadcast dimension (`Phased`)

`Phased(solve=..., simulate=...)` gives any phase-admitting regime slot per-phase variants; a bare value broadcasts to both phases.

**Carried states.** A carried state is a state that the policy does not condition on, but whose true value the simulation must track — e.g. actual wealth in private pensions when the policy was solved on a value imputed from public pensions etc. in order to keep the state space manageable. Spelled `Phased(solve=callable, simulate=Grid)` in `states`: during backward induction the callable computes the value from other variables (the state contributes no axis to the value-function grid), while in simulation it is a genuine state — seeded from initial conditions, evolved by its ordinary law of motion in `state_transitions`. Decisions are evaluated at the solve-phase imputation; every other simulate consumer reads the carried true value.

The engine carries two frozen phase namespaces (`regime.solution` / `regime.simulation`); every phase-dependent read names its phase. Everything else — constraints, actions, the `active` predicate, derived categoricals, and the regime's identity/params metadata — is phase-invariant and lives directly on the regime; `Phased` is rejected in those slots with an explanation.

## 2. Canonical target-granular forms

- `lcm.fixed_transition(state_name)` is the spelling for fixed states — an ordinary identity law, legal bare, in `Phased` sides, and in per-target cells. `None` is rejected.
- Regime transitions gain the granular form `transition={target_regime: MarkovTransition(cell)}` — **the key set declares reachability** (coarse forms keep meaning "every regime reachable"). Reachability has a single source of truth, resolved once at canonicalization; the coverage-inference and incomplete-target runtime subsystems are deleted.
- Every law of motion reaches the engine as `Mapping[target_regime, law]` over exactly the reachable targets carrying the state. Rule: the params template reads the user spec, the engine reads the canonical spec.

## 3. Model-level regime slots

`Model(functions=..., constraints=..., states=..., state_transitions=..., actions=...)` declares shared structure once. Each name (function, constraint, state, state transition, action) may be defined at the model level or at the regime level, **never both** — defining it at both raises an ambiguity error, exactly like supplying a parameter at two levels of the params dict. A regime opts out of a model-level entry by setting that name to `None` at the regime level (the *mask*); masking a state also drops its broadcast law of motion, and masking a name that has no model-level entry is an error.

A model-level (broadcast) state or action survives in a given regime only if some root computation of that regime — utility, `H`, a constraint, a derived categorical, the regime transition, or a law of motion toward a reachable target that carries the state — transitively reads it. Because "a law toward a reachable target that carries the state" refers to *other* regimes' carried states, pruning one variable in regime B can make a variable in regime A newly dead, so the pruning iterates across all regimes until nothing more can be dropped (a cross-regime fixed point). It runs separately on the solve slice and the simulate slice of each regime; a variable is dropped only when dead in **both** phases. `model.pruned_variables` records the outcome per regime. `distributed=True` is restricted to model-level states.

## Pipeline (post review-fixes)

A `UserRegime` is finalized at model build into the same class, but with slots specified at the model level merged in, broadcast variables pruned, the default `H` injected, and completeness validated — exposed as `model.user_regimes`. `process_regimes` then compiles each finalized regime into the engine `Regime`; the per-phase split (`PhasedRegimeSpec`) exists only inside that step and is never exposed. Internal signatures mark the finalized form with `FinalizedUserRegime`, an alias that the type checker treats as plain `UserRegime` — it enforces nothing and only documents that finalization has happened.

## Sanctioned breaks

- `state_transitions={"s": None}` dies loudly (use `fixed_transition`).
- Completeness errors raise at `Model(...)` instead of `Regime(...)` (regime-level declarations may be completed by model-level slots).

## Performance

Full GPU benchmark matrix run on both layers: execution time and GPU peak memory flat everywhere (≤1.06 / 1.00). Two alerts, both understood: aca-baseline compile +11 % and peak host RAM +63 % (the granular representation traces one probability cell per declared target instead of one callable per regime — one-time build cost; cell-dedup is the follow-up if it ever matters), and Precautionary-Savings simulate 1.13× — traced to a pre-existing per-call vmap re-trace in the membership draw and fixed in #384. A100 cross-check: seeded 1000×45 ACA benchmark panels bit-identical across the PR except ≤3.6e-13 relative in the recorded `value` (E[V] float-sum regrouping; argmaxes unmoved).


src/_lcm/regime_building/processing.py
```diff
@@ -99,23 +111,38 @@ def process_regimes(
...
-    for regime_name, user_regime in user_regimes.items():
-        nested_transitions[regime_name] = _extract_transitions_from_regime(
-            user_regime=user_regime,
-            states_per_regime=states_per_regime,
+    # Each phase's transitions come from its own slice against its own state
+    # sets: the simulate slice additionally holds every carried-only state and
+    # its law of motion, so the extraction registers the law for each
+    # reachable target that carries the state — including targets reached
+    # through nothing but the carried state.
+    nested_transitions = {
```
hmgaudecker:
```suggestion
    solve_nested_transitions = {
```
Make solve/simulate symmetric here and elsewhere by including prefixes on both legs.

src/_lcm/regime_building/phases.py
```diff
@@ -0,0 +1,404 @@
...
+
+    Raises:
+        RegimeInitializationError: If any slot value violates the phase
+            grammar.
+
+    """
+    errors: list[str] = []
+
+    solve_functions, simulate_functions = _split_functions(user_regime, errors)
+    solve_grid_states, simulate_grid_states, carried_imputations = _split_states(
```
hmgaudecker: Strict conventions in the project are
1. Pure functions (even appending to errors is a no-go)
2. As soon as a function takes more than one argument, it must be keyword-only.

Applies to the entire module and probably elsewhere.

src/_lcm/regime_building/phases.py
```diff
@@ -0,0 +1,404 @@
...
+            simulate_functions,
+            simulate_grid_states,
+            simulate_state_transitions,
+            simulate_transition,
+        ),
+    )
+
+
+@dataclass(frozen=True, kw_only=True)
+class PhasedRegimeSpec:
```
hmgaudecker, line 164: Looking ahead to #379, we'll have:

- `UserRegime`
- `EffectiveUserRegime`
- `PhasedRegimeSpec` 
- `Regime`

It seems to me like this is too much. Can't we combine `EffectiveUserRegime` and `PhasedRegimeSpec`?

src/_lcm/simulation/result_metadata.py
```diff
@@ -104,10 +103,10 @@ def _compute_metadata(
     regime_to_actions: dict[RegimeName, tuple[ActionName, ...]] = {}
 
     for regime_name, regime in regimes.items():
-        regime_to_states[regime_name] = regime.variables.state_names
-        regime_to_actions[regime_name] = regime.variables.action_names
-        all_states.update(regime.variables.state_names)
-        all_actions.update(regime.variables.action_names)
+        regime_to_states[regime_name] = regime.simulation.state_names
+        regime_to_actions[regime_name] = regime.simulation.variables.action_names
```
hmgaudecker: Why is `state_names` now a property of `simulation` and `action_names` a property of `simulation.variables`? Can we get rid of `variables`?

src/_lcm/engine.py
```diff
@@ -308,10 +321,133 @@ class SolveFunctions:
     host-visible memory.
     """
 
+    resolved_fixed_params: FlatRegimeParams = MappingProxyType({})
```
hmgaudecker, line 328: Isn't the `resolved_` superfluous now that it clear this has been processed?

src/_lcm/model_processing.py
```diff
@@ -418,8 +421,8 @@ def _partial_fixed_params_into_regimes(
             ),
         )
 
-        # Build new simulate_functions with partialled functions
-        simulate_funcs = regime.simulate_functions
+        # Build new simulation phase with partialled functions
+        simulate_funcs = regime.simulation
```
hmgaudecker: Weird that `regime.simulation` becomes `simulate_funcs`

src/_lcm/variables.py
```diff
@@ -41,6 +42,54 @@ def _bind_forward_refs(*, regime_cls: type) -> None:
     UserRegime = regime_cls  # ty: ignore[invalid-assignment]
 
 
+def _grid_states(user_regime: UserRegime) -> dict[StateName, Grid]:
```
hmgaudecker: Remember the deep module convention.

hmgaudecker (commented):
Autoreview.

src/lcm/phased.py
```diff
@@ -0,0 +1,38 @@
...
+A thin leaf module — the class definition only, with no dependency on
+`Regime`, the validators, or the regime-building code, so the user-facing
+`Regime`, the engine-internal normalizer, and the regime validators can all
+import it without an import cycle.
+"""
+
+from lcm.exceptions import RegimeInitializationError
+
+
+class Phased[S, T]:
```
mj023, line 12: I think this name carries too little information what `Phased` actually is. Maybe we could choose something more explicit, I couldn't come up with something better tough.

mj023 (approved):
Im fine with these changes, they seem to mostly increase the usability (model scope functions, coarse vs granular). The split with different model functions for solve and simulate is mostly for the Beta-Delta preferences, right?

src/lcm/phased.py
hmgaudecker: I'd hope that it'll become clear from the docs what we mean with phases... Might have to sprinkle it in more.

hmgaudecker:
> The split with different model functions for solve and simulate is mostly for the Beta-Delta preferences, right?

Functions yes, we've already had that special case for a while (with an awkward interface). Then I needed the same thing for states (would love to have one more, but not quite worth it -- so imputing it in the solution phase, but updating the initial conditions with the actual rule in the simulation phase). So we got rid of special cases here and allow it for all slots where we may need it.

PR #374, opened 2026-06-10
Promote the Iskhakov et al. (2017) model to an example with docs

hmgaudecker:
Promotes the deterministic retirement model behind the analytical-solution tests to a proper example (`lcm_examples.iskhakov_et_al_2017`), with an example page and an explanation notebook plotting the paper's signature saw-tooth consumption function — each tooth is one optimal retirement age: as people retire later, their lifetime wealth increases, changing the optimal consumption path. The notebook's closing section is reserved for the brute-force vs DC-EGM comparison once the solver lands; this PR is the first piece of that stack and is independent of #373.

`tests/test_models/deterministic/base.py` becomes a re-export shim, so the test suite keeps its import location.


hmgaudecker (commented):
Autoreview.

mj023 (approved):

PR #377, opened 2026-06-10
Add solver-config classes, builder registry, and DC-EGM build-time validation

hmgaudecker:
Second DC-EGM prequel (stacked on #374; see also #375, #376): the public solver abstraction and the full DC-EGM build-time validation, with the solver itself still to come.

- `lcm.solvers`: `BruteForce` (default) and `DCEGM` — pure-config frozen dataclasses, no engine internals. `Regime` gains a `solver` field (`default_factory=BruteForce`, fully backward compatible).
- `_lcm/solution/registry.py`: `SOLVER_KERNEL_BUILDERS` maps `type(regime.solver)` to the per-period kernel builder; `process_regimes` dispatches through it (the brute-force builder is the previous `_build_max_Q_over_a_per_period`, moved verbatim). Adding a solver = one public config class + one registered builder, no `isinstance` chains.
- `_lcm/egm/validation.py`: the DC-EGM model contract, enforced at `Model` construction — state/action classification (exactly one Euler state + one continuous action; process states exempt; passive continuous states rejected pointing at the upcoming extension), required functions (`resources`, post-decision, `inverse_marginal_utility`), DAG-ancestor rules (transition reaches wealth/consumption only through the post-decision function; utility independent of the state; constraints discrete-only; savings-node-stage functions independent of the decision), default `H`, deterministic Euler transition, grid hygiene, target-regime compatibility, and numeric spot checks (`post_decision ≈ R − c`, `(u′)⁻¹(u′(c)) ≈ c`, monotonicity).
- A `DCEGM`-configured model constructs and validates, but `solve()`/`simulate()` raise a clear `NotImplementedError` until the EGM step lands.

Tests were written first: 14 spec tests covering one violation per contract rule, a non-DCEGM-target rule, and a brute-force regression (explicit `solver=BruteForce()` solves identically to the default).

Merge-integration notes (deliberate, to resolve when the siblings land):
- #375 extends `get_max_Q_over_a` with taste-shock kwargs; the registry's brute-force builder call site needs those threaded after the merge.
- #376 also creates `src/_lcm/egm/__init__.py` (trivial union).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #376, opened 2026-06-10
Add the FUES upper-envelope kernel and padded-grid interpolation

hmgaudecker:
Second solver-stack prequel for DC-EGM (with #375): a native JAX reimplementation of the Fast Upper-Envelope Scan (Dobrescu & Shanker 2022, SSRN 4181302) plus tie-safe linear interpolation on NaN-padded envelope rows — pure functions with no new dependency, unused until the DC-EGM step wires them in. Tests were written first: analytic crossing-segments envelopes (duplicated kink abscissae with left/right policies, dominated-point deletion), passthrough invariants, vmap consistency, and overflow signaling via `n_kept`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #375, opened 2026-06-10
Add regime-level EV1 taste shocks (logsumexp solve, Gumbel-max simulation)

hmgaudecker:
Closes #247. First solver-stack prequel for DC-EGM: taste shocks are a model property (`Regime(taste_shocks=ExtremeValueTasteShocks())`), so a smoothed model solves to the same value function under any solver — this gives the upcoming DC-EGM solver an in-framework λ>0 oracle. Solve aggregates discrete actions via `scale·logsumexp(Qc/scale)` (the shared `_lcm/logsum.py` kernel, reused by DC-EGM later); simulation draws the discrete action by Gumbel-max, exactly logit-consistent with the solve. Tests were written first and include an exact numpy reference for the solve and a simulated-frequency-vs-softmax convergence check.

Also adds `benchmarks/bench_iskhakov_et_al_2017.py`: brute-force solve and simulate asv baselines for the Iskhakov et al. (2017) retirement model with EV1 taste shocks (solve on a dense 1000×5000 wealth×consumption grid, simulate 100k subjects via the Gumbel-max path, plus GPU peak-memory companions). This is the model the DC-EGM solver will be benchmarked on; the `solver=DCEGM(...)` classes join the same file once the solver lands, so the asv PR comment will show the head-to-head.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #378, opened 2026-06-10
DC-EGM solver abstraction, build-time validation, and the concave EGM step

hmgaudecker:
Third DC-EGM stack PR (stacked on #376; sibling prequels #374, #375): the public solver abstraction, the full DC-EGM build-time validation, and the concave EGM step. Solving a `solver=DCEGM(...)` regime works end-to-end for the concave case (no discrete actions, no taste shocks); forward simulation still raises `NotImplementedError`. (Absorbs the former solver-config/validation PR #377 — one reviewable unit, since the validation rules are best read against the kernel that relies on them.)

Solver abstraction and validation:

- `lcm.solvers`: `BruteForce` (default) and `DCEGM` — pure-config frozen dataclasses, no engine internals. `Regime` gains a `solver` field (`default_factory=BruteForce`, fully backward compatible).
- `_lcm/solution/registry.py`: `SOLVER_KERNEL_BUILDERS` maps `type(regime.solver)` to the per-period kernel builder; `process_regimes` dispatches through it. Adding a solver = one public config class + one registered builder, no `isinstance` chains.
- `_lcm/egm/validation.py`: the DC-EGM model contract, enforced at `Model` construction — state/action classification (exactly one Euler state + one continuous action; process states exempt), required functions (`resources`, post-decision, `inverse_marginal_utility`), DAG-ancestor rules (transition reaches wealth/consumption only through the post-decision function; utility independent of the state; constraints discrete-only; savings-node-stage functions independent of the decision), default `H`, deterministic Euler transition, grid hygiene (no batched/distributed/runtime kernel grids), target-regime compatibility, and numeric spot checks (`post_decision ≈ R − c`, `(u′)⁻¹(u′(c)) ≈ c`, monotonicity).

The concave EGM step:

- `_lcm/egm/euler.py` — Euler inversion per savings node with the degenerate-inversion guard (expected marginal continuation ε-clamped after discounting, so all-mass-on-stateless-terminal children and `discount_factor=0` yield the consume-everything corner instead of an infinite endogenous point).
- `_lcm/egm/carry.py` — `EgmCarry` (registered pytree dataclass): NaN-padded weakly-ascending R-space rows of endogenous grid, policy, value, and marginal utility (exactly 0.0 on infeasible rows), plus a 0-d taste-shock scale (0.0 in this PR).
- `_lcm/egm/step.py` — the per-(period, regime) kernel: child states at each savings node via the existing next-state/stochastic-weights builders with the post-decision function removed from the DAG (savings enters as an external input); child carry interpolated in the child's R′-space through a closed-over child resources function; the Euler expectation multiplies by the composed gradient ∂R′/∂A = grad of A ↦ R′(𝒯(A)) — never a hard-coded ∂M′/∂A; unreachable-target contributions zeroed `jnp.where`-on-results; closed-form credit-constrained candidates concatenated into the FUES scan; publish takes max(interpolated envelope, closed-form constrained value) on the regime's exogenous grid — exact wherever the constraint binds. Target/transition extraction is isolated in one adapter (`get_egm_continuation_targets` + `_build_target_closures`).
- `_lcm/egm/terminal.py` — terminal carry producers: wealth-grid terminals carry terminal utility and its wealth gradient (identity R), evaluated with the solve loop's `period`/`age` so age-dependent bequest utilities work; stateless terminals carry constant-value zero-μ rows.
- `_lcm/egm/interp.py` — padded-grid interpolation with tie-safe duplicated-kink brackets and a −inf-safe endpoint blend (a dead node forces its bracket interior to −inf instead of NaN).
- Driver: `SolveFunctions` gains `egm_step` / `egm_carry_producer` / `egm_carry_template`; `solve()` maintains a rolling `next_regime_to_egm_carry` next to the V mapping (template key order, JIT-stable pytree); AOT lowering handles per-solver arguments. Envelope overflow NaN-poisons the outputs so the existing NaN diagnostics surface the offending (regime, period).
- Out-of-scope configurations build kernels that raise a precise `NotImplementedError` at solve time — `Model` construction always succeeds for a validated regime.

Tests were written first: one spec test per validation contract rule, a brute-force regression (explicit `solver=BruteForce()` solves identically to the default), and the headline kernel spec asserting DC-EGM matches the analytic Iskhakov et al. (2017) retired-life solution on the **full** wealth grid — including the low-wealth nodes the brute-force oracle test must exclude — at tighter tolerance; plus an EGM-vs-brute non-inferiority diagnostic, closed-form corner cases (zero discount factor, age-dependent bequests), NaN-param surfacing, a V-layout test, and simulate-raises.


hmgaudecker:

PR #379, opened 2026-06-10
Canonical target-granular forms, per-target regime transitions, model-level broadcast

hmgaudecker:
Stacked on #373. Pushes "canonical = granular" to its fixed point: one PR, three commits.

## Commit 1 — the Spine

The engine reads one uniform, target-granular representation of every law of motion; the user pipeline gains an explicit effective-regime layer; fixed states get a first-class spelling.

- **`lcm.fixed_transition(state_name)`** replaces `None` in `state_transitions`: an ordinary identity law — legal bare, inside `Phased` sides, and in per-target cells — whose argument must match its dict key (checked, including inside `Phased`/per-target leaves). `None` is rejected with a migration hint.
- **`EffectiveUserRegime`**: the regime as the model runs it — model-level `derived_categoricals` merged, default `H` injected, completeness validated (utility entry, state-transition coverage, state/action overlap, distributed-grid rules) in its constructor. A bare `Regime` validates only local value-shape properties; completeness relocates to model build, the seam that lets model-level slots satisfy it later (commit 3). `model.user_regimes` exposes the effective form.
- **`canonicalize_regimes`**: the model-level canonicalization stage. Every phase slice's `state_transitions` value becomes the canonical `Mapping[RegimeName, law]` over exactly the reachable targets carrying the state — bare laws broadcast, per-target dicts pass through, `fixed_transition` desugars into grid-annotated identities. Reachability is resolved once, here; `_extract_phase_transitions` becomes a pure transpose and `_classify_transitions` is deleted. Rule: **the params template reads the user (effective) spec, the engine reads the canonical spec.**

Behavior-preserving except two sanctioned breaks: the `None` spelling dies loudly, and completeness errors raise at `Model(...)` instead of `Regime(...)`.

## Commit 2 — per-target regime transitions

```python
transition={"retired": MarkovTransition(prob_retired), "dead": MarkovTransition(prob_dead)}
```

- The key set declares reachability: omitted regimes are structurally unreachable. Coarse forms (bare callable / `MarkovTransition`) keep meaning "every regime reachable"; terminal stays `transition=None`; `transition={}` errors; cells must be `MarkovTransition`-wrapped (plain callables rejected as "not yet supported").
- Reachability now has a **single source of truth** — the regime transition — resolved in `canonicalize_regimes`. The coverage-based reachability inference and the incomplete-target runtime subsystem (`_validate_no_reachable_incomplete_targets`, the validate_V hint) are deleted; `get_complete_targets` becomes `get_period_targets`.
- Cell params surface in the template under `to_<target>_next_regime`; `Phased` granular transitions require matching forms and identical key sets across phases. The simulation membership draw intersects active targets with the declared key set.

## Commit 3 — model-level broadcast + pruning + sharding

`Model(functions=..., constraints=..., states=..., state_transitions=..., actions=...)`, each slot accepting exactly what the regime-level slot accepts (incl. `Phased`, processes, per-target dicts, `fixed_transition`).

- **Exactly-one-level rule** (params-style): a name lives at model level or regime level, never both — ambiguity errors. A regime-level `None` masks the model entry uniformly across slots; masking a state also drops its broadcast law; an unbound mask errors (with the `fixed_transition` hint for `state_transitions`). Broadcast laws are not merged into terminal regimes.
- **DAG pruning**: broadcast states and actions survive per regime only where a root computation (utility, `H`, constraints, derived categoricals, the regime transition, or a law toward a reachable target that keeps the state) transitively reads them — a cross-regime least fixed point, computed per phase slice, pruned only when dead in both phases. Regime-level declarations are never pruned. `model.pruned_variables` reports the result; `validate_model_inputs` exempts broadcast names from the unused-variable check.
- **Sharding**: `distributed=True` is legal only on model-level states; a sharded state pruned from a non-terminal regime errors (unshard, or make the regime use it). `batch_size` stays per-declaration.

## Verification

- pylcm: full suite green serially after each commit (1134 passed); prek + ty clean.
- aca-model migrated on `feat/canonical-broadcast-grammar` (one mirror commit per pylcm commit): `pref_type` via `fixed_transition`; granular regime transitions declaring each regime's true target set; shared functions/constraint/states/laws broadcast from `Model(...)` with `dead` masking what the bequest DAG does not read — 236 tests green.
- GPU regression (seeded panel diff on an A100, expectation bit-identical) to follow before merge.


hmgaudecker (commented):
Review start, continuing later.

PR #380, opened 2026-06-10
DC-EGM discrete choices: per-combo kernel, EV1 logsum aggregation, dcegm cross-check

hmgaudecker:
Fourth DC-EGM stack PR (stacked on #378; carries #375's taste shocks in its base): discrete states, discrete actions, and EV1 taste shocks in the EGM step. With this, DC-EGM solves the full Iskhakov–Jørgensen–Rust–Schjerning (2017) model — kinked (λ=0) and smoothed (λ>0) — and matches three independent oracles.

What lands:

- The EGM kernel is vmapped over the regime's own discrete combos; per-combo Euler inversion + FUES, then the own discrete-action axes of the published V are aggregated by the EV1 logsum with the regime's λ (`taste_shocks__scale` flat param) or the hard max — discrete-state axes remain V axes, byte-compatible with the brute-force layout.
- Child side, at each savings node: the composed resources map returns the child's discrete-state codes as `value_and_grad` auxiliaries (their transitions are savings-independent by validation rule 9, so the ∂R′/∂A gradient is untouched); the codes integer-index the child carry's leading state axes; the remaining child discrete-action rows are interpolated at R′ and aggregated by `logsum_and_softmax` with the child's carry-stored λ — smoothed value = logsum, smoothed marginal = Σ P·μ (exact for EV1 by Danskin). λ=0 flows through the same kernel's hard-max branch; all-−inf (infeasible) rows carry zero probability and exactly zero marginal utility.
- Discrete-only constraints are fully supported: infeasible combos become −inf value rows with μ = 0 (`SolverBuildContext` gains a `constraints` field as the feasibility source).
- `EgmCarry` rows gain leading discrete dims (states then actions, axis order single-sourced from the V layout); the child's λ travels to parents as carry data.
- The unsupported-feature filter is relaxed exactly as far as the implementation goes; process states and stochastic carry-target transitions keep precise `NotImplementedError`s.
- Benchmarks: `IskhakovEtAl2017DCEGMSolve` (+ GPU peak-mem companion) joins `bench_iskhakov_et_al_2017.py`, solving the same taste-shock model as the brute-force baseline classes with `solver=DCEGM(...)` — the asv PR comment now shows the head-to-head.

Tests were written first (red at the spec commit with the kernel's unsupported-feature `NotImplementedError`):

- Full-IJRS λ=0 model vs the in-repo analytic solution (MSE at `decimal=2`, tighter than the brute test).
- λ>0: brute-with-taste-shocks vs DC-EGM agreement on V (solver-independent semantics of `Regime.taste_shocks`).
- Cross-check against fixtures generated by the `OpenSourceEconomics/dcegm` package (pinned commit in the fixture README). **Note:** 7 of 108 fixture rows (retiree wealth = 1.0, periods 0–6) are excluded with documentation — the stored values contradict the fixture's own policy column (e.g. period 0: stored −66.0, while consuming the stored policy is worth ≈ −54), and an independent fine-grid recursion sides against them (truth ≈ −53.24; pylcm brute −53.6). 101/108 rows assert at full tolerance.
- New unit specs for the discrete-choice kernel (child-state indexing, feasibility, λ-flow).

Verified independently of the implementing run: full suite 1172 passed / 46 skipped, ty clean, prek clean.


hmgaudecker:

PR #381, opened 2026-06-10
DC-EGM passive continuous states + generalized-resources stress tests

hmgaudecker:
Sixth DC-EGM stack PR (stacked on #380, sibling of the docs PR): passive continuous states and the generalized-resources stress tests — the ACA-enablement machinery.

What lands:

- **Passive continuous states** (e.g. AIME: deterministic transition whose ancestors include neither the continuous action, the post-decision function, nor the Euler state) are supported in DC-EGM regimes. They become carry/V axes alongside the discrete dims (canonical order single-sourced: discrete states → passive states → discrete actions → grid axis); FUES still runs 1-D per combo. Child-side, the passive value lands off-grid, so the carry is read **mixed**: every row interpolated 1-D at the child's R′, then each passive axis reduced by edge-clamped linear weights (`locate_on_grid`), and only then does the EV1 logsum aggregate the action rows — blending after the logsum would interpolate the aggregate, which is wrong wherever the CCPs differ between neighboring passive nodes. On-node reads are exact by construction of the weights; −inf neighbors blend via `jnp.where`-on-results (no `0·(−inf)` NaNs). The child's passive values ride as `value_and_grad` auxiliaries of the composed resources map — constant in savings by passivity, so the ∂R′/∂A gradient is untouched.
- **Validation rule 1 flips**: passivity is now *verified* (forbidden-ancestor checks; stochastic passive transitions and batched/distributed passive grids rejected with precise messages) instead of passive states being rejected wholesale.
- **Generalized-resources stress tests**: a means-tested model (`R = max(wealth, floor)` — flat region, ∂R/∂M = 0) vs dense brute force on the full grid; the consume-everything corner (all child mass in the flat region → expected marginal exactly zero → the degenerate-inversion guard yields the closed-form constrained solution, asserted at 1e-6); and a FUES unit case where the flat-resources continuation kink's two branches cross (dominated points dropped, kink inserted twice with the left/right policies, analytic envelope). **Honest finding: all three passed before any code change** — the composed-gradient machinery from the concave-step PR already covers generalized resources, exactly as planned; they are committed as regression locks (stated in the red commit).

Tests first throughout: the passive specs were red at `Model` construction with the old rule-1 error. The exactness spec is the sharp one: with an identity passive transition, every skill slice must reproduce an independently parameterized skill-free model at rtol/atol 1e-9 — the mixed read may introduce no interpolation where none is needed.

Verified: full suite 1179 passed / 46 skipped (independently re-run), ty clean, prek clean.

Follow-ups (out of scope here): the ACA-model acceptance run (separate repo); child resources functions reading the child's passive/discrete states (precise `NotImplementedError` today — ACA's full budget needs it); process states in DC-EGM regimes.


hmgaudecker:
Extended with the recorded kernel follow-ups (reviewed + independently verified, full suite 1204 passed / 46 skipped, ty + prek clean):

- **Child resources reading child discrete actions / discrete states / passive states** — per-row R′ queries and per-row composed gradients (`Σ P·μ·(∂R′/∂A)` per choice, Danskin per choice's own R-space); the identity-resources fast path is preserved.
- **Process states in DC-EGM regimes** — node-valued carry/V axes on the own side; child-side weighted expectation over child process nodes using the existing intrinsic-weight machinery (runtime `mu/sigma/rho` params flow exactly as in Q_and_F), placed **outside** the action aggregation to match brute force. Rouwenhorst + Gauss-Hermite IID specs vs dense brute oracles.
- **Declared reachability in the DC-EGM validation** — `_reachable_target_names` reads the regime transition (granular mapping → its key set; coarse → all regimes), aligning rule 11 with the canonical-broadcast-grammar semantics.

Still rejected with precise messages: stochastic non-process transitions into carry targets, child-resources params beyond the state/action set, runtime-param process grids feeding resources.

hmgaudecker:
**aca-side refactor result for the budget chains (gap #1)** — grounded in `aca-model@feat/dcegm-solver`, re-verified against `ddb6b98d` (same two chains still reported; the new kernel follow-ups don't touch this, as planned):

The "fold action-independent terms into generalized R" move is half right: `oop_costs` is foldable, `pension_assets_adjustment` is not — and the fold would cost us the borrowing-constraint encoding.

1. **`oop_costs` is structurally legal in `R`.** Its full chain (`total_health_costs` ← hcc states; `is_medicaid_eligible` ← `countable_income` ← `capital_income(assets)`; `buy_private` is a discrete action) reads the Euler state only at decision time and never the continuous action.

2. **But the fold changes the feasible set.** The model's true constraint is `consumption ≤ max(cash_on_hand, floor)` — OOP (and the pension adjustment) hit *next-period assets*, deliberately not current resources (the agent does not condition consumption on the HCC realization; next assets can go negative, the assets grid floor is `-max_annual_labor_income` for that reason). Under the current spelling, `savings >= 0` with `R = max(coh, floor)` encodes that constraint *exactly*. Under the folded spelling, `consumption <= R - s_min` would need `s_min = -oop + adj` — a state-dependent grid lower bound, which a fixed savings grid cannot express. Not semantics-preserving for this model.

3. **`pension_assets_adjustment` cannot fold at all, for two independent reasons:**
   - It is **target-dependent**: `imputed_pension_wealth_next_period` indexes its imputation tables by `target_his` and consumes `next_aime` — a transition-DAG output. `R` is a single per-regime decision-time function; there is no target axis to put this on. (It is also phase-variant: `Phased(solve=assets_adjustment, simulate=zero)` — the correction exists only because solve's continuation uses imputed pension wealth.)
   - The `dead` target must *not* receive it (no future imputation to reconcile; `next_assets_when_dead` omits it today). A shared `R` cannot carry a per-target asymmetry.

**Conclusion:** the residual is real and it is the whole adjustment term, not a timing edge case. What the ACA model needs from the contract: *additive assets-law terms that are constant in the continuous action*, composed into the carry as per-node (and per-target) constants — the law shape is `next_assets = savings + g(state, discrete actions, target)` with `dg/dc = 0`, so the Euler inversion is untouched; only the carry's resource argument shifts per node/target. The aca side keeps `R = max(coh, floor)` + `savings >= 0` (exact constraint encoding) and the per-target laws `savings + adj - oop` / `savings - oop` (dead: no adj).

Minimal failing fixture: `aca-model@feat/dcegm-solver`, `tests/test_dcegm_model_creation.py::test_dcegm_benchmark_model_builds` (non-strict xfail quoting both chains).

hmgaudecker:
## Execution plan: admit the ACA budget chains (gap #1) — additive action-constant assets-law terms

Closes the last build blocker for `aca-model@feat/dcegm-solver` (minimal failing fixture: `tests/test_dcegm_model_creation.py::test_dcegm_benchmark_model_builds`, non-strict xfail quoting both chains). Design basis: the budget-chain analysis above — the R-fold is rejected (it breaks the exact `savings >= 0` ⇔ `consumption <= max(coh, floor)` constraint encoding and cannot express per-target asymmetry); what the contract admits instead is the law shape

```
next_assets = savings + g(...)        with  ∂g/∂c = 0
```

per target regime, where `g` may read own discrete states, discrete actions, process nodes, params, `period`/`age`, the target regime — and, in the ACA chains, the current Euler state through decision-time functions (`capital_income(assets)` feeding `oop_costs` and `pension_assets_adjustment`).

### Step 0 — settle the one open numerical point: where is `g`'s Euler-state dependence evaluated?

Everything else is mechanical; this is not. At a savings node, current `assets` is *endogenous* (recovered only after the Euler inversion), so `g(assets, …)` is not a known constant at child-read time. Decide and document one of:

- **(a) Exact, two-pass:** evaluate the child read with `g` expressed as a function of the endogenous point — i.e. compose `child_euler_state(savings) = savings + g(assets(savings))` where `assets(savings)` comes from the same-row inversion; the composed gradient `∂(next_assets)/∂A` picks up `g'·(∂assets/∂A)` and the per-row machinery (`_compute_row_queries_and_gradients`) already carries row-wise gradients, so this is a sequencing change, not new math. Check whether the existing row structure lets the inversion and the read be interleaved per row without a global fixed point.
- **(b) Exogenous-row evaluation:** evaluate `g` on the own asset-grid rows the carry already indexes, making it a per-row constant by construction. Exactness against brute force must be argued, not assumed — brute force evaluates `g` at the *exact* current grid point, so (b) may actually be exact whenever the EGM step is grid-row-indexed in current assets; verify against the kernel's actual row semantics.

Whichever you pick, write the decision into `egm/step.py`'s module docstring with the brute-force-equivalence argument. If neither is exactly brute-force-equivalent, stop and surface — do not ship an approximation silently.

### Step 1 — red tests first (small oracle models, brute force as the spec)

In `tests/` (pylcm), one minimal model per property, each solved by brute force and DC-EGM and compared at tight tolerance:

1. `g` reading a discrete state and a discrete action (`next_assets = savings + b*health - oop(buy_private)`), single target — the base case.
2. `g` reading the Euler state through a decision-time function (`capital_income = r*assets`, threshold inside `g` à la Medicaid eligibility) — pins Step 0's choice, including a kink.
3. Per-target asymmetry: two targets, `g` applied toward one, omitted toward the other (`dead` receives `savings` only).
4. Phase-variant `g`: `Phased(solve=adjustment, simulate=zero)` — the correction exists only because solve's continuation uses imputed values; simulate must apply the zero variant.
5. Target-dependent inputs: `g` consuming a transition-DAG output (`next_aime`) and target-indexed tables (`target_his`-style), mirroring `imputed_pension_wealth_next_period`.

All red against the current validator (they die in `_fail_if_euler_transition_bypasses_post_decision`).

### Step 2 — contract + validator

`egm/validation.py::_fail_if_euler_transition_bypasses_post_decision`: instead of rejecting any transition whose ancestors reach `assets`/`consumption_dollars` outside the post-decision function, *decompose*: with `savings` opaque, require (i) the transition consumes `savings`, (ii) `∂(transition)/∂savings ≡ 1` is not checkable symbolically — instead require the residual `g := transition − savings` (build it as a DAG by subtracting the opaque leaf) has **no path to the continuous action** (`consumption_dollars` and `resources` stay forbidden ancestors of `g`); the Euler state is allowed in `g` per Step 0's semantics. Keep precise messages: name the offending chain when the continuous action is reachable. New checks: `g` toward each target validated independently (per-target dicts), the `Phased` variants validated per phase.

### Step 3 — kernel wiring

`egm/step.py`: the child read (`_get_child_carry_reader` / `_ChildRead`) already evaluates the full next-state function at the savings node and the per-row queries/gradients exist. Wire `g` per Step 0: shift `child_euler_state` by the term, route its gradient contribution through the composed gradients, keep the identity-resources fast path for regimes where `g ≡ 0` (no cost when unused). Per-target `g` rides on the existing per-target read structure (`get_egm_continuation_targets`); the dead-target omission falls out of `g` simply not being declared toward that target.

### Step 4 — green, then the acceptance flip

- pylcm suite + the new oracle tests green; ty + prek clean per commit.
- Run `aca-model@feat/dcegm-solver` against the tip: `test_dcegm_benchmark_model_builds` must XPASS. Flip the xfail there (that commit lands in aca-model; ping and it gets done from this side) and run the full aca-model DC-EGM acceptance: build + small-grid solve parity vs brute force.

### Constraints

- No aca-model workarounds — if a chain turns out not to fit the contract above, surface it on this thread; the spec on the aca side is `R = max(coh, floor)`, `savings >= 0`, per-target laws `savings + adj − oop` / `savings − oop` (dead: no `adj`), and that spelling stays.
- Benchmark note while you're in there: the stack's own CI shows aca-baseline compile 1.35× on this PR — worth a look once gap #1 is in, since the per-row machinery is the likely contributor.

hmgaudecker:
**Follow-up after the additive-g landing (verified against `eb1f77e`): the next ACA blocker is the savings-stage rule.**

With the additive action-constant contract in, `test_dcegm_benchmark_model_builds` now fails one validator later: `_fail_if_savings_stage_function_depends_on_decision` rejects the ACA **regime transition probabilities**, which read `assets` through `is_medicaid_eligible ← countable_income ← capital_income(assets)` (plus the assets test itself).

Proposed extension, same semantics you just shipped for `g`: with the per-exogenous-asset-node evaluation, current `assets` *is* known at the node — admit savings-stage functions (regime-transition probs, stochastic weights, non-Euler laws) that depend on the Euler state, evaluated per node, **under a smoothness assumption**: the aca side is replacing the eligibility indicator with a compactly supported smoothstep (C², ~$100 band around the statutory threshold, dedicated grid nodes across the band), so the admitted dependence is smooth at node resolution by construction. Cliffs in savings-stage functions stay rejected — the kernel cannot find discontinuities in opaque DAGs, and the contract shouldn't pretend it can.

Failing fixture unchanged: `aca-model@feat/dcegm-solver`, `tests/test_dcegm_model_creation.py::test_dcegm_benchmark_model_builds` (the xfail reason will be updated to quote the savings-stage error once the smoothing lands on the aca side).

hmgaudecker:
**Agreed — with three requirements that are load-bearing, not polish. (1/3): the `∂P/∂a` term must survive differentiation.**

Per-node evaluation makes `P(a_i)` a constant in the *consumption* direction, so the Euler inversion itself is untouched — but the carry's exported marginal value is not. In asset-row mode the published marginal value at node `a_i` is

```
dV/dR = u'(c*) + discount_factor · (∂W/∂a at fixed A*) / R'(a_i)
```

and once regime-transition probabilities (or stochastic weights) read `assets`, the `∂W/∂a` term picks up first-order contributions `Σ_targets ∂P_target/∂a · EV_target`. The Danskin cancellation that lets us ignore `∂P/∂R` at the discrete-choice stage does **not** extend here: regime-transition probs are not the softmax of the values they weight, so nothing cancels. Concretely: the probabilities/weights have to sit *inside* the `jax.grad`-differentiated continuation closure (`_get_expected_continuation_value`), never be precomputed as per-node constants outside it.

This failure mode is silent — a missing term shifts the policy smoothly and passes every structural test. The only thing that catches it is a self-targeting, multi-period oracle: a smoothstep transition probability in `assets` spanning several grid cells, with *material* value gaps across target regimes inside the band, compared against dense brute force. That test is the headline red test of the implementation, not an afterthought.

hmgaudecker:
**(2/3): the continuity guard must check the contract's actual criterion — smooth at node resolution — or it will reject your smoothstep.**

The current cliff guard (`_fail_if_euler_law_jumps_in_euler_state`: 401 uniform samples over the grid span + isolated-spike heuristic) false-rejects exactly the artifact you propose: a C², ~$100 band on an assets grid spanning O($100k) is invisible-then-violent to any uniform sampler at feasible sample counts — it looks like a jump even though it is smooth by construction *and properly resolved by the grid*.

The contract's criterion is "smooth at node resolution", so the check should test that, literally: evaluate the savings-stage function on the regime's actual grid nodes plus one–two levels of cell-midpoint refinement; a genuine cliff's increment does not shrink under subdivision, while a band that has dedicated nodes across it halves like a derivative bound says it should (scale-invariant criterion, no absolute jump threshold). One shared implementation for the existing Euler-law check and the new savings-stage checks — the current sampler gets replaced, not duplicated.

Corollary that belongs in the error message: a smoothstep *narrower than a grid cell* is a cliff at node resolution and stays rejected. Your "dedicated grid nodes across the band" is load-bearing, not cosmetic — if the aca side ever shrinks the band below cell width without adding nodes, the validator should fail the build and say to add band nodes.

hmgaudecker:
**(3/3): the relaxation is conditional on asset-row mode — and the mode trigger must widen with it.**

Two couplings to keep straight:

1. **Passive-state taxonomy.** The passive classification currently forbids Euler-state ancestors outright. The relaxation is *conditional*: a non-Euler law reading `assets` is well-defined only under per-node evaluation, so it is admitted in asset-row mode and stays rejected in the single-post-state kernel. Same for transition probs and stochastic weights — this is one rule applied to three function classes, not three rules.

2. **Mode trigger.** Today asset-row mode fires when the *Euler law* reads the Euler state. If we relax the validator for savings-stage functions without widening the trigger, a regime whose only `assets`-dependence sits in a transition probability would pass validation and then be solved by the single-post-state kernel, which evaluates that probability at an undefined asset value — silently. The trigger must become: ANY savings-stage function (Euler law residual `g`, regime-transition probs, stochastic weights, non-Euler laws) reading the Euler state ⇒ asset-row mode.

Each leg gets its own oracle test vs dense brute force (Markov weights in `assets`; a passive law in `assets`; cliff negatives that must still die at build time). Implementation of all three requirements plus the validator/trigger changes is underway on this branch; the smoothing on the aca side and this extension can land independently — the acceptance fixture's xfail reason will move from the savings-stage error to "waiting on the aca smoothstep" once this side is in.

hmgaudecker:
**DC-EGM × ACA: the terminal-pref_type fix works; next blocker is stochastic discrete-state transitions into carry targets**

Verified `2e6ea584` against `aca-model@feat/dcegm-solver` (`1165b0e`) on `pylcm@ddef786` (A100, Marvin): the terminal-`pref_type` stop is gone — the benchmark ACA baseline now solves *past* it. It stops one check later:

```
NotImplementedError: The DC-EGM solver cannot solve regime 'retiree_nomc_inelig_canwork' yet:
the transitions ['next_health', 'next_spousal_income'] into regime 'nongroup_dimc_inelig_canwork'
are stochastic. This configuration is outside the DC-EGM kernel's current scope.
```

**Repro** (you have the aca-benchmark example): build the benchmark baseline with `solver="dcegm"`, `n_assets=8` (so the SSI smoothstep band is node-resolved) + construction-time consumption points, then `.solve()`. The build passes every validator; the solve raises in `egm/step.py::_get_raising_egm_step`.

**Scope** — `egm/step.py::_find_unsupported_target_feature` (~L2097), the stochastic check at L2133–2137:
```python
process_transition_keys = {f"next_{name}" for name in target_process_states}
stochastic = sorted(
    (set(transitions[target]) & stochastic_transition_names) - process_transition_keys
)
if stochastic:
    return f"the transitions {stochastic} into regime '{target}' are stochastic."
```
The kernel admits stochastic **process** states (continuous AR1 — `hcc`, `wage` — integrated via the `weight_{target}__next_{process}` quadrature) but rejects any other stochastic transition into a carry target. ACA's `health` (good-health/disability) and `spousal_income` are discrete states with stochastic **Markov** transitions — not AR1 processes, so not in `target_process_states` — and they evolve across every regime boundary every period.

**Why this gates the whole ACA solve, not a corner.** Unlike the "Markov weights reading `assets`" T2 boundary case, `next_health` / `next_spousal_income` fire on *every* cross-regime carry — there is no living ACA regime whose targets don't carry a stochastically-transitioning discrete state. The first target checked just happens to surface it. So the kernel needs to integrate the carry over a target's discrete-state Markov transition (index the child's discrete axis, weight by the Markov row) — the discrete analogue of the existing process-state quadrature path. This is the "stochastic non-process transition into a carry target" you flagged as a separate feature; it turns out to be load-bearing for ACA, not optional.

No aca-side workaround: `health` / `spousal_income` are genuinely stochastic and central to the model, so I won't make them deterministic.

For context, the brute-force baseline solves + simulates fine on the identical config (GPU peak ~2.5 GB solve / 2.9 GB sim; solve compile ~206 s vs exec ~17 s), so the rest of the DC-EGM profiling deliverable (DC-EGM-vs-brute exec, GPU peak, asset-row-mode contribution) is staged and ready to capture the moment this lands.

hmgaudecker:

PR #384, opened 2026-06-12
Canonical params: the target is a genuine tree level

hmgaudecker:
Stacked on #373; #374 is retargeted onto this branch.

Continues the house pattern into the params system. **Coarse vs. granular**: a *granular* spelling names each target regime explicitly, a *coarse* spelling is target-blind and means "the same thing toward every (reachable) target"; the canonical form the engine stores is always granular, coarse user spellings are sugar that broadcasts into it. #373 applied this to state and regime transitions — this PR applies it to parameters: the target regime is a genuine level of the params tree, and canonical (engine-side) params always store the target-granular form.

## Template & user spellings

- Per-target transition params nest under the target regime — `template[regime][target_regime][func][param]` — replacing the mangled `to_<target>_…` keys. Param qnames parallel engine function qnames.
- `UserParams` resolve at **four levels**, most to least specific:
  1. target level — `{regime: {target_regime: {func: {param: v}}}}`: a value for that target's law only
  2. function level — `{regime: {func: {param: v}}}`: one value broadcasts over the law's targets
  3. regime level — `{regime: {param: v}}`
  4. model level — `{param: v}`

  Exactly one level per param; multi-level specs are ambiguity errors. Missing and unknown keys are reported together.

## Canonical storage

- Canonical flat params key every state-transition-law param per target regime (`<target_regime>__<law>__<param>`), matching the engine's target-prefixed function qnames. A coarse user value materializes one entry per reachable carrying target, **all targets sharing one leaf object** (`cast_params_to_canonical_dtypes` casts once per distinct input object) — bit-identical arithmetic, no per-target copies, identity-based dedup downstream.
- Fixed params materialize to the granular form before partialling; the template trim stays on the user-coarse form.
- A coarse regime transition remains a single shared evaluation: one `next_regime` entry, per-target params rejected.

## Unification rounded out

- Model-level `derived_categoricals` follow the exactly-one-level rule of the other model-level slots: a name declared at model level and regime level is an ambiguity error, also when the grids match (the silent merge-if-equal is gone).

## Performance

- The per-subject regime draw and the optimal-action `unravel_index` are jitted at module level instead of re-tracing an un-jitted vmap on every period-regime call, and the params-completed state-action space is built once per regime per solve/simulate call instead of every period (runtime grid completion — e.g. process gridpoint recursions — rode on it). Bit-identical results; ~40 % fewer host-side function calls on the Precautionary-Savings simulate workload. The draw re-trace was the source of the simulate-runtime alert on #373's benchmark run.

## Docs

- New "Per-target parameters" section in the transitions guide (template nesting + the four resolution levels), coarse/granular vocabulary defined where the transition forms are introduced, a "wrong beliefs" example in the phase-grammar page, CHANGES.md entry, `to_<target>_…` spelling removed everywhere.

Verified: full CPU suite green, ty clean, prek clean; aca-model suite green against this tip with no downstream changes needed.


src/_lcm/transition_checks.py
```diff
@@ -46,6 +47,9 @@
 )
 from lcm.typing import FloatND, IntND, ScalarFloat, ScalarInt
 
+# Within-regime flat param qname `target__law__param`.
+_NUM_PARTS_PER_TARGET_PARAM = 3
```
mj023: Maybe `_NUM_LEVELS_PER_TARGET_PARAM` fits better?

src/_lcm/params/processing.py
```diff
@@ -44,6 +44,7 @@
 from lcm.typing import UserParams
 
 _NUM_PARTS_FUNCTION_PARAM = 3
+_NUM_PARTS_PER_TARGET_PARAM = 4
```
mj023: There are many of these, maybe they could be collected in some central module?

hmgaudecker:
```suggestion
_NUM_PARTS_PER_TARGET_SPECIFIC_PARAM = 4
```
A quick comment of a lin or two above these two might also be worth it.

src/_lcm/params/processing.py
```diff
@@ -53,11 +54,15 @@ def process_params(
...
-    Users can provide parameters at exactly one of three levels:
+    Users can provide parameters at exactly one of these levels:
 
     - Model level: `{"arg_0": 0.0}` — propagates to all functions needing arg_0
     - Regime level: `{"regime_0": {"arg_0": 0.0}}` — propagates within regime_0
-    - Function level: `{"regime_0": {"func": {"arg_0": 0.0}}}` — direct specification
+    - Function level: `{"regime_0": {"func": {"arg_0": 0.0}}}` — direct
+      specification; for per-target transition params this broadcasts over
+      the targets
+    - Target level: `{"regime_0": {"target_0": {"func": {"arg_0": 0.0}}}}` —
```
hmgaudecker:
```suggestion
    - Target-regime level: `{"regime_0": {"target_regime_0": {"func": {"arg_0": 0.0}}}}` —
```
Watch out in the entire codebase to be more specific here. "target" is too ambiguous, unless the context is really tight.

docs/user_guide/transitions.ipynb
```diff
@@ -38,18 +38,35 @@
...
    "metadata": {},
    "source": [
+    "## Coarse and Granular Spellings\n",
+    "\n",
+    "Two words recur throughout this page, describing how precisely a spelling\n",
+    "names its target regimes:\n",
+    "\n",
+    "- A **granular** spelling names each target regime explicitly —\n",
+    "  `{\"retired\": law_a, \"dead\": law_b}` — so the engine knows exactly which\n",
+    "  target gets what.\n",
```
hmgaudecker:
```suggestion
    "  target regime gets what.\n",
```
or is there actually an example for states?

docs/user_guide/transitions.ipynb
```diff
@@ -38,18 +38,35 @@
...
+    "- A **coarse** spelling is target-blind — a bare law, a bare\n",
```
hmgaudecker:
```suggestion
    "- A **coarse** spelling is target-regime blind — a bare law, a bare\n",
```

src/_lcm/regime_building/transitions.py
```diff
@@ -181,8 +183,8 @@ def _add_raw_transition(
     if callable(raw) or isinstance(raw, Phased):
         transitions[f"next_{name}"] = cast("UserFunction", raw)
     elif isinstance(raw, Mapping):
-        for target_name, target_value in raw.items():
-            key = f"next_{name}{QNAME_DELIMITER}{target_name}"
+        for target_regime_name, target_value in raw.items():
```
hmgaudecker: Here you went too far, I think. Double-check.

hmgaudecker (commented):
Autoreview.

mj023 (approved):

src/_lcm/transition_checks.py
hmgaudecker: Consolidated rather than renamed in place (`d5f97f1`): all four of these depth constants now live in one `ParamsQnameDepth` class in `_lcm/utils/namespace.py`. Each attribute name spells the qname pattern it measures (`REGIME__FUNC__PARAM`, `REGIME__TARGETREGIME__FUNC__PARAM`, `TARGETREGIME__FUNC__PARAM`), so the disambiguator travels to every call site — `len(parts) == ParamsQnameDepth.TARGETREGIME__FUNC__PARAM` now says exactly which depth-3 pattern it means. Used "depth" (your LEVELS instinct) over "parts".

src/_lcm/params/processing.py
hmgaudecker: Done (`d5f97f1`). The four scattered `_NUM_PARTS_*` constants — including the literal duplicate `_NUM_PARTS_PER_TARGET_PARAM` defined in both `transition_checks.py` and `pandas_utils.py` — are collected into one `ParamsQnameDepth` class in `_lcm/utils/namespace.py` (beside the qname-namespace helpers), imported by all three modules. A test pins each named depth to a concrete qname of that pattern.

PR #388, opened 2026-06-17
Redesign the solver seam as an ABC (GridSearch + DCEGM stub)

hmgaudecker:
> Stacked on #389 (`feat/target-batch-eviction`).

Behavior-preserving refactor that introduces a per-regime **solver-selection
seam** as a polymorphic `Solver` ABC, so the upcoming DC-EGM PR is a pure
addition whose review can concentrate on the DC-EGM logic alone.

## What this does

- `Solver(ABC)` (`_lcm/solution/contract.py`): an abstract
  `build_period_kernels(*, context) -> SolverKernels` plus a default no-op
  `validate(*, context)`. The engine dispatches polymorphically on the solver
  instance — `solver.validate(context)` then
  `solver.build_period_kernels(context)` — with no switch on solver type and no
  `BruteForce | DCEGM` union. `contract.py` is a true engine leaf (no import
  cycle: `import lcm` resolves cleanly).
- `GridSearch(Solver)` (the renamed `BruteForce`) and `DCEGM(Solver)` live in
  `_lcm/solution/solvers.py`, with `validate` + `build_period_kernels` methods
  and deferred numerical imports; `lcm/solvers.py` is a thin re-export façade.
  The `type(solver)`-keyed `SOLVER_KERNEL_BUILDERS` registry and `registry.py`
  are deleted.
- `Regime.solver: Solver` (default `GridSearch()`), exported from `lcm`
  alongside the `DCEGM` configuration class.
- `DCEGM` is published as the **final** configuration surface (fields + field
  validation), but its engine is not yet wired in: a regime requesting it is
  rejected at model build with a clear `NotImplementedError`.

## Not in scope (the DC-EGM PR, stacked on top of this)

The `_lcm/egm/*` engine, build-time DC-EGM validation, the `SolutionPhase` EGM
fields, the EGM-specific `SolverBuildContext`/`SolverKernels` fields, and the
DC-EGM kernel builder. `solve_brute.py` and the simulation phase are unchanged.

## Verification

- Numerics unchanged: routing the existing grid search through the ABC contract
  is a faithful move; `solver=GridSearch()` explicitly ≡ the default value
  function.
- Full gpu-01 suite green (1208 passed, 15 skipped). `tests/test_solvers.py`
  asserts the default solver, the polymorphic-dispatch parity, `DCEGM`
  construction + field validation, and the build-time guard.
- `ty` and `prek` clean.


timmens (commented):
I really like the idea of making the backward induction solver a replaceable component.

I am not a huge fan of the current implementation, but I could be convinced otherwise. Some things that I find odd:

1. To me a `Solver` abstraction is the perfect use-case for inheritance, or at least a `Protocol` interface. I have looked for a single interface / abstraction in the PR that tells me what the "solver contract" is. Having a clear abstraction should also make it simple for users to implement custom solvers (thinking of optimagic's algorithm interface here).
2. The tests don't seem to test anything interesting again.
3. I actually don't see why we need a DCEGM stub in this PR. I would expect the first PR to design a solver interface / abstraction, for which we then implement the `GridSearch(Solver)` and only check that this works. Then in the next PR we add DCEGM.

mj023 (approved):
I also like the object-oriented approach more. I think the stub is fine, the PR is small enough for this. The interface is a bit limiting as I can imagine that some solvers (Reinforcement learning) would need different aspects of the Regime, like the transition functions, or would return something else then the `max_Q_over_A` functions. But the interface is not complicated, so I would think we can cahnge this if we ever want to implement such a solver.

PR #389, opened 2026-06-19
Add target_batch_size to to_dataframe for sharded target eval

hmgaudecker:
Adds a `target_batch_size` parameter to `SimulationResult.to_dataframe`, chunking the `additional_targets` evaluation (and evicting each chunk to host) independently of the simulate's `subject_batch_size`.

## Why

`subject_batch_size > 0` cannot be combined with distributed (sharded) grids — the value-function array is sharded across devices and can't be gathered onto one, so pylcm rejects the combination. But under a shard the post-simulate `additional_targets` eval still pulls the (host-resident, per-shard) panel back onto the mesh's device 0 and evaluates the target DAG over the whole population in one pass, which can exhaust that device. This knob bounds that eval's device residency without touching the sharded solve.

`target_batch_size=None` (default) falls back to the simulate's `subject_batch_size`, so current behavior is unchanged.

## Test

`tests/simulation/test_subject_batching.py::test_to_dataframe_targets_are_invariant_to_target_batch_size` — simulates single-pass, then chunks only the target eval at `target_batch_size ∈ {2, 3, 100}` (even split, uneven, chunk-larger-than-population), asserting the computed-target column is identical to the single-pass result.

### DIFF
--- a/src/_lcm/simulation/additional_targets.py
+++ b/src/_lcm/simulation/additional_targets.py
@@ -99,12 +99,12 @@ def _compute_targets(
     targets: list[str],
     regime: Regime,
     regime_params: FlatRegimeParams,
-    subject_batch_size: int | None = None,
+    target_batch_size: int | None = None,
 ) -> dict[str, FloatND | IntND | BoolND | np.ndarray]:
     """Compute additional targets for a regime.
 
     The target DAG is vmapped over the regime's in-regime subject-period rows. When
-    `subject_batch_size` is a positive value below the row count, the rows are
+    `target_batch_size` is a positive value below the row count, the rows are
     processed in chunks and each chunk's outputs are pulled to host before the next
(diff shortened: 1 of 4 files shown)


hmgaudecker:
Closing as a false alarm.

The device-0 OOM this `target_batch_size` knob was meant to bound was misattributed. It was the estimation orchestrator's **import-time GPU preallocation** — a module-level `jnp.array` model constant that reserved 95% of device 0 the moment the model was imported, in the pytask process that only launches GPU ranks and never computes. Every run died at ~8 s in that import, **before `to_dataframe` ever executed**, so the post-simulate target eval this knob chunks was never the binding allocation. (It is also panel-shaped — n_subjects × n_periods × a few columns — orders of magnitude below the grid-product buffers.)

Fixed at the source by making the constant a host array (OpenSourceEconomics/aca-model#13); the aca-slurm `target_batch_size` usage is removed. No need for this knob.

PR #390, opened 2026-06-19
Add the endogenous-grid solver family: EGM, DC-EGM, and NEGM

### LINKED ISSUE #387: FUES bounded scan can accept a run of dominated points (n_points_to_scan default under-scans)
## Summary

`refine_envelope`'s bounded same-segment scan (`_find_same_segment_point`, `src/_lcm/egm/upper_envelope/fues.py:400`) inspects only `n_points_to_scan` candidates (default **10**, `src/lcm/solvers.py:102`). When more than `n_points_to_scan` off-segment candidates interleave two points of the same segment in sorted-grid order, the scan never sees the segment's continuation, so the intervening dominated candidates are wrongly accepted onto the envelope.

This is **finding F4** of the external adversarial FUES correctness audit. Full verification write-up: `FUES_AUDIT_VERIFICATION_RESULTS.md` (commit a72bb1a9).

## Reproduction

Upper line `A(x)=x` (anchors `(0,0),(0.1,0.1),(12,12)`, `p=x`) plus eleven points 0.5 below it (`(i, i-0.5)` for `i=1..11`, `p=x-100`), at the shipped default `n_points_to_scan=10`, `jump_thresh=2`:

| `n_points_to_scan` | `n_kept` | `max|dev|` of interpolated value from `A(x)=x` |
| --- | --- | --- |
| 10 (default) | 14 | **0.50** |
| 11 | 3 | 0.00 |
| 14 | 3 | 0.00 |

All eleven below-envelope points survive and the interpolated value function sits a uniform 0.5 below the true envelope. Widening the window to cover all candidates recovers the exact envelope — so this is a genuine correctness defect at the shipped default, not just an approximation knob.

Locked as `tests/solution/test_fues_upper_envelope_audit.py::test_f4_interleaved_segments_give_analytic_envelope_at_default_scan` (`xfail(strict=True)`), with `::test_f4_failure_resolves_when_scan_window_covers_all_candidates` pinning the boundary.

## Reachability

Not exercised by the IJRS examples: every candidate row captured from a small IJRS worker-model DC-EGM solve is **fully concave** (`segment_jumps = 0`, nearest-same-segment-successor offset `0`), so the multi-segment finely-interleaved input never arises there. Triggering F4 needs (a) genuine multi-segment non-concavity and (b) more than `n_points_to_scan` points of one segment falling between two consecutive points of another — a strong per-segment grid-density asymmetry inside an overlap region. Plausible for strongly non-concave targets (ACA's discrete-choice structure, the Laibson liquid/illiquid model) with the 200-node clustered savings grid; should not be assumed safe there.

## Recommended fix

Either:
1. a correctness-critical mode that scans exhaustively (`n_points_to_scan = n_candidates`), or
2. carry explicit segment/bracket ids and jump directly to the next same-segment point (this also closes the related F5 missing-capability gap).

Raising the default `n_points_to_scan` only pushes the boundary out; it does not close the gap. A fix should remove the `xfail(strict=True)` marker on the F4 regression test.

hmgaudecker:
## Summary

This PR adds the endogenous-grid solver family to pylcm and makes solver choice an
explicit part of a regime's economic structure:

- `EGM` for the classic one-dimensional consumption-savings problem;
- `DCEGM` for a liquid Euler margin combined with discrete choices or other
  non-Euler axes;
- `NEGM` for a liquid inner DC-EGM problem nested inside an outer search over a
  durable or illiquid margin;
- typed upper-envelope backends, including a certified exact backend and four
  approximate alternatives.

The implementation also introduces a constraint-routing contract. A solver must
account for every applicable constraint on every candidate-production route: it
must evaluate it, prove it by construction, compile its boundary, or reject the
model during construction.

## Motivation

Grid search spends most of its work evaluating a dense continuous-action grid.
EGM reverses the Euler equation instead: it starts from an exogenous
post-decision-state grid and constructs the current-state grid on which the Euler
condition holds. This makes finely resolved consumption-savings problems much
cheaper, but only when the model exposes the economic roles and dependency
structure that the algorithm needs.

DC-EGM extends that idea to models with discrete choices. Those choices generate
overlapping, potentially non-concave value branches, so the solver must take an
upper envelope. NEGM adds a second continuous margin whose kinks or adjustment
costs make a second Euler inversion inappropriate; it retains DC-EGM for the
liquid margin and searches the outer margin directly.

## Public interface

Solver and envelope names live under `lcm.solvers`. The specialized economic
declarations live under `lcm.consumption_savings_regime`:

```python
from lcm.consumption_savings_regime import (
    ConsumptionSavingsRegime,
    LiquidMargin,
)
from lcm.solvers import DCEGM, ExactEnvelope


LIQUID = LiquidMargin(
    state="wealth",
    action="consumption",
    resources="wealth",
    post_decision_state="savings",
)

working = ConsumptionSavingsRegime(
    liquid=LIQUID,
    solver=DCEGM(
        savings_grid=savings_grid,
        envelope=ExactEnvelope(),
    ),
    states={"wealth": wealth_grid},
    actions={"consumption": consumption_grid},
    functions={
        "utility": utility,
        "savings": savings,
    },
    state_transitions={"wealth": next_wealth},
    transition=next_regime,
)
```

When resources are simply the liquid state, the state can fill that role
directly; an identity `resources(wealth) -> wealth` DAG node is unnecessary.
`LiquidMargin.resources` can instead name a resources function or a
`NetOfAdjustmentCost` declaration when the budget has another investment
decision with adjustment cost (e.g., housing).

The regime declares economic roles; the solver object contains numerical choices.
This keeps names such as `wealth`, `consumption`, and `savings` out of solver
configuration and lets validation explain structural mismatches at `Model(...)`
construction.

| Regime/solver | Intended structure |
|---|---|
| `Regime` / `GridSearch` | General state-action product and arbitrary feasible set |
| `ConsumptionSavingsRegime` / `EGM` | Classic one-dimensional cash-on-hand problem, without discrete/process axes or an envelope |
| `ConsumptionSavingsRegime` / `DCEGM` | One liquid Euler margin plus discrete, passive, or stochastic axes; overlapping branches are resolved by an upper envelope |
| `NestedConsumptionSavingsRegime` / `NEGM` | Inner DC-EGM over the liquid margin and an outer search over a durable/illiquid post-decision grid |

`ConsumptionSavingsRegime` and `NestedConsumptionSavingsRegime` also accept
`GridSearch`, so the same economic declaration can be used as a reference solution.
A plain `Regime` cannot select an endogenous-grid solver because it does not declare
the required margin roles.

## Constraints and `Condition`

Ordinary constraint callables remain fully supported:

```python
def can_pay_premium(cash, premium):
    return cash >= premium
```

For constraints whose comparison structure matters to a solver, the same Boolean
can be written as a `Condition`:

```python
from lcm import implies, ref


constraints = {
    "borrowing_limit": ref("savings") >= 0.0,
    "means_test": ref("assets") < ref("asset_limit"),
    "insured_can_pay": implies(
        premise=ref("insured") == 1,
        consequent=ref("cash") >= ref("premium"),
    ),
}
```

`Condition` supports named/literal and named/named comparisons, intersections
(`&`), unions (`|`), complements (`~`), and implication. A reference can name a
state, action, DAG output, parameter, or declared margin role.

The callable and structured forms have the same meaning and produce the same
feasibility mask. Use `Condition` when:

- a solver needs to retain the named comparison in order to prove the constraint
  from its grid or compile a boundary; or
- the structured form is clearer as a declarative authoring choice.

It is not otherwise required. `Condition.from_callable(...)` preserves an arbitrary
callable but is intentionally opaque: it does not give a solver comparison structure
that was absent from the callable.

This distinction matters for endogenous-grid solvers because they do not construct
the same candidates as grid search. For each solver route and phase, every constraint
receives exactly one disposition:

- **evaluate** it where all of its inputs exist;
- **prove** it because the construction already enforces it;
- **compile** a structured boundary into candidate construction; or
- **reject** the model with an explanation of the unavailable inputs or unsupported
  shape.

The borrowing limit is the main example. The lower node of an EGM-family savings
grid already enforces a lower bound on the post-decision liquid state. Authors can
make that contract explicit with either a structured comparison or the margin-aware
helper:

```python
from lcm.consumption_savings_regime import post_decision_lower_bound


constraints = {
    "borrowing_limit": post_decision_lower_bound(
        margin=LIQUID,
        lower=0.0,
    )
}
```

Under `EGM` and `DCEGM`, pylcm checks that the declared value equals the savings
grid's lower node and then proves the constraint by construction. Under
`GridSearch`, the returned ordinary predicate is evaluated normally. The declaration
is optional if the savings grid is intended to be the sole source of the borrowing
limit. An opaque callable that happens to spell the same inequality cannot be silently
treated as proven, because the solver cannot inspect what it says.

## Solver contracts

The endogenous-grid solvers differentiate the model's declared utility and law of
motion rather than reconstructing a hard-coded savings equation. An analytical
inverse marginal utility function is optional; when absent, pylcm differentiates
utility and numerically inverts marginal utility.

Model construction validates the contract before solving. Among other checks, it
verifies:

- that declared margin roles resolve to continuous states/actions and assembled DAG
  nodes;
- that transition and utility dependencies are available at the stage where a solver
  needs them;
- that every constraint has a legal disposition on every candidate route;
- that the savings grid and any declared post-decision lower bound agree;
- that a nested regime has exactly the liquid and outer structure NEGM supports; and
- that backend-specific numerical configuration belongs to the selected typed
  backend.

Unsupported structures therefore fail during `Model(...)` construction instead of
being accepted and producing a plausible but invalid policy.

## Upper envelopes

`DCEGM.envelope` is a typed configuration:

- `ExactEnvelope` certifies ownership among the finite candidates supplied to it;
- `FUESEnvelope` provides the Fast Upper-Envelope Scan;
- `RFCEnvelope` provides Rooftop Cut;
- `LTMEnvelope` provides a quadratic local-upper-bound reference;
- `MSSEnvelope` provides a HARK-style left-to-right segment sweep.

`ExactEnvelope` is the default. It relies on pylcm's own compiled exact-affine kernel,
built and shipped as part of the package, to make candidate-ownership decisions in
exact integer arithmetic over the stored floating-point operand bits. If that kernel
is unavailable, selecting `ExactEnvelope` raises a capability error rather than
falling back to floating arithmetic that cannot provide the same guarantee. Builds
that deliberately do not need the certified backend can set
`LCM_SKIP_EXACT_AFFINE=1` and select one of the approximate envelope configurations.

The certification is deliberately scoped: it identifies the envelope of the finite
candidate set presented to the kernel. The resolution of a sampled constrained branch
remains controlled by `DCEGM.n_constrained_points`.

## What this changes internally

- Adds one- and two-margin regime declarations and keeps their invariants out of the
  general `Regime` module.
- Adds canonical EGM continuation carries, Euler inversion, endogenous-grid sorting,
  interpolation, and policy publication.
- Carries discrete states/actions, stochastic processes, passive states, fixed
  parameters, and target-specific laws through DC-EGM's candidate construction.
- Applies the intrinsic liquid-budget mask in simulation as well as solution.
- Adds route-aware constraint planning shared by grid search and the EGM family.
- Adds the native exact-affine envelope kernel with CPU support and optional CUDA
  compilation.
- Adds benchmark models and focused analytical, brute-force-comparison, batching,
  constraint-routing, envelope, installation, and import-surface tests.

## Documentation

The documentation starts from model and constraint shape, then explains which solver
fits that shape. It includes:

- the `EGM` / `DCEGM` / `NEGM` decision boundary;
- ordinary callable and structured `Condition` authoring;
- named/literal, named/named, intersection, union, complement, and implication syntax;
- the precise relationship between a post-decision lower-bound declaration and the
  savings grid;
- typed envelope selection and the scope of exact certification; and
- complete one-margin and nested consumption-savings examples.


src/_lcm/solution/backward_induction.py
```diff
@@ -256,7 +271,183 @@ def solve(
...
+        next_regime_to_V_arr=next_regime_to_V_arr,
+        next_regime_to_egm_carry=next_regime_to_egm_carry,
+        flat_params=flat_params,
+        period=period,
+        ages=ages,
+    )
+    if result.carry is not None:
+        period_egm_carries[regime_name] = result.carry
+    if result.sim_policy is not None:
+        period_sim_policies[regime_name] = result.sim_policy
```
mj023: Maybe the result should be returned instead of modifying the dicts inside the function.

src/_lcm/egm/ds_pension_benchmark.py
mj023, line 1: This should probably be in the benchmarks folder and not inside `_lcm`.

src/_lcm/egm/ds_pension_driver.py
mj023, line 1: This also not in `_lcm`.

mj023 (commented):
Great additions, I was obviusly not able to check everything super thoroughly, but the algorithms look interesting and most of it seems to be well tested againt multiple benchmarks. Just some general thoughts: 

General Structure:

I think the structure of the EGM folder is a bit cluttered, if possible, I think it would be better if the solvers could be split up so that each one has their own submodule.

Solvers:

As I understand it, the solvers are all specialized for slightly different models, which is fine for now. I feel like there is some overlap between the solvers, for example the `one_asset_egm_step` could reasonably be solved by the standard DC-EGM. I think we should overall try to find the most general solver that works well on GPU and focus on that for further development, maybe even remove some of the others later.

Upper Envelope Kernels:

I think here the same applies as for the solvers, we should try to at some point determine which methods are good and only support these. A good method would ideally return a constant number of gridpoints and be easily parallelizable to work on the GPU. The best fitting here will probably be the interpolation based algorithms like the query-algorithm or ltm-algorithm (which are nearly the same on the first look?). I am now quite sure, that both can also be implemented in a way, such that they are on average close to $\Theta(n)$ and can still be parallelized (we would need to write the CUDA kernel ourselves).

src/_lcm/solution/backward_induction.py
hmgaudecker: Agreed — the period kernel already returns a `KernelResult`, so mutating the caller's dictionaries here obscured ownership. The helper (now `_run_period_kernel`) returns the result and the backward-induction loop accumulates the optional outputs (65c7469). As follow-ups in the same series, the generic layer now uses solver-neutral names throughout (`KernelResult.continuation` / `.simulation_policy`, `next_regime_to_continuation`, and the capability `solves_from_continuation` instead of `solves_via_dcegm`), `backward_induction.solve` returns a named `BackwardInductionResult`, and the solver implementations moved into per-solver modules (`grid_search.py`, `dcegm.py`, `negm.py`, `one_asset_egm.py`, `two_dim_egm.py`).

src/_lcm/egm/ds_pension_benchmark.py
hmgaudecker: Agreed that this is a research artifact rather than runtime `_lcm` code. We are moving non-ASV solver comparisons to a dedicated solver-benchmarks repository, so the file was removed from pylcm entirely rather than relocated inside the package (8ffe800; the whole `benchmarks/ds_replication/` workflow collection went with it). The DS pension test model and its solver/oracle tests stay — they validate library contracts against the brute solve.

src/_lcm/egm/ds_pension_driver.py
hmgaudecker: Agreed — the experiment driver belongs with the external comparison benchmark, not in the installed engine. Moved out together with the pension benchmark in 8ffe800; its harness tests went along and nothing in the runtime package references it anymore.

src/_lcm/solution/contract.py
```diff
@@ -6,27 +6,81 @@
...
+# The cross-period continuation channel a continuation-based parent
+# interpolates. Named solver-agnostically on the seam so the engine threads it
+# without knowing it is an EGM carry; today the only continuation payload is
+# the EGM carry itself.
+type ContinuationPayload = EGMCarry
+
+# The published off-grid simulation-policy artifact, under the same rule: the
+# engine stores and returns it opaquely; today the only implementation is the
+# EGM-published policy.
+type SimulationPolicy = EGMSimPolicy
```
timmens: This confuses me a bit. The names of the types and the fact that they are used in the contract tells me that they are abstract. But then we set them to EGM versions. Is that desired?

src/_lcm/solution/two_asset_egm.py
```diff
@@ -0,0 +1,384 @@
...
+        pension: Float1D,
+        next_value_retired: Float1D,
+        next_marginal_retired: Float1D,
+        **params: FloatND,
+    ) -> FloatND:
+        result = g2egm_retiring_step(
+            next_value_retired=next_value_retired,
+            next_marginal_retired=next_marginal_retired,
+            liquid_grid=liquid,
+            m_grid=liquid,
```
timmens: My AI tells me:

**Bug:** liquid/pension here are this period's grids (they come from state_action_space.states in _core_args, which the solve loop replaces with period-t age-specialized nodes). But next_value_* is tabulated on the target's period t+1 grid, and m_grid/n_grid are used as its interpolation axes inside the step. Result: silently wrong values whenever grids change with age. The boundary call here is worse: liquid_grid=liquid uses the working regime's grid for the retired regime's value, wrong even without age specialization if the two regimes have different liquid grids. Same issue in the interior branches (rfc_two_asset_step, g2egm_step below). OneAssetEGM and DCEGM were fixed for this in this PR; TwoDimEGM was not. Tests miss it because all test grids are identical and static. Please pass the target's next-period grids (like OneAssetEGM's next_liquid_grid) and add a regression test, or reject AgeSpecializedGrid and mismatched grids in validate().

src/_lcm/solution/egm.py
timmens, line 1: Specifying a model with a "wealth" state builds fine but then fails because of a missing keyword-only argument "liquid". Either I am doing something wrong or these models have hardcoded some state names. If this is true, then we should either

1. Raise a more informative error during regime construction, or
2. Try to actually make it independent of the name.

src/_lcm/solution/two_asset_egm.py
timmens, line 1: Same comment about the hardcoded "liquid" state name as for the `one_asset_egm.py`.

src/_lcm/solution/contract.py
hmgaudecker: Yes, deliberately — but you are right that the aliases do not say so loudly
enough on their own.

The two names exist to state *what the engine is allowed to know* about the
payload, which is nothing. The backward-induction loop stores a continuation
payload, threads it to the next period, and hands it back to a solver; it never
reads a field. Naming the slot `ContinuationPayload` rather than `EGMCarry` is
what keeps that ignorance checkable — if the engine ever grew a `.endog_grid`
access, it would read as a layering violation at the call site instead of as an
ordinary attribute read.

What the alias is *not* is a polymorphism point. There is exactly one
implementation today, and a second one would not arrive by widening the alias to
a union: it would arrive as a protocol, because the engine would then need a
declared operation to call. So the alias is a documented single-implementation
seam, not an abstraction with one subclass so far.

Two things could make that clearer, and I am happy to do either:

1. rename to `ContinuationPayloadT` / `SimulationPolicyT` to signal a
   type-level alias rather than a base type; or
2. keep the names and make the comment state the invariant as a rule the engine
   must obey ("the engine treats this opaquely; adding a field access here is
   the signal to introduce a protocol"), rather than as a note about today's
   binding.

My preference is (2) — the confusion is about the *contract*, not the spelling.

src/_lcm/solution/two_asset_egm.py
hmgaudecker: Confirmed and fixed — thank you, this was a real one.

`TwoAssetEGM` now resolves both interpolation axes from the *target*, the same
way the one-asset solver does, in `86637cd2`:

```python
next_liquid_grid=target_period_grid(
    context=context, period=period, target=target,
    target_state_name=self.liquid_state,
),
next_pension_grid=target_period_grid(
    context=context, period=period, target=target,
    target_state_name=self.pension_state,
),
```

so the axes are the target regime's period-`t+1` nodes rather than the source's
period-`t` ones. That covers both halves of what you flagged: age-specialized
grids, and two regimes that simply declare different liquid grids.

One thing to check me on: the **boundary** call still passes `own_liquid_grid`.
That branch is the retirement boundary, where the step is evaluated against the
source's own grid by construction rather than against a target's tabulation — so
I believe it is correct rather than an oversight, but it is exactly the case
your comment says is "worse", so please say if you read it differently.

src/_lcm/solution/egm.py
hmgaudecker: You were not doing anything wrong — and we took option (2).

The one-asset solver never hardcoded the *state*: it fills its Euler role
positionally from the regime's single continuous state, so a `wealth` state is
fine. What was hardcoded was the *law parameter* it reads (`return_liquid`), and
that is what produced the missing-keyword failure you hit. `43c444b3` makes it
declarable:

```python
return_param: str = "return_liquid"   # the conventional spelling stays the default
```

`tests/solution/test_egm_target_state_name.py` now covers the case your report
describes end to end — a working regime holding `wealth` continuing into a
terminal regime holding `estate`, i.e. source and target that do not even spell
the state the same way.

Worth noting the resolution rule that fell out of it, since it is the part that
was actually ambiguous: the Euler inversion reads the target's value on the
target's grid and the target's transition parameters out of the target's
namespace, so both are resolved *at the target*, never at the source.

src/_lcm/solution/two_asset_egm.py
hmgaudecker: Same fix, and here the *state* names are declarable too, since a two-asset
regime has two continuous states and position alone cannot say which is which:

```python
liquid_state: StateName = "liquid"
pension_state: StateName = "pension"
```

Both default to the conventional spelling, so existing models are unaffected,
and `validate()` rejects the degenerate case where the two are given the same
name.

src/_lcm/solution/contract.py
timmens: I don't have a preference here. Please go ahead with what you believe is best.

hmgaudecker: Done, option (2) — the comment now states the rule the engine must obey rather than today's binding, and says explicitly that a second implementation arrives as a protocol, never as a widened union.

Writing it out turned up something the old comment got wrong, so the result is not the symmetric pair I proposed. The two seams are not the same rule:

- `ContinuationPayload` **is** opaque, as claimed — `backward_induction.py` only types the dicts it threads and never touches a field.
- `SimulationPolicy` is not. `simulation/simulate.py` interpolates the published rows directly (`endog_grid`, `value`/`policy`, `marginal_utility`) and reads the row-axis name tuples besides. The old comment's "under the same rule: the engine stores and returns it opaquely" was false, and stating the invariant as a rule is what exposed it — a note about today's binding could have stayed wrong indefinitely.

So the second alias now states a *fixed read set* instead: those are the fields the engine may read, a second implementation has to publish them, and any further field it reaches for is the protocol signal. Same discipline, honestly scoped.

hvgaudecker@snellius.surf.nl
timmens, line 1: Is this supposed to be committed here?

hmgaudecker: Ofc not, thanks for catching!

timmens (approved):
Very nice!

1. **Plain EGM bypasses its lower-bound proof**
   (`src/_lcm/solution/egm.py`): `validate_model` rejects every constraint before
   route planning, including a matching `post_decision_lower_bound`. Let
   proof-carrying constraints reach the planner and reject only unmet ones.

2. **`Condition` is broader than current use** (`src/lcm/condition.py`,
   `src/_lcm/constraints/`): the only shipped structural consumer recognizes one
   numeric lower bound; no boundary compiler exists. A typed `LowerBound` or
   `InequalityConstraint` would cover this with much less complexity. What concrete
   second consumer justifies the Boolean DSL?

3. **`name_in_dag` leaks internals**
   (`src/lcm/consumption_savings_regime.py`): rename it to `output` or `name`, or
   infer it.

4. **`Regime.solver` overstates support** (`src/lcm/regime.py`): it lists EGM,
   DCEGM, and NEGM, although these require specialized regime classes. Document the
   pairing explicitly.

hmgaudecker:
Thanks! 1, 3, 4 are implmented.

>     2. **`Condition` is broader than current use** (`src/lcm/condition.py`,
>        `src/_lcm/constraints/`): the only shipped structural consumer recognizes one
>        numeric lower bound; no boundary compiler exists. A typed `LowerBound` or
>        `InequalityConstraint` would cover this with much less complexity. What concrete
>        second consumer justifies the Boolean DSL?

The genesis here is that we needed it in #400, but then introduced it here so we don't introduce a transient interface. #400 already contains the substantive boundary-processing algorithm: it maps declared case and piecewise-affine metadata into asset-axis breakpoints and interval-specific EGM problems. But your comment was constructive in the sense that some elements have not been wired in yet, will do so now!

PR #391, opened 2026-06-21
Device-local continuation-V for fixed shards (no all-gather)

hmgaudecker:
## Problem

At distributed-estimation scale the grid-search solve OOMs on device 0: the kernel passes `next_regime_to_V_arr` unmapped, so each vmapped state cell dynamic-indexes the *full* target V leaf. Under `NamedSharding` on a sharded discrete axis (e.g. a permanent type), XLA inserts an **all-gather**, materializing the entire continuation V-array on every device.

## Fix

For fixed (identity-transition) distributed states — the leading V-array axes — `max_Q_over_a` co-maps each with the matching axis of every `next_regime_to_V_arr` leaf that carries it (outer `vmap`, per-leaf `in_axes`: `0` where carried, `None` where pruned). The interpolation then reads only the device-local slice; no all-gather.

**Solve-path only.** Simulate keeps the full continuation V (subjects are not type-aligned across devices); solve and simulate already build separate `Q_and_F` sets, so the contracts don't collide.

## Mechanism, file by file

- `dispatchers.py` — `vmap_1d` gains `co_mapped_in_axes` (per-argument `in_axes` override).
- `max_Q_over_a.py` — split co-mapped (leading) states from the inner productmap; wrap in per-state co-map vmaps.
- `V.py` — the interpolator drops co-mapped states from its discrete lookup axes.
- `Q_and_F.py` — omit the sliced next-states from the interpolation call.
- `processing.py` — detect co-mappable states (distributed ∧ identity-transition) and build per-state, per-leaf `in_axes` (a target that prunes the state keeps its full leaf).
- `contract.py` / `solvers.py` — thread through `SolverBuildContext`.

## Tests (`tests/test_distributed.py`, 4 CPU devices)

- **Structural** (the red driver): the compiled solve kernel HLO contains no `all-gather` of the continuation V.
- **Correctness**: a sharded solve equals the single-device solve *per type* — numerics identical, device-local index correct (this is what a wrong local index would break).
- Plus a `vmap_1d` co-map unit test in `tests/test_dispatchers.py`.


hmgaudecker:
Ouch, thought I had asked you for a review ages ago, @mj023... Apologies. Since the already-approved PRs depend on this one, it would be great if you could take a quick look. If we find a better way to save mem down the road, we can always revert again.

mj023 (approved):
Everything looks correct! I think there might be some way to handle this with `jax.with_sharding_constraint`, but you are right, we can merge and revert this, if we ever find a better solution.

PR #393, opened 2026-06-29
Pin numpy to <2.5.0

mj023:
Numba does not support the new Numpy version 2.5.0. This created a dependency mismatch because Pixi tried to install a very old and deprecated version of Numba instead. Therefore Numpy will be pinned to a version <2.5.0 until this is fixed with Numba.


hmgaudecker (commented):
Great analysis, thanks!

hmgaudecker (approved):

PR #395, opened 2026-07-02
Certainty-equivalent seam for non-EU preferences (Epstein-Zin)

### LINKED ISSUE #385: ENH: Expose/generalize the aggregator so the certainty-equivalent sees the value distribution (non-EU preferences)
### Motivation

The model class is meant to admit a general certainty-equivalent operator $\mathcal{H}$
(expected utility, Epstein–Zin, and other non-expected-utility recursive preferences).
The current solvers only realise the **linear expectation**: the engine forms the scalar
`E_next_V` as a weighted average of the next-period value and passes *only that scalar*
into the aggregator `H`.

`src/_lcm/regime_building/Q_and_F.py` (≈ L187–200):

```python
next_V_expected_arr = jnp.average(next_V_at_stochastic_states_arr, weights=...)  # linear E[·]
...
E_next_V = E_next_V + active_regime_probs[k] * next_V_expected_arr
...
Q_arr = functions["H"](utility=U_arr, E_next_V=E_next_V, **_build_H_kwargs(...))
```

Consequently a **custom `H` cannot express a nonlinear certainty-equivalent** such as
Epstein–Zin, `CE = (E[V^{1-α}])^{1/(1-α)}`, because the nonlinear transform must hit `V`
*before* the expectation, and `H` never sees the distribution of `V` — only the scalar
`E[V]`. There is no `certainty_equivalent` hook on `Regime` today (only `H`). The
`max_Q_over_a.py` comment ("`H(U,v)=u+β·v` … the only one that …") already hints at this
limitation.

### Proposal

1. Introduce a certainty-equivalent / value-transform seam so non-EU preferences are
   expressible. Options to weigh:
   - a `certainty_equivalent` callable on `Regime` that receives the next-period value
     distribution (values + transition weights) and returns the scalar CE, with the
     current linear expectation as the default; or
   - generalise the aggregation so a user transform `g` is applied to `V'` before the
     weighted average and `g⁻¹` after (sufficient for power-mean / Epstein–Zin).

   Keep the default path (linear `E[V]`) unchanged for performance and backward
   compatibility.
2. Once the seam exists, ship **example models** exercising it — at minimum an
   **Epstein–Zin** lifecycle model, and ideally one or two further non-EU forms.

### Acceptance

- A documented way to specify a nonlinear CE without monkey-patching the engine.
- An Epstein–Zin example model with a test pinning its solved values.
- Default (expected-utility) behaviour and fixtures unchanged.

hmgaudecker:
Closes #385.

## What this PR does

A regime can declare a nonlinear certainty equivalent over the next-period value distribution, so non-expected-utility recursive preferences (Epstein–Zin, risk-sensitive) are expressible without touching the engine:

```python
Regime(
    functions={"utility": utility, "H": H_epstein_zin},
    certainty_equivalent=PowerMean(),
    ...
)
```

The solve then aggregates the continuation as

$$\mathrm{CE} = g^{-1}\Bigl(\sum_r p_r \, \mathbb{E}_w\bigl[g(V'_r)\bigr]\Bigr)$$

— `g` applied elementwise before the expectation over stochastic state transitions, per target regime, `g⁻¹` once after the regime-probability-weighted sum — and `H` receives the CE through its `E_next_V` argument. Solve and simulate share the same decision functions, so simulated choices are CE-consistent; the NaN-diagnostics path reproduces the same math.

## User API

- **Certainty equivalents** (`lcm.certainty_equivalent`, defined engine-side behind a thin re-export façade like `lcm.solvers`):
  - `CertaintyEquivalent` — ABC the engine dispatches on; `certainty_equivalent=None` (the default) is the linear expectation and traces a graph byte-identical to the status quo.
  - `QuasiArithmeticMean(transform=g, inverse=g_inv)` — the generic transform pair. Callables take the value array via the reserved argument `value`; every further signature argument becomes a runtime param under the pseudo-function name `certainty_equivalent` in the params template.
  - `PowerMean()` — the Epstein–Zin power mean with runtime param `risk_aversion` (γ). `risk_aversion = 1` is the geometric-mean (log) limit `exp(E[log V'])`; `risk_aversion = 0` reduces to the linear expectation.
- **Koopmans aggregators** (`lcm.temporal_aggregation`):
  - `H_linear` — `U + β·CE`; also the default injected when a non-terminal regime supplies no `H`.
  - `H_epstein_zin` — `((1-β)·U^ρ + β·CE^ρ)^(1/ρ)`, parametrized directly by the `intertemporal_elasticity_of_substitution` ψ (curvature ρ = 1 − 1/ψ computed inside; ψ = 1 is the Cobb–Douglas limit).
- **Validation at model build**: terminal regimes reject a CE (no continuation to aggregate), and `DCEGM` + CE is rejected with a message naming `GridSearch` — Euler-inversion EGM assumes expected utility.

## Example + tests

- `lcm_examples/epstein_zin.py`: an Epstein–Zin lifecycle model in the spirit of the Atal–Fang–Karlsson–Ziebarth (2025) consumer block — savings, a two-state health Markov chain, health-dependent survival into a terminal bequest regime — parametrized mortality-style (`n_periods`, grid-size knobs, `survival_probs`).
- `docs/examples/epstein_zin.ipynb`: executed notebook covering the recursion, the mapping onto pylcm, the pitfalls (positivity/`0^{1-γ}=∞`, stateless targets, solver restriction), and a rendered figure comparing mean wealth paths for two risk aversions at a common IES.
- Solved values and simulated policies are pinned against an independent NumPy backward induction written in the tests (γ = 0.5 and the γ = 1 log limit; `rtol=5e-5` / `1e-5`); a reduction test confirms `risk_aversion = 0` equals the no-CE solve, and a divergence test that a nonlinear CE changes it.
- API tests: params-template discovery, name collisions, terminal/DCEGM/`Phased` rejection, aggregator units (CES form, Cobb–Douglas limit, default-H identity).

## Non-goals

Full-distribution CE callables (enabled by the `CertaintyEquivalent` subclass seam, not shipped), EZ-EGM, CE params from DAG outputs, model-level broadcast of `certainty_equivalent`.

## Coordination

- Based on #391 (`feat/type-local-continuation-v`); retarget to `main` once that lands.
- `feat/dcegm` (#390) is a sibling on the same base: shared-file hunks here are deliberately small and local (`Q_and_F.py`, `processing.py`, `contract.py`); whichever merges second resolves them. `feat/dcegm`'s real `DCEGM.validate` should keep rejecting regimes whose `SolverBuildContext.certainty_equivalent` is not `None`.
- Also fixes `docs/user_guide/tiny_example.ipynb` (rotted `regime`/`regime_name` column names); every notebook in the docs now executes cleanly from a cold cache.


src/_lcm/certainty_equivalent.py
```diff
@@ -0,0 +1,161 @@
...
+    """Apply `g(v) = v^(1 - risk_aversion)`, or `log(v)` at `risk_aversion = 1`."""
+    return jnp.where(
+        risk_aversion == 1.0, jnp.log(value), value ** (1.0 - risk_aversion)
+    )
+
+
+def power_inverse(value: FloatND, risk_aversion: FloatND) -> FloatND:
+    """Apply `g^(-1)(v) = v^(1 / (1 - risk_aversion))`; `exp(v)` in the log case."""
+    # The unselected power branch must not divide by zero at `risk_aversion = 1`.
+    safe_risk_aversion = jnp.where(risk_aversion == 1.0, 0.0, risk_aversion)
```
mj023, line 100: Not sure if it matters if we divide by zero, if we then select the other value anyways.

src/lcm/temporal_aggregation.py
```diff
@@ -0,0 +1,43 @@
...
+
+    The runtime parameter is the intertemporal elasticity of substitution
+    `psi`; the aggregator curvature is `rho = 1 - 1/psi`. `psi = 1` is the
+    Cobb-Douglas (log) limit `U^(1-beta) * CE^beta`. Pair with
+    `certainty_equivalent=PowerMean()` for the full Epstein-Zin recursion;
+    both `U` and `CE` must be strictly positive.
+    """
+    rho = 1.0 - 1.0 / intertemporal_elasticity_of_substitution
+    # The unselected CES branch must not divide by zero at `ψ = 1`.
+    safe_rho = jnp.where(rho == 0.0, 1.0, rho)
```
mj023, line 41: Same as above.

mj023 (approved):
Looks good! I am wondering if it might be good to increase the scope of the h function, such that it also contains the whole aggregation of `next_V_at_stochastic_states`. Then it would be even more flexible, but we could still offer the defaults for normal expectations and Epstein-Zin preferences. Maybe it's easier to keep it seperate for people who only want state dependent `beta` etc. though.

PR #397, opened 2026-07-03
Make the persistent compilation cache actually cache pylcm's programs

### LINKED ISSUE #392: BUG: Recompilation even with activated Compilation Cache
#### Problem description

Even with activated Compilation Cache subsequent Re-Runs of the same Model seem to trigger compilation on the very first `model.simulate()`. This is probably caused by the combination of AOT-Compilation and the Compilation Cache, which seem to not work together.

#### Proposed Fix

We can build our own compilation cache. It is possible to export AOT-compiled functions. If we can hash the model and save the associated functions so that they can be retrieved upon starting a new run, we would be able to avoid recompilation.

hmgaudecker:
Closes #392.

## Diagnosis

The persistent compilation cache was never populated, so every fresh process recompiled the whole model. The cause is not an AOT/cache incompatibility: JAX only writes an executable to the cache when its compile time exceeds `jax_persistent_cache_min_compile_time_secs`, which defaults to **1 second**. A pylcm model compiles as many small per-regime/per-period programs, and nearly all of them compile in well under a second — so with JAX's default threshold the cache directory stays empty and nothing is ever reused.

Once the cache is populated, the AOT-compiled programs (`.lower().compile()` in both solve and simulate) hit it just like lazily jitted ones. On the mortality example (CPU and GPU alike), a fresh-process re-run drops from ~1.7s to ~0.5s in `simulate` with ~70 cache entries written; the residual is Python-side tracing and lowering, which no XLA cache can skip.

## What this PR does

Importing `lcm` zeroes the threshold:

```python
if os.environ.get("JAX_PERSISTENT_CACHE_MIN_COMPILE_TIME_SECS") is None:
    jax.config.update("jax_persistent_cache_min_compile_time_secs", 0)
```

- Applied via `jax.config` (not `os.environ.setdefault`) so it takes effect even when `jax` is imported before `lcm` — JAX reads the environment variable only once, at import time.
- A user-set `JAX_PERSISTENT_CACHE_MIN_COMPILE_TIME_SECS` wins.

This makes the third JAX default pylcm sets at import, next to `XLA_PYTHON_CLIENT_PREALLOCATE` and `JAX_COMPILATION_CACHE_DIR`. The user guide (installation and tuning pages) documents it.

## Tests

Fresh-interpreter subprocess tests in `tests/test_compilation_cache.py`: the threshold is zero after `import jax; import lcm` (the adversarial import order); a user-set threshold survives the import; and — the actual #392 regression — compiling a sub-second kernel writes an entry to a fresh cache directory.

## Base branch

Targets `feat/terminal-rows` (#396); lands on `main` via the usual cascade. The change is independent of that branch's content — it just continues the stack.


mj023 (approved):
Looks good, I am not sure if we really need these new tests, the feature is simple enough and the test suite for lcm starts to become quite big. Feel free to include them though.

hmgaudecker:
> Looks good, I am not sure if we really need these new tests, the feature is simple enough and the test suite for lcm starts to become quite big. Feel free to include them though.

Thanks! 

Just not sure how to square this with your [comment here?](https://github.com/OpenSourceEconomics/pylcm/issues/392#issuecomment-4875129883)

mj023:
> Just not sure how to square this with your [comment here?](https://github.com/OpenSourceEconomics/pylcm/issues/392#issuecomment-4875129883)

I just wanted to prove that I wasn't crazy when I created the Bug Report. The issue seems to happen rather infrequently and I could not find the conditions to reproduce it on every re-run. I am not sure we can find the reason without spending a lot of time on it and because it's not so severe, I think we can safely ignore it for now. Nonetheless I think reducing the compilation time threshold is good, specifically for the reasons mentioned in this PR, that you can get many small functions that are just <1m.

PR #398, opened 2026-07-06
Age specialization: per-age regime functions/constraints and age-dependent continuous-state grids

hmgaudecker:
## Summary

Adds age specialization to pylcm: one regime definition whose *functions,
constraints, or continuous-state grids* vary with age. Two user-facing markers
share a common base (`_AgeSpecialized`), each wrapping a `build(age)` factory
plus a `signature(age)` function — cells with equal signatures share one built
object, so compiled programs are deduplicated across ages. Equal-signature
trust differs by marker: a function's closure can't be inspected, so
`AgeSpecializedFunction` trusts the signature on the author's word; a grid can
be, so `AgeSpecializedGrid` cross-checks a shared group's resolved nodes at
build time and raises if an equal signature turns out to hide genuinely
different grids.

### `AgeSpecializedFunction` — per-age functions and constraints

- Swaps the implementation of a regime function or constraint by age. The
  motivating use case is calendar-policy tracking in cohort models: a GETTSIM
  tax function whose policy environment moves with the agent's age, rebuilt per
  `(cohort, age)` cell without duplicating regimes.
- Solution: `Q_and_F` resolves the specialized function per age during backward
  induction; representative-age resolution in `_process_regime_core` keeps
  templates and the params template consistent (including
  `Phased(AgeSpecializedFunction)` in a carried state's `solve` side).
  `build(age)`'s parameter list is validated to stay invariant across periods —
  only the closure's bound behavior may vary.
- Simulation: `next_state` is period-indexed so per-age transitions simulate
  correctly; `to_dataframe(additional_targets=...)` rejects age-specialized
  functions it cannot represent at a single age. Initial-conditions feasibility
  checking rejects subjects starting at a period other than the regime's
  representative one, rather than silently checking them against the wrong
  age's policy closure.
- Guardrails (each rejected at model construction with a targeted error):
  deterministic state transitions that are age-specialized (bare or via
  `MarkovTransition(...)`, including per-target and `Phased` forms); regime
  transitions that are age-specialized, wrapped, or whose dependency graph
  reads an age-specialized helper (scanned across both `functions` and
  `constraints`); terminal-regime markers; a `build(age)` whose parameter list
  varies by period.
- Docs: user-guide page `docs/user_guide/age_specialized.md`.

### `AgeSpecializedGrid` — age-dependent continuous-state grids

- A continuous state's grid **bounds** may move with age while `n_points` (and
  the grid class) stay fixed — shape-invariant, so V-array shapes, the rolling
  continuation template, and kernel-dedup topology are untouched. The
  motivating use case is an asset state with an age-dependent borrowing floor
  `a̲(age)` (Guvenen & Smith 2014, and any life-cycle model with an age-varying
  constraint): on a single fixed grid, cells below a tighter age's floor are
  infeasible, and their `-inf` continuation poisons the value function through
  interpolation (`0 · (−inf) = NaN`, propagated backward).
- Solution: the current-period state axis is runtime data — each period's node
  values are substituted into the shared compiled kernel without retracing;
  the continuation `V_{t+1}` is interpolated on the **target regime's
  period-(t+1) grid**, with the `Q_and_F` group signature folding in the
  continuation-grid identity so periods never false-share a program.
- Simulation: the forward pass interpolates the continuation on the same
  per-period grids as the solve.
- Validation: `build(age)` must return the same grid class, `batch_size`,
  points mode, and — for concrete grids — the same resolved node shape,
  dtype, and weak_type at every age (`RegimeInitializationError` otherwise); a
  stochastic process cannot be wrapped; markers are accepted only as *state*
  grids. A shared `Q_and_F` group's resolved nodes are additionally
  cross-checked at build time, so an under-specified or constant `signature`
  that collides two genuinely different grids raises instead of silently
  sharing the wrong one.
- Design notes: `DESIGN-age-dependent-grids.md`.

### Usage

Downstream, an annual lifecycle model with per-(cohort, age)
GETTSIM policy environments solves and simulates on this branch. The grid
surface is exercised end-to-end by the Guvenen–Smith (2014) replication in
lcm-replications, where the age-tracking asset floor eliminates the `-inf`/`NaN`
poisoning of the fixed-grid solve.

## Test plan

- [x] Full suite: 1261 passed, 48 skipped (`pixi run -e tests-cpu tests`);
      `ruff check` and `ty` both fully clean.
- [x] Functions: `tests/regime_building/test_age_specialization.py`
      (validation surface, sharing by signature),
      `tests/solution/test_age_specialized_solve.py` (solve/simulate
      equivalence against a hand-built per-age model).
- [x] Grids: `tests/solution/test_age_specialized_grid_solve.py` — an
      age-invariant `AgeSpecializedGrid` reproduces the plain fixed-grid solve
      **bit-for-bit**; an age-tracking floor solves with no `NaN` poisoning;
      V monotone in wealth; simulation yields finite, positive consumption;
      non-shape-invariant grids are rejected.
- [x] Equivalence gates: constant `AgeSpecializedFunction` signature
      reproduces the unspecialized solution exactly; models without any
      age-varying grid build byte-identically to the pre-feature path.
- [x] Hardening from a full-branch review pass: resolved-node cross-check for
      grid dedup, `build(age)` parameter-list invariance, `Phased`
      carried-state marker detection, the initial-conditions feasibility
      age-guard, and rejection of stochastic-process grids — covered in
      `tests/regime_building/test_age_normalization.py` and
      `tests/solution/test_age_specialized_solve.py`.

https://claude.ai/code/session_016BLLuvQwD2bdXgaTcJWWUm


DESIGN-age-dependent-grids.md
timmens, line 1: This should probably go right?

tests/solution/test_age_specialized_grid_solve.py
timmens, line 1: Could consider renaming this to `test_age_specialized_grid.py` and the other file to `test_age_specialized_function.py`.

timmens (approved):
Approved, very nice feature!

_One architectural question:_ What we are implementing here is conceptually similar to our `Phased` solution. The normalization / resolve step (`normalize_regime_phases`) for `Phased` is relatively local and early. Is there a reason why we cannot follow a similar design for this age-specialized feature? And further, wouldn't it be possible to pass the continuation grid nodes at runtime (similar to what we already do for `IrregSpacedGrid` with `pass_points_at_runtime`), forgoing the whole fingerprint stuff?

I must admit that I did not fully understand the entire implementation, so please excuse me if those are trivial questions.

mj023 (approved):
Good feature overall, could already be solved with regimes, but this makes it much easier for the user. I have nothing regarding the implementation, I think we should try to keep building everything we already know at model creation time and only build an object at runtime if it depends on model parameters. The only small change I would propose is mentioning the `AgeSpecializedGrids` in the documentation. Nothing long, as they are quite self-explanatory, but more as a hint for the user that such a feature exists.

hmgaudecker:
Thanks — both comments point to a useful boundary. Implemented in 485580e.

I agree that age specialization should follow the same normalization principle as `Phased`; the difference is only the scope. Phase normalization is local to one regime, while age normalization needs the model `AgeGrid` and each regime's active periods. So the order is now `normalize_regime_phases` first, followed immediately by a model-level `normalize_age_specialization` (new `age_normalization.py`). After that step, downstream processing sees only representative concrete objects plus period schedules — no public age-specialization factories.

Everything known at model creation is now built there. Every concrete age-specific function, grid, node array, and continuation interpolation object is constructed during model creation; solve, simulation, AOT, and diagnostics only *select* those prebuilt objects — they never call `build(age)` or construct grids. The existing runtime-points `IrregSpacedGrid` remains the separate path for grids that genuinely depend on model parameters, and an `AgeSpecializedGrid` that resolves to a runtime-points grid is now rejected at model creation.

Given that boundary, I did **not** add a general runtime continuation-grid protocol in this PR. Current-period nodes come from a precomputed `AgeGridSchedule`, and each continuation closure captures the prebuilt target grid for `t + 1`. The numeric fingerprint machinery (`_grid_identity` / `_node_fingerprint` / `_continuation_grid_signature`) is gone; program sharing is keyed on the explicit `AgeSpecializedGrid.signature(age)` contract. Passing precomputed grid values as JAX arguments remains a possible follow-up if benchmarks show per-signature compilation is too expensive, but it is not needed for correctness or normalization.

The age-specialization user guide now has an `AgeSpecializedGrid` section (moving-borrowing-floor example, the fixed-shape contract, model-creation-time resolution, and the distinction from parameter-dependent runtime irregular grids).

Full CPU suite: 1253 passed, 48 skipped; ruff and ty green.

PR #402, opened 2026-07-12
Fold IID utility shocks out of the stored value (quadrature-average at solve)

hmgaudecker:
## Fold IID utility shocks out of the stored value (quadrature-average at solve)

**Draft — stacked on `collective-regimes-draft`** (the collective-regimes extension). Review only the single commit on this branch; the base carries the extension.

### Motivation

An IID shock that enters the *current* period's utility (a wage draw, match quality, a pregnancy/taste innovation) is today materialized as a **stored grid axis of every value array**. With several such shocks the stored `V` picks up a multiplicative node factor (×27–81 in Eckstein–Keane–Lifshitz-scale collective models), which is what makes full-resolution solves intractable. But such a shock is only ever *integrated out* against its own quadrature weights — it need not be stored at all.

### What this PR does

Adds `fold: bool = False` to the IID process declaration. `fold=True` means "integrate this shock into the value at solve time; do not keep it as a stored axis." The shock is materialized only *through* the max-over-actions / collective readout, then averaged away by its own quadrature weights before the value is stored:

```
V_stored(x_det) = Σ_k w_k · [ max_{a∈Γ} Q(x_det, ε_k, a) ]        (single regime)
V^s_stored(x_det) = Σ_k w_k · [ Q^s(x_det, ε_k, a*(x_det,ε_k)) ]   (collective, per stakeholder)
```

This is **exact, not an approximation** — the same quadrature a stored shock would use, reduced one step earlier. The stored `V` loses the folded axis.

**Not the same as `ExtremeValueTasteShocks` (#375):** those fold a Gumbel shock on the *discrete-action* axis via a closed-form `logsumexp`. This folds a Normal/LogNormal shock on a *state-node* axis via numerical `jnp.average(weights=)`. Same *concept* (integrate out, don't store), different operator and different reduction axis — no shared code.

### Where the reduction lands (the load-bearing bit)

In `max_Q_over_a.py::_wrap_with_fold_reduction`: the fold axes are carried like co-map axes **through** the inner `productmap`/readout unchanged, then `jnp.average`'d out **after** the max/collective-readout and **before** any co-map `vmap` wrapping. Reduced from highest inner-position down (removing one never shifts a not-yet-reduced one). The collective **stakeholder axis is trailing on `V` only**, so it never collides with a folded state axis; the divorce flag `D` is folded with `jnp.any` to stay strictly boolean. `backward_induction._get_regime_V_shapes_and_shardings` drops folded axes from the stored-V shape/sharding; the continuation read is unchanged (the stored V already lacks the axis).

### Validation (rejects unfoldable declarations)

`fold=True` is rejected with a clear error when the shock is: not IID (AR(1) must be stored); read by a **same-period gate / value-constraint / reference projection** (E2/E3′ need unfolded per-node values — walks the gate/constraint DAG ancestors); **conditioned on by any next-period transition** (a folded shock is integrated out, nothing downstream may depend on its realization — walks the transition DAGs); combined with `taste_shocks`; on a non-GridSearch solver; or carrying runtime-supplied distribution params. A folded state that structurally persists into a reachable regime raises `ModelInitializationError`.

### Tests

`tests/regime_building/test_fold_iid_shocks.py` (10, written first):
- **Exactness oracle** — solve a model both ways (`fold=False` stored axis vs `fold=True`); assert value-identical (`assert_allclose`) with the folded `V` carrying exactly one fewer axis (single shock `(5,)`→`()`; 3-shock model drops 3 axes).
- Collective composition: per-stakeholder folded values equal the unfolded-then-averaged baseline.
- Validation rejections (same-period-gate read, AR(1), transition-conditioning, taste, non-GridSearch).
- Default path byte-identical (`assert_array_equal`).

### Verification

- Fold tests 10/10; `tests/regime_building` 109 passed.
- Broad `pytest tests/solution tests/simulation`: **570 passed, 5 skipped, 2 xfailed** — baseline exactly (default path byte-identical; everything gates on `fold=True`).
- `ty` (run in the main tree — the hook can't run in a git worktree): **All checks passed** (one real annotation fix included: `solvers.py` casts the folded grid to `_ContinuousStochasticProcess` before reading its transition probs).
- ruff check + ruff format: clean.

### DIFF
--- a/src/_lcm/engine.py
+++ b/src/_lcm/engine.py
@@ -644,6 +644,16 @@ class Regime:
     has_taste_shocks: bool = False
     """Whether the regime declares EV1 taste shocks on its discrete actions."""
 
+    fold_state_names: tuple[StateName, ...] = ()
+    """IID-process states declared `fold=True`, or empty (the default).
+
+    A folded state is integrated out of the stored value by quadrature at
+    solve time, so it is NOT an axis of the regime's stored `V`-array: the
+    backward-induction V topology (`_get_regime_V_shapes_and_shardings`)
+    excludes it from the shape/sharding it computes for this regime. Empty
+    keeps the default path byte-identical.
+    """
+
     certainty_equivalent: CertaintyEquivalent | None = None
     """Nonlinear certainty equivalent declared by the regime, if any."""
 
--- a/src/_lcm/processes/iid.py
+++ b/src/_lcm/processes/iid.py
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 from dataclasses import dataclass, fields
 from types import MappingProxyType
+from typing import ClassVar
 
 import jax
 import jax.numpy as jnp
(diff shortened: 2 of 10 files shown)


hmgaudecker:
Superseded by #409, which targets `main` and folds this branch's head into the audited collective-regimes tip. Closing.

PR #404, opened 2026-07-14
Fix four bugs in the mahler_yum_2024 example

hmgaudecker:
Four bugs in the `mahler_yum_2024` example, each verified against the authors' published Fortran. No pylcm API is touched — the diff is confined to `src/lcm_examples/mahler_yum_2024/__init__.py` plus its test baseline.

They surfaced in an adversarial audit of the `lcm-replications` wrapper (ChatGPT 5.6 Pro). Because this example *is* the model — `lcm-replications` imports it rather than copying it — the fixes have to land here.

## The fixes

**Income normalization.** `_compute_income_normalization` read `sigx` as an innovation *standard deviation* where it is a *variance*, and squared each log-normal mean correction twice — `exp(x⁴/2)` where `exp(x²/2)` is meant. The normalizer came out at **1.0037 instead of 1.3972**, i.e. the multiplicative wage shock was left essentially uncorrected. This is faithful to the Fortran (`sub_main.f90:251-252` has it character for character), but it is an authors' bug, and the estimated wage intercept cannot absorb it: the consumption floor, average earnings and the tax function are all in absolute model units. Now `cosh(θ)·exp(σ²/(2(1−ρ²)))`.

**Spline positivity.** A cubic through non-negative knots is not itself non-negative — the feasible knot vector `[0.01, 2, 0.01, 2]` dips below −0.3 in the interior, which would make work or effort a *good* and flip the sign of the choice. The estimator bounds the four knots, not the interpolated path, so it can walk straight into that region. The Fortran clips at zero; this did not. Now clamped.

**Pension.** The paper writes `P(e)` — education only. The Fortran additionally scales the pension base by the permanent productivity type θ (`sub_main.f90:487`); "median productivity shock" pins down `z`, not `θ`. Dropped.

**Initial habit.** `searchsorted` rounds up rather than to nearest, biasing every agent's `f₋₁` upward by up to one grid interval. The bias is one-sided, and `f₋₁` both drives the non-adjuster moments (`f₀ == f₋₁`) and enters the health logit with the *larger* effort coefficient (λ₂ = 0.734).

## This example implements the paper — the authors' Fortran does not

Recorded in the module docstring, because it determines what this example can be expected to reproduce.

The Fortran builds its continuation array at the *incoming* effort habit `f_{j−1}` and never rebuilds it inside the golden-section search over the effort choice (`sub_main.f90:464, 681-686`; `mod_solve.f90:171-186`), so its agent never internalises that today's effort becomes tomorrow's habit — contradicting the paper's own eq. (6), which discounts `E V_{j+1}(s_{j+1}, f_j)`. Its *simulation* meanwhile does advance the habit (`mod_simul.f90:634`), so its solve and its simulate are mutually inconsistent.

`next_lagged_effort` implements eq. (6). **Consequently this example must not be expected to reproduce the paper's published model column at the published Table-II estimates** — those estimates were fitted against the other recursion.

Two further defects in the authors' code are likewise not reproduced, and are documented rather than copied: the simulation reads survival at a stale education index (`mod_simul.f90:609`), and re-encodes the permanent-type index with the opposite convention from the one its policy arrays were solved under (`mod_simul.f90:365` vs `dimreduce3_new`), swapping β and η for four of eight types.

## Test baseline

`tests/test_mahler_yum_2024.py` pins ~20 behavioural moments, and all of them move. Re-frozen on an A100 (seed 32 / 42, n = 10 000, x64), same protocol as before. The structural invariants — all retired after `retirement_period`, zero income in retirement, non-negative wealth, positive consumption — still hold.


hmgaudecker:
NB: I did not update the fixtures yet, so the tests are failing ofc. @mj023 -- I will do so if you agree with the changes!

mj023 (approved):
Looks good. I knew only about the normalization and the pension bugs. A bit annoying that we can not follow either the paper, or the Fortran, in all cases. Nonetheless I think the decisions that have been made in each case are reasonable and well documented, so it's fine.

PR #405, opened 2026-07-14
State-conditioned shock parameters (stochastic volatility)

hmgaudecker:
## State-conditioned shock parameters (stochastic volatility)

Lets a continuous stochastic process condition its `sigma` on a discrete regime state:

```python
NormalIIDProcess(
    n_points=7, gauss_hermite=False, mu=0.0, n_std=3.0,
    sigma=0.30,  # the FIXED common node grid (widest regime)
    state_conditioned=StateConditioned(on="uncertainty", by={"low": 0.05, "high": 0.30}),
)
```

This expresses **regime-switching income risk / stochastic volatility** — a construct pylcm
could not previously represent. The motivating consumer is the uncertainty-shock panel of
Attanasio, Larkin, Ravn & Padula (2022) "(S)Cars and the Great Recession".

### Design (v1 scope)

- The scalar `sigma` defines a **fixed common node grid**; each regime's transition row is
  computed **directly at the from-value** with that regime's `sigma` — no interpolation of
  precomputed node rows.
- The conditioned `sigma` is gathered in **both** the solve kernel and the simulate next-state
  wrappers (via a single shared `sigma_array_by_code`), so the two phases cannot drift apart;
  a fixed non-zero `mu` is threaded through the conditioned row builders.
- Supported: CDF-binned `NormalIIDProcess` (`gauss_hermite=False`) and `TauchenAR1Process`
  (their transition CDF carries `sigma`).
- **Rejected at construction**: Gauss-Hermite IID and Tauchen (nodes scale with `sigma`) and
  Rouwenhorst (`rho`-only transition — fixing nodes erases `sigma`). Also requires a
  fully-specified, finite node grid, finite positive `sigma`, exact category keys, and one
  shared category→code map across every regime carrying the conditioning state.
- **Current-regime conditioning only** (the engine's product-of-marginals expectation).
- Least-invasive: one non-param field `state_conditioned` on the process base class; the
  existing scalar-`sigma` path is untouched.

### Tests

Regression tests derived from the audit counterexamples, each verified to fail on the pre-fix
code: kernel reduction to the stock Tauchen/NormalIID rows, analytic conditional moments, the
four construction-rejection guards, and — critically — **end-to-end `solve`+`simulate`** with
both equal-σ degeneracy and a distinct-σ response (a solve-only suite could not see the
simulate path). See `lcm_examples/stochastic_volatility.py`. Existing FGP and
runtime-process-parameter tests are unregressed.

### Certification

The design and the branch's implementation were both audited by an external
computational-methods review (ChatGPT 5.5 Pro); the final verdict is **clean**. The
generalizable lesson: a primitive that touches a process law must be tested *through*
`simulate`, not only `solve` — a correct solve kernel bolted to an unmodified simulation path
is invisible to a solve-only suite.

https://claude.ai/code/session_01DW9D1WhLN8JhhiEPCScbaw


timmens (commented):
I think the API would be clearer if conditioning were expressed on the parameter itself:

```python
NormalIIDProcess(
  ...
  sigma=StateConditioned(on="uncertainty", by={"low": 0.05, "high": 0.30})
  ...
)
```

This makes the conditioned parameter explicit, avoids the invalid sigma plus state_conditioned combination, and generalizes naturally to conditioning other parameters later.

Is there a reason this design wasn't chosen?

hmgaudecker:
> Is there a reason this design wasn't chosen?

Yes. operational blindness. On it!

Thanks a lot!

timmens (approved):
Very nice!

PR #406, opened 2026-07-14
Close the phase split: flow on truth, continuation on belief

hmgaudecker:
A simulated agent should act on what it believes and live in what is true. This
makes that hold for the continuation value, and makes perceived **risk**
expressible.

## What this adds

`Phased` state transitions accept `MarkovTransition` laws, so the kernel the
agent perceives and the kernel that generates the data can differ:

```python
state_transitions={
    "health": Phased(
        solve=MarkovTransition(health_believed),
        simulate=MarkovTransition(health_true),
    )
}
```

That is the natural home for perceived mortality, perceived health or income
risk, and policy rules the agent misreads. The two phases need not agree on
whether the law is stochastic — a deterministic law is a degenerate kernel — so
an agent may perceive risk where there is none, or treat as certain a transition
that is not.

## Which phase supplies which part of the decision

The simulated state-action value splits in two, along the agent's information
set at the moment of choice:

| part of $Q$ | phase | why |
|---|---|---|
| period utility, feasibility, the Koopmans aggregator | **simulate** | today's payoff and today's feasible set are known when the action is chosen |
| the continuation — next-period kernels, regime-transition probabilities, and every helper they read | **solve** | the future is only perceived, and $V_{t+1}$ was solved under those beliefs |

The realized draw follows the simulate laws, so the trajectory follows the truth
whatever was believed. In the vocabulary of the model class, `solve` supplies the
*solution* phase while `simulate` supplies both the *decision* phase and the
*realization* phase.

The agent is naive throughout: it never anticipates the divergence. A
sophisticated agent is a different object — its continuation is no longer the
solved $V_{t+1}$ — and is out of scope here.

Each half needs its function pool as well as its transitions, because a law's
argument names resolve against whichever pool it is handed, transitively.
Handing the continuation the solve transitions alone would still read `Phased`
helpers from the simulate phase.

## Defects this closes

Each was reproduced in a working model before being fixed, and each regression
test was confirmed to fail on the tree without the fix.

- **The continuation was priced under the true law.** The agent optimized against
  a truth it was not supposed to know. In the regression model the argmax
  reverses — `stay` becomes `switch`.
- **Period utility read a law belonging to neither phase.** `_get_U_and_F` took
  the outer `next_<state>` from the solve bundle but resolved its helpers from
  the simulate pool. A utility reading a chosen next state — the NEGM
  service-flow pattern — therefore valued its current flow under a rule the agent
  only believes.
- **The perceived law was never checked numerically.** Both `Phased` variants
  were collected under one key, so one record overwrote the other. A belief
  kernel returning `(1.2, -0.2)` solved clean, pricing backward induction with
  negative weights. Records are keyed per phase and failures name the phase.

The obvious fix for the first — overlaying names by their `weight_` prefix — is
wrong: `weight_` is not reserved, so a legitimate user helper named
`weight_bonus` would be silently swapped.

## What is now rejected

Constraints must be phase-invariant through their whole dependency chain, not
merely at the top level. A phase-specific feasible set would let the simulated
agent choose actions its value function was never computed for, so a constraint
that reaches a `Phased` helper or law of motion transitively is rejected when the
model is built.

The rule requiring both phases to agree on stochasticity is removed. Its
rationale — that the state would not otherwise have the same kind — does not
hold, since a deterministic law is a degenerate kernel. Both mixed directions are
covered by regressions.

The outer form `Phased(solve={target: ...}, simulate={target: ...})` is
normalized rather than reaching the engine as a raw dict. A per-target dict
inside `Phased` must be per-target in both phases and cover the same targets;
`Phased` is outermost-only, so this is the only spelling for a per-target law
that varies by phase.

## Compatibility

A model's numbers are unchanged **unless a `Phased` function lies in the
dependency ancestry of a continuation transition, or of a `next_<state>` read by
period utility or feasibility.** The solve phase is untouched, and for every
phase-invariant name both phases hold the same function.

Models that *do* have such a dependency change **by design** — the helper was
resolved from the wrong phase. That pattern was reachable before this branch
through a `Phased` helper under a phase-invariant law, so the correction can move
published results. Release-noted in `CHANGES.md`.

## Documentation

`docs/explanations/phase_grammar.ipynb` gains a section on which phase supplies
which part of the decision, and a worked perceived-health-risk example: a worker
who believes good health persists consumes more today than one who knows the
medical expense is coming, while both fall ill on schedule because the draw
follows the true kernel either way.

## Known limitation (pre-existing, not widened)

The runtime validator supplies arguments from grids and params, so a
`MarkovTransition` that reads a named DAG helper is warn-skipped. This affects
phase-invariant laws equally. Closing it needs parameter qualification reworked
for helper sub-DAGs — a separate change.

## Open question for a maintainer

In simulation, $Q$'s utility drives **only the argmax**: realized utility is
recomputed independently in `simulation/additional_targets.py` from the simulate
pool. That makes *flow = truth* versus *flow = belief* a modelling choice rather
than an implementation detail. This branch implements flow = truth, on the
information-timing argument above. Worth a second opinion before the contract
sets.

## Testing

Measured with this PR's changes applied, float64, deterministic order
(`-p no:randomly`), on a shared machine:

- Everything outside `tests/solution`: **1994 passed, 52 skipped, 5 xfailed**
  (`pytest tests --ignore=tests/solution -n 4`), pytest's own exit status 0.
- `prek run --all-files`: exit 0 — ruff, formatting, `ty`, mdformat, nbstripout.
- `docs/explanations/phase_grammar.ipynb` executes end to end, and the numbers in
  its prose are read off its own output.

**Six reproducible failures remain in `tests/solution`,** identical across two
independent whole-suite runs. They are **not** introduced by the documentation
commit at the tip of this branch: that commit changes only docstrings, comments,
and the wording of one error message — established by an AST comparison over all
14 edited files with every bare string statement stripped, which leaves exactly
that one executable difference. No test under `tests/solution` references the
message, and doctests are not enabled, so no docstring is ever executed.

Naming them is unfinished. Three attempts to run the directory to completion were
killed by the machine's shared memory ceiling or a wall-clock limit rather than by
the tests themselves; a 65-file slice of the 247 completed clean. Worth attributing
to the EGM/envelope work these tests cover before this merges.

https://claude.ai/code/session_01FtWM7QWAuS8aBiCCLJcqBL


src/_lcm/regime_building/Q_and_F.py
```diff
@@ -69,16 +69,35 @@ def get_Q_and_F(
...
+    continuation_functions: EconFunctionsMapping | None = None,
 ) -> QAndFFunction:
     """Get the state-action (Q) and feasibility (F) function for a non-terminal period.
 
     `age` and `period` are runtime arguments (via `**states_actions_params`),
     not closure constants. This allows periods with the same target
     configuration to share a single JIT-compiled function.
 
+    Q mixes two phases when it is built for the simulate phase: the *current* flow
+    (utility, feasibility, `H`) is simulate-phase, while the *continuation* is priced
```
timmens, line 81: Isn't `H` called koopmans aggregator or something like that now?

timmens (approved):
Very very nice!

PR #407, opened 2026-07-20
Inference-grade continuous outer choice (GSS) for NNBEGM 

hmgaudecker:
Implements continuous outer choice for `NNBEGM`. In economic terms, this changes how one continuous adjustment margin is solved; it does not change the model's preferences, constraints, or transition equations. The finite-grid method remains available.

## Economic example

Consider the Mahler–Yum model. A household enters the period with wealth and last period's effort habit, say 0.40. It can:

1. keep effort at 0.40 without paying an adjustment cost; or
2. pay an observed adjustment cost, choose a new effort level anywhere between 0 and 1, and then choose consumption and saving conditional on that effort.

With a 17-point effort grid, the adjuster can choose only 0, 0.0625, 0.125, and so on. If its economic optimum is 0.43, the finite solver must publish a nearby grid point such as 0.4375. The continuous solver instead evaluates exact conditional consumption-saving problems at selected effort nodes and adaptively refines promising intervals between them.

There are two distinct simulation cases:

- With a deterministic keeper-versus-adjuster comparison, simulation interpolates the conditional policies at each subject's actual state and reruns the safeguarded outer search. Separately solved values can be replayed with the exact policies returned by `solve(return_simulation_policy=True)`; supplying NNBEGM values without their matching policy fails closed.
- With Mahler–Yum's observed fixed adjustment cost, solution integrates the uniform cost analytically. A simulated subject still needs a realized draw to determine whether it keeps or adjusts. Until that draw-and-replay channel exists, `simulate()` now fails before an automatic solve with a dedicated unsupported-operation error; it cannot silently substitute the ordinary-grid decision.

The analytic integration supplies the expected value and exact adjustment probability at each solved state. It does not replace the subject-level cost realization needed for path-dependent simulation outcomes.

## Benefits and costs

Benefits:

- Less grid-snapping error in the solved outer action.
- Smoother policy responses for comparative statics, moments, and—in smooth regions—derivatives.
- The inner consumption-saving problem retains NB-EGM's exact treatment of declared kinks, jumps, and hard constraints.
- A supported uniform observed adjustment cost is integrated analytically during solution, avoiding another solve-state dimension and yielding the adjustment probability in closed form.
- Numerical diagnostics expose approximation error, bound solutions, close competitors, and fallbacks.

Costs and limitations:

- Continuous search requires more conditional solves, compilation, memory, and runtime than a fixed outer grid. Large models need batching or streaming.
- Results depend on mesh and tolerance settings; difficult value surfaces may exhaust the node budget.
- `fail_closed=True` covers mesh convergence specifically: an exhausted node or round budget with marked intervals outstanding raises. Bound optima, close branch or runner-up margins, policy fallbacks, and all-invalid cells remain diagnostics; a broader inference release gate is the caller's responsibility.
- The analytic adjustment-cost fold currently supports only the declared uniform observed-cost specification and validates every evaluated scale as finite and nonnegative.
- Fixed-cost lifecycle simulation still needs cost draws and contingent keeper/adjuster replay; `simulate()` explicitly refuses this configuration in the meantime.
- The continuously replayed outer proposal interpolates the inner policy between solved outer nodes. Canonical rescoring must agree one-sided with the policy surrogate's winning value within the configured mesh absolute/relative tolerances; otherwise simulation emits the safe baseline and records `nested_policy_fallback`.
- Implicit derivatives are available only for resolved, smooth, interior optima.
- `FiniteOuterGrid` remains the simpler choice for quick runs, historical comparisons, and debugging.

## User-facing API

Import the solver and its strategies from `lcm.solvers`.

- `NNBEGM(..., outer_search=...)` requires an explicit outer-search strategy.
- `FiniteOuterGrid(...)` preserves finite candidate-grid behavior.
- `AdaptiveOuterMesh(...)` performs the continuous-outer approximation.
- `DeterministicOuterMaximum()` keeps the hard keeper/adjuster maximum and gives the keeper exact ties.
- `UniformObservedFixedCost(...)` analytically integrates a uniform fixed adjustment cost during solution. It requires `AdaptiveOuterMesh`, publishes the analytic adjustment probability, and rejects negative, NaN, or infinite evaluated scales before solving.

The branch-aggregation slot is a closed set, not a polymorphic extension point: `NNBEGM` rejects any `OuterBranchAggregator` subclass other than the two implemented folds. Inert outer-search options and the unwired `LegacyGoldenSection` configuration have been removed.

## Numerical contract

The continuous path materializes the exact conditional NB-EGM solution at every outer mesh node, constructs a shared validated cubic-Hermite surrogate, identifies every node-local candidate bracket, and refines each bracket with safeguarded golden-section maximization. It does not assume global unimodality.

Exact mesh nodes remain in the candidate set, so refinement cannot lose the finite-grid winner. Candidate selection retains deterministic keeper/adjuster, tie, NaN, and fallback semantics. The numerical audit covers exponent extremes, subnormals, blocking, and both precisions. Only the framed-arithmetic helpers used by the live NB-EGM path remain in production; the parked 1,458-line alternative envelope implementation and its private tests were removed.

The solver publishes `SolverDiagnostics`, including normalized interpolation error, final bracket width, mesh size, bound masks, branch and runner-up margins, fallback masks, unresolved cells, and analytic adjustment probabilities where applicable.

For deterministic hard-maximum models in the supported reader scope, the nested simulation payload carries the outer candidate bank and conditional inner policies. Simulation reconstructs candidate actions at the subject's actual state, reruns the safeguarded outer search, and scores emitted pairs under canonical Q. A proposal is accepted only when feasible, weakly better than the admissible baseline, and not materially below the policy surrogate's winning value under the configured absolute/relative value band. Otherwise simulation emits the canonically rescored baseline and records `nested_policy_fallback`.

The baseline it has to beat is the best action the subject could actually be given, not one privileged candidate. Where the solved policy cannot be reproduced at a subject's own state, the fallback ranks the raw grid pair, the no-adjustment keeper, and every published outer node under the *same* canonical Q — flow payoff plus discounted continuation at that subject's state — and emits the highest survivor. Infeasible and non-finite branches are excluded before the comparison, not after it; each branch carries its own conditional inner policy, so an outer node is never paired with another node's consumption rule; ties are broken deterministically, with the grid pair taking an exact tie and the keeper preceding ascending node order within the bank. If nothing survives, simulation raises before emitting an action rather than publishing an infeasible one. The values the solve published for those branches are not consulted at this point: they ranked branches at the solved state, which is not the state being simulated.

One consequence is worth stating plainly, because it changes simulated behaviour rather than only internals. A maximum over candidates is weakly higher than the grid pair alone, so the interpolated proposal is weakly harder to accept and weakly more subjects are served by the fallback. The direction follows from the acceptance inequality; the size of the effect on any particular model's moments is not characterised here.

Both outer searches recover the outer action by inverting a *declared* affine map whose
coefficient is certified from the map's structure, and admit the result through one shared
tolerance-free predicate: the forward image must be finite and inside the published inverse
domain, and where the target is a declared endpoint the image must reproduce it bit-for-bit.
There is no residual tolerance and no slope probe on either route, so solve and simulation
cannot gate on different predicates. An outer grid reaching outside the outer state's
declared domain is refused at build time. A candidate the predicate refuses is replaced by the
best admissible branch of the published bank — another outer node, the keeper, or the
grid pair — rather than by a stock displaced from the endpoint.

The optional implicit-derivative primitive uses a mesh-plus-golden-section primal and an implicit-function custom JVP only for resolved smooth interior optima. Boundary solutions, flat curvature, competing basins, and nonstationary/kink optima are marked unresolved rather than assigned a misleading derivative.

## Mahler–Yum integration

`create_mahler_yum_model(implementation="paper")` uses:

- continuous effort as the NNBEGM outer action;
- analytic integration of the observed fixed adjustment cost during solution;
- consumption as the inner Euler action;
- retirement feasibility in flow utility; and
- a declared minimum-resource floor.

Two limitations remain explicit:

1. Paper mode does not yet draw the analytically integrated adjustment cost during simulation and replay the contingent keeper/adjuster policy. `simulate()` raises a dedicated unsupported-operation error for this configuration until that channel exists.
2. The authors' Fortran uses `c = max(cash_on_hand - saving, min_consumption)`. Once its floor binds, every additional dollar saved increases the implied government transfer dollar-for-dollar; with no asset test, the household may preserve all its own resources while government finances the entire consumption floor. The declared pylcm model instead lets consumption and saving divide `max(cash_on_hand, min_consumption)`. The implementations therefore differ only in below-floor states, and the difference is documented rather than hidden.

The incomplete `legacy_fortran` compatibility layer has been deleted. The factory now offers only the paper-equation model and the brute-force oracle; it does not expose switches that cannot truthfully describe the constructed model.

## Scope

This PR contains the continuous-outer solver, deterministic nested-policy simulation reader, diagnostics, and paper-equation Mahler–Yum model. Fixed-cost draw-and-replay remains unimplemented but is now a fail-closed unsupported simulation operation; materially inconsistent interpolated inner policies are controlled by the configured value band and explicit safe fallback. Moment construction, parameter transforms, standard errors, estimation release gates, and published-number decomposition remain outside this PR.

## Verification

Head: `a5c6fe4fc1c70354cf16efd341c419a8e8975fb7`.

- Fallback selection, `tests/simulation/test_outer_endpoint_admission.py`: 37 tests, **0
  failures and 0 errors** at `--precision=64` and at `--precision=32`. The module is a
  discriminating instrument, not merely a passing one: run against the previous selection
  rule with the same tests, 5 of the 37 fail at both precisions.
- Resource cost of ranking every candidate rather than one, measured on the adaptive
  workload with the two rules alternated across three blocks per precision so that machine
  drift cancels within a block: warm simulate **+0.5%** (f64) and **+0.1%** (f32),
  best-of-7; cold solve+simulate within [-0.6%, +1.5%]; peak resident memory within
  [-0.7%, +2.2%]. All inside the 5% warm and 10% compile/memory review thresholds.
  Alternating mesh shapes reuse compiled programs: the compilation-cache entry count is
  flat across eight repetitions on both rules.
- `prek run --all-files` exit 0: 34 hooks run, 34 passed, 0 failed, 0 skipped, including
  Ruff, formatting, `codespell`, and `ty` 0.0.75. The 35th declared hook, `pixi-lock-check`,
  runs at the pre-push stage.

Not claimed here: the full suite at this head. CI covers it — the `cpu` context is green
only if all of the fp64 leg, the fp32 leg, and three sharded slow-solution jobs at both
precisions succeed, alongside macOS and Windows platform legs and the exact-kernel
capability matrix — and its result on this head is what should be read, not a number
transcribed into this description.

## API change for downstream branches

`NNBEGM` no longer accepts `branch_aggregator`; it now carries exactly
`{inner, outer_search}`. The adjustment cost is declared on the margin instead:

```python
OuterContinuousMargin(
    ...,
    adjustment_cost=UniformObservedFixedCost(...),
)
```

Leaving `adjustment_cost` unset means adjusting is free at the margin, i.e. the
deterministic maximum. `UniformObservedFixedCost` still requires
`outer_search=AdaptiveOuterMesh(...)`, now reported against the regime that
declared it rather than against the solver.

Nothing on `main` is affected — `branch_aggregator` was introduced in this PR and
never existed on `main`. Branches stacked on this one inherit the migration with
their next merge; any `NNBEGM(branch_aggregator=...)` written on such a branch
**before** taking that merge becomes a `TypeError`.


hmgaudecker:
Status note: the base branch moved while this PR was being assembled — `feat/nested-nbegm-ez` merged the dcegm solver-boundary refactor (round-15 repairs) at 03:16, which **splits `solvers.py` into per-solver modules** (`nbegm.py`, `nnbegm.py`, …). This PR's 18 commits modify the pre-split `solvers.py`, so the PR currently shows conflicts.

The port is mapped and tractable: ~907 diff-lines of solver delta to re-route into the new modules (everything else — the new `_lcm/egm/` modules, `lcm_examples` paper mode, tests — should merge clean). Plan: a single `git merge` conflict session rather than a rebase, then the frozen-baseline neutrality test, the NBEGM/NNBEGM batteries, the Mahler merge gates, and the full CPU suite. The base merge should also fix the two pre-existing `test_n_nbegm` failures documented in the PR body — they are expected to flip green after the port.

timmens (commented):
## A. Merge-blocking output correctness

### 1. Return and consume one complete solution artifact

`Model.solve()` returns value arrays or an optional `(values, policies)` tuple, while
`Model.simulate()` accepts only the values. The two standard workflows therefore
produce different continuous-NNBEGM actions. With the same toy model, parameters,
initial conditions, and seed, automatic simulation emits continuous outer actions;
passing separately solved values returns grid actions. The first subject changes from
`-1.156151` to `-20.0`, the maximum difference is `34.0`, and the supplied-values
result has no fallback column.

Could `solve()` return a `SolutionResult` carrying value functions, simulation
policies, and solver diagnostics, with `simulate()` consuming that object? An
equivalent interface is fine, provided a separately solved model can replay the same
policy and expose the same diagnostics as an automatic solve.

### 2. Control or expose conditional inner-policy error in simulation

Nested simulation continuously optimizes the outer-value surrogate and linearly
interpolates the associated inner policies across outer nodes. Canonical rescoring
checks feasibility and dominance over a grid baseline; it does not verify that the
emitted inner action is conditionally optimal.

An end-to-end reproduction emits outer action `4.592250281` and consumption
`1.856252064`. Holding the outer action fixed, dense maximization of the same canonical
Q gives consumption `1.718175495` and improves value by `0.007090397`. The emitted pair
is feasible and reports `nested_policy_fallback=False`.

Could simulation recover the conditional inner optimum, or publish a meaningful bound
on this policy error? The PR description should also avoid saying that simulation
reoptimizes the inner action while the implementation scores the interpolated pair.

### 3. Refuse or implement fixed-cost simulation

The solution kernel integrates the observed adjustment-cost draw. Simulation draws no
such shock and deliberately withholds the nested policy whenever the fixed-cost
aggregator is active, leaving the ordinary grid action path. In an end-to-end
two-period toy, changing the fixed-cost scale from `0.01` to `100` changes 109 of 120
solved-value cells, with a maximum change of `28.6581`. Across the same 120 initial
states, zero simulated outer actions and zero inner actions change, and neither result
has a fallback column. For intuition, keeper value `10`, gross adjuster value `12`,
and cost `4 * chi` with `chi ~ Uniform(0,1)` imply a 50% adjustment probability.

Could fixed-cost configurations either draw the shock and replay the contingent
keeper/adjuster policy, or raise a clear unsupported-operation error from `simulate()`?

### 4. Validate the fixed-cost scale before applying the closed form

`UniformObservedFixedCost` requires a finite nonnegative scale `B`, while the evaluated
`scale_function` output is unchecked. The kernel treats every nonpositive value as its
zero-scale branch. With `V_K=1`, `V_A=0.5`, `B=-2`, and `chi ~ Uniform(0,1)`, it returns
value `1.0` and adjustment probability `0.0`; direct integration gives approximately
`1.5625` and `0.75`.

Could the parameter preflight reject negative, NaN, or infinite scale values for every
relevant period?

## B. Merge-blocking truthful configuration

### 5. Remove or implement inert outer-search options

Several newly public settings are accepted without changing the computation:

- `local_refiner="quadratic"` still runs golden-section refinement.
- `evaluate_all_endpoints`, `retain_second_best`, and
  `freeze_mesh_for_derivatives` have no source readers.
- `LegacyGoldenSection` is publicly exported, then rejected during NNBEGM kernel
  construction because only `FiniteOuterGrid` and `AdaptiveOuterMesh` are wired.

Could these options be implemented and tested, or removed from the public API until
their behavior exists?

### 6. Make compatibility manifests match the constructed model

`MahlerLegacyCompatibility` exposes independent switches and
`single_switch_variants()` promises switch-by-switch decomposition. In practice,
configurations enabling only historical effort search, only saving search, only chi
nodes, or only the VSLY measurement all return the same complete brute-force model.
The manifest can therefore say `historical_partial:historical_effort_search` while the
model also uses historical saving and chi discretization.

Could the factory construct each advertised partial configuration independently? If
only the bundled brute configuration is representable, partial combinations should be
rejected and the bundle should be exposed as one supported configuration.

## C. Non-blocking scientific and API decisions

### 7. Disclose the consumption-floor approximation at the public entry point

`implementation="paper"` changes the feasible set below the minimum-consumption floor.
With own resources `0.50` and floor `1.00`, the Fortran rule permits consumption `1.00`
plus positive saving from own resources. The new floored-cash-on-hand budget requires
consumption and saving to divide `1.00`, so consumption `1.00` implies zero saving.

The module documents this limitation clearly. Could the public
`create_mahler_yum_model()` docstring also disclose the structural approximation? Its
current description of `implementation="paper"` as canonical can otherwise imply exact
reproduction of the paper or Fortran feasible set.

### 8. Derive branch aggregation from the economic specification

Could we discuss why `branch_aggregator` is a public field on `NNBEGM`? Its two current
choices are `DeterministicOuterMaximum` and `UniformObservedFixedCost`. The appropriate
choice follows from the economics: the absence of an adjustment shock implies the hard
maximum, while an observed uniform fixed cost implies the analytical expectation over
that shock. Declaring the fixed cost, distribution, timing, and scale on
`OuterContinuousMargin` would let the engine derive the branch operation and let the
solver advertise whether it supports the required integration.

The public extension contract also appears misleading. `OuterBranchAggregator` says a
nested solver accepts any concrete aggregator and dispatches polymorphically, while
`_resolve_branch_fixed_cost()` recognizes only `UniformObservedFixedCost`; every other
subclass is silently handled as the deterministic maximum. If custom aggregation is an
intended extension point, it needs real dispatch or an explicit unsupported-type error.
If it is an internal implementation detail, the public base class and solver field seem
unnecessary. This can remain a non-blocking API-design discussion.

### 9. Align `fail_closed` with its implemented guarantee

`AdaptiveOuterMesh.fail_closed` raises only when the node or round budget is exhausted
with marked intervals remaining. Bound optima, close branch or runner-up margins,
policy fallbacks, and invalid cells remain diagnostics. The PR description currently
says `fail_closed=True` may raise for several of those conditions.

Could the PR description state the narrower mesh-convergence guarantee? Broader
inference release gates can be designed separately.

## D. Non-blocking scope and cross-cutting effects

### 10. Explain the cross-solver benchmark regressions

The current benchmark report flags existing workflows that do not use continuous
NNBEGM: ACA execution `+23%`, precautionary-savings simulation `+15%`, and Iskhakov
GridSearch solve and simulation `+14%` and `+26%`. These deserve attribution because
the feature should ideally impose no runtime cost on unrelated solvers.

One candidate for the simulation regressions is the new all-false
`nested_policy_fallback` subject vector, which is allocated, stored, concatenated, and
synchronized for every regime-period even when no nested policy exists. This is a
hypothesis and does not explain the GridSearch solve result. Could the alerts be
reproduced and either attributed or tracked for follow-up?

### 11. Move the parked envelope implementation out of production code

The PR adds 1,461 source lines for `_envelope_at_query_407`. It has no caller, the PR
describes it as parked audit/reference code, and its name embeds the PR number in a
production numerical module. Keeping a second envelope implementation beside the live
one creates a substantial maintenance and drift surface.

Could the parked implementation move to a test/reference location or a follow-up? The
small arithmetic helpers imported by `nbegm_step.py` can remain in a focused production
utility.

## E. Separate project-level discussion

### 12. Define a lifecycle for expensive model-level tests

The new `manual` marker keeps paper-scale tests out of routine CI and leaves open when
the project expects them to run. This PR does not need to settle that policy.

Could we discuss whether expensive suites should run for release candidates, through
maintainer-triggered PR profiles, or through both routes? A defined lifecycle would
make these model-level gates useful without imposing their cost on every PR.

timmens (commented):
## Blocking

### Accept the continuous NNBEGM replay policy returned by `solve()`

I reproduced this on the reviewed head. With `AdaptiveOuterMesh`, automatic simulation
succeeds and `solve(return_simulation_policy=True)` returns a
`NestedEGMSimPolicy`. Passing that exact `(values, policies)` pair to `simulate()`
raises `InvalidSimulationInputError`, because
`_check_supplied_replay_policies()` accepts only `NNBEGMSimPolicy`. Automatic
solve-and-simulate and the equivalent documented split workflow therefore behave
differently.

Could the validation derive the expected replay type from the configured outer search,
accepting `NestedEGMSimPolicy` for `AdaptiveOuterMesh` and `NNBEGMSimPolicy` for the
finite search? A regression test could assert that automatic and separate simulation
produce identical results for both routes. Issue #428 can later bundle values and
policies in a `SolutionResult`. The currently documented interface still needs to
accept the artifact it returns.

## Nonblocking

### Clarify or complete reverse-mode support for the exact-affine derivative rule

I also reproduced this on the reviewed head. For the affine function `V(x) = 2x`, the
new exact-affine custom JVP returns the correct derivative `2.0` through `jax.jacfwd`.
The same scalar interpolation raises `AssertionError` under `jax.grad`.

Could we either make the derivative rule transposable and add a `jax.grad` regression
test, or explicitly document exact-affine differentiation as forward-mode-only? The
forward-only contract would cover the current implicit pilot. `jax.grad` remains a
natural expectation for scalar objectives.

### Should branch aggregation be derived from the economic specification?

Thanks for making `OuterBranchAggregator` an explicit closed set and rejecting
unsupported subclasses. That resolves the silent-dispatch concern.

I still wonder whether `branch_aggregator` belongs on `NNBEGM`.
`UniformObservedFixedCost` describes economic structure: the adjustment-cost
distribution, its timing, and how it enters the keeper-versus-adjuster comparison.
Could this be declared on `OuterContinuousMargin`, or another economic-specification
object, with the solver only declaring whether it supports the resulting aggregation?

For example:

```python
OuterContinuousMargin(
    ...,
    adjustment_cost=UniformObservedFixedCost(...),
)
```

The absence of an adjustment cost would imply the deterministic maximum. This would
keep `NNBEGM` focused on numerical choices such as the inner solver and outer-search
strategy, and the engine would derive the branch operation from the model.

## Low priority, fully nonblocking

### Clarify what the Mahler–Yum implicit-derivative pilot currently demonstrates

This has no effect on ordinary solution or simulation and should not hold up the PR.
`capture_pilot_problem()` leaves `owner_provenance` unset, while `run_pilot()` requires
an owner certificate. Consequently, every Mahler–Yum cell is marked unresolved because
the provenance is missing. The manual test accepts that outcome, so it currently
demonstrates safe refusal and provides no real-model evidence for a certified implicit
derivative or a model-induced kink. Could the pilot state this limitation explicitly,
for example by skipping with an unsupported-provenance reason until the required owner
information is available?

hmgaudecker:
Thanks — all four are addressed. Current state at `1851326a`, item by item.

## Reverse-mode exact-affine

Forward-mode only, deliberately. On finite primals and finite directions the affine
differential is linear and has a mathematical transpose; the rule also inspects tangent
finiteness, so JAX cannot derive that transpose automatically, and no separate reverse
rule is registered. Reverse mode is therefore **unavailable, not impossible** — reachable
by writing an explicit rule with its own non-finite-cotangent contract, without touching
the direction guard.

The guard stays: `test_nonfinite_direction_fails_closed_without_changing_primal` pins that
an infinite direction yields a NaN tangent, which is the behaviour I want on an
inference-grade path. That is a preference about which rule to write, not a claim that
none exists. If you want `jax.grad` on the certified path, it is a real piece of work with
a contract to settle — say so and I will scope it.

`f894a409` adds a `no-impossible-transpose` pre-commit hook so the phrasing cannot
reappear: 6/6 on synthetic positives, 4/4 on the historical instances, no false positives
on the tree.

`_scaled_bits_jvp` in `probability.py` is the same shape: a bitcast is not an obstruction
to transposition there either. The bit-level and multiply spellings of that tangent agree
to `max|diff| = 0.0`, and JAX's actual complaint is `Transpose rule ... not implemented`.

## Blocking — replay policy

Fixed test-first in `04373292`. The expected replay type follows the configured outer
search, recorded as `NNBEGMPolicyRead.replay_policy_is_nested`.

**I did not use the symmetric per-cell rule you proposed, and I think a symmetric rule
would be wrong here.** `backward_induction.py` stores a policy only when it is not `None`,
and the adaptive kernel publishes one only when `nested_published` holds. So under
`AdaptiveOuterMesh` a solved `(period, regime)` cell can legitimately have *no* entry, and
`simulate()` falls back there on the automatic route too. Requiring an entry per cell would
make the split workflow stricter than the route it has to reproduce. The implemented rule:

- finite outer grid ⇒ `NNBEGMSimPolicy` required per solved cell (the finite solve always
  publishes one, so absence is a caller error);
- adaptive outer mesh ⇒ a present entry must be `NestedEGMSimPolicy`; an absent entry is
  accepted;
- a wholly absent `policies` mapping ⇒ refused on both routes.

The known cost: a caller who hand-trims a *partial* mapping gets grid-argmax fallback on
the dropped cells with no error, because the check cannot separate "solve did not publish
here" from "the caller dropped it" without solving. If you would rather have strictness
there, say so and I will take the parity hit instead.

`tests/simulation/test_nnbegm_split_workflow_parity.py` asserts your regression test —
automatic vs separate simulation identical for both routes — plus two guards that force the
check to raise, since widening a validator can silently turn it into a no-op.

## Nonblocking — branch aggregation from the economic specification

Implemented as you proposed. `NNBEGM` carries exactly `{inner, outer_search}`;
`OuterContinuousMargin` gains `adjustment_cost`, with unset meaning the deterministic
maximum. The closed-set check lives in `_lcm/egm/branch_aggregation.py` and fires where the
cost is declared. Zero migration burden: `branch_aggregator` never existed on `main`.

Declaring the cost at the margin means a regime could name an adjustment cost the chosen
solver would never execute — `GridSearch` and `NEGM` substitute the hard maximum, so a
declared economic object would disappear at binding. `408f418f` refuses that combination at
construction and names the regime, the solver, and the cost. It is written against the
*class* — any non-trivial `OuterBranchAggregator` under any solver that cannot run it —
rather than against the `UniformObservedFixedCost` witness, and an explicitly declared
`DeterministicOuterMaximum` is accepted everywhere, since naming the default changes
nothing.

`consumption_savings_regime` binds the solvers module rather than importing
`OuterBranchAggregator` by name, matching the `import lcm.solvers as _solvers` idiom
already in that file; a `TYPE_CHECKING` import is unavailable, since PEP 649 plus the
beartype claw needs the annotation to resolve at construction.

## Low priority — the Mahler–Yum pilot

The per-cell test skips with an explicit unsupported-provenance reason (`a2229d04`), and
`capture_pilot_problem` documents that it builds no witness and why. I verified the skip
fires with a synthetic report. `test_real_model_kink_is_materially_nonstationary` carries
real-model content — its screen does not consult the witness.

## The outer endpoint class

Both outer searches recover the outer action by inverting a **declared** affine map whose
coefficient is certified from the map's structure, and admit the result through one shared
tolerance-free predicate: the forward image must be finite and inside the published inverse
domain, and where the target is a declared endpoint the image must reproduce it
bit-for-bit. There is no residual tolerance and no slope probe on either route, so solve
and simulation cannot gate on different predicates. One `OuterReplayCapability` is resolved
per regime-period before either search runs and carried on both policy payloads as a
required field with no default, so construction without it is a `TypeError`. An outer grid
reaching outside the outer state's declared domain is refused at build time. A candidate
the predicate refuses falls back to the grid argmax.

Two things worth having explicitly, since neither is obvious from the diff:

**The class is not float32-specific.** A signed-domain witness fires at float64 as well,
displaced `2.22e-16` (2⁻⁵²) against `1.19e-7` (2⁻²³) at float32; a scale-relative gate of
`1e-6` admits both. Float32 is where the displacement is large enough to reach a reported
quantity, not where the misclassification lives.

**It requires an outer domain that straddles zero.** Recovering the action is exact
whenever the stock and the target share a sign, because the subtraction stays inside the
operands' own binades. On a non-negative outer state reached through a direct law the class
cannot arise, which is why the accept-side controls carry as much weight as the refusals: a
replay that dropped every candidate would satisfy the refusals on its own.

Evidence, all reproduced locally before being relied on:

- 19 tests over both declared endpoints × both outer searches × {refuse, publish}, plus
  interior and non-negative direct-law controls and a cross-workflow parity pair. Witness
  stocks are derived at the working dtype, so each is genuine at float32 and float64
  rather than pinned to one.
- Exhaustive float32 characterisation of the class: 2,097,152 scale-admitted non-exact
  acceptances per binade at a signed lower endpoint, and 0 in the non-negative control.
- Candidate-set error between a scale-relative rule and the shared predicate: `0.0` across
  every published numeric column, at both precisions, on an ordinary non-negative model.

## Verification

- Full suite at `efc6e05e`: 5948 testcases at `--precision=64` and 5913 at
  `--precision=32`, 0 failures and 0 errors in both. Counts are `<testcase>` element
  counts from the junit XML, not console text or root attributes.
- Endpoint admission tests: 19/19 at both precisions.
- Resource effect of the shared predicate, medians of 7 warm reps, admission rule swapped
  in place so the comparison isolates the predicate: warm simulate −0.6% (f64) and −1.7%
  (f32); cold solve+simulate +0.5% and −4.2%; peak RSS +0.16% and −0.33%. Inside the 5%
  warm / 10% compile and memory thresholds.
- `prek run --all-files` exit 0 at `1851326a`: 30 hooks, 30 passed, 0 failed, 0 skipped.

Not measured, so not claimed: the full suite at `1851326a` itself, which adds a type
annotation and a hook/lock refresh on top of `efc6e05e` — CI covers it. GPU, macOS, and
Windows likewise.

One open measurement I would rather name than paper over: the candidate-set error above is
`0.0` on a non-negative outer domain, which is what current models run. The value-level
effect on a *signed*-domain model is unmeasured — it needs a model whose outer domain
straddles zero and in which a subject lands on a declared endpoint during an ordinary
simulation, which no fixture here constructs.

timmens (approved):
# PR 407 review comments

Reviewed against `a5c6fe4fc`.

Approving with one explicit merge condition: please fix the keeper-domain issue below.
Once that blocker is addressed and covered by a regression test, this is ready to merge
from my side without another review round.

## Blocking: outer-policy replay correctness

### Preserve the exact keeper outside a narrower adjuster mesh

`AdaptiveOuterMesh` solves the keeper as a separate exact branch, and the current
validation explicitly permits the adjuster mesh to lie strictly inside the outer
state's domain. Replay then applies the mesh endpoints to the keeper too. A keeper
outside that narrower adjuster mesh is removed before the canonical fallback ranks the
bank.

I reproduced this with the ordinary outer-state domain `[0, 20]`, adaptive adjuster
mesh `[2, 18]`, and a subject at outer stock `0`. Terminal utility strictly prefers the
lower stock, so the solve chooses no adjustment and reports `-8.230213`. Replay rejects
that keeper solely because `0` lies below the adjuster mesh, sets
`nested_policy_fallback=True`, emits investment `2` and stock `2`, and reports
`-18.415411`.

Could the keeper row be checked against the outer state's domain while adjuster rows
remain checked against the mesh? Requiring the adaptive mesh to span the full state
domain would also make the two phases agree, though preserving the separately solved
keeper seems closer to the current solver contract. An end-to-end regression should
cover a current stock outside a legal, narrower adjuster mesh.

## Nonblocking: numerical robustness

### Bound the round-trip error for interior outer targets

I reproduced an end-to-end solve/simulation disagreement on this head. Consider the
declared unit-slope map `new_outer = outer + outer_investment`, an outer state/search
grid `(0.0, 0.5, 1e10)`, and a subject at `outer=1e10`. The solve ranks the interior
target `0.5` and reports a Bellman value of `0.693147`. In float32, inversion recovers
the action `-1e10`, whose forward image is `0.0`. The current predicate admits that
candidate because `0.0` lies somewhere inside the declared state domain. Simulation's
canonical rescoring consequently emits `new_outer=0.0` and reports `0.455647`.

This means the conditional problem is solved at one stock and replayed at a materially
different stock. Domain containment does not bound the interior approximation, and the
distance can be arbitrarily larger than the target's local ULP. Canonical rescoring
reports the value of the emitted action honestly, though it cannot repair the Bellman
value already computed at the nominal target during backward induction.

This requires an unusually wide and unevenly scaled state domain, so I do not consider
it merge-blocking. Could interior admission eventually enforce a target-local
round-trip bound, for example exact or adjacent-representable reproduction, and refuse
candidates outside that bound? An end-to-end regression with the values above would
pin that a published target and its replayed stock remain numerically associated.

PR #401, opened 2026-07-12
Dissolution routing: revert each individual to its own single regime (synthetic mode)

hmgaudecker:
## Dissolution routing: revert each individual to its own single regime (synthetic mode)

**Draft — stacked on the collective-regimes extension.** Review the row-split commit; the base carries the extension.

> Naming: this routes an individual to its own single regime when a **collective/matched regime dissolves** (the general `D` dissolution flag — empty feasible set). "Dissolution" is the engine term; in a marriage model (e.g. EKL) the instance is divorce.

### Motivation

The collective-regimes extension's forward simulation is a fixed-size pass, so on **dissolution** a household row could not become two independently-tracked single rows — the continuing membership followed the *first declared* edge leg regardless of which partner the row represents. Wrong for any two-sided dissolution simulation.

The immediate driver is **Eckstein–Keane–Lifshitz (2019)** (`ecksteinCareerFamilyDecisions2019`). EKL's simulation (Appendix F) tracks **individuals with synthetic partners**: two independent single-gender cohorts, each drawing a synthetic spouse on marriage, no cross-population linkage. On dissolution each tracked individual must revert to *its own* single regime (`single_f` for women, `single_m` for men). The old "first leg" convention routed both cohorts to the same leg — silently wrong for one gender.

### What this PR does (synthetic mode)

Adds `own_stakeholder: str | None = None` to the simulation router (`route_gated_edges`), threaded through `simulate()`. When set, a dissolved row routes to the leg whose `source_stakeholder` matches — its *own* single regime with its own projected state — instead of unconditionally `legs[0]`:

```python
women = simulate(..., own_stakeholder="f")   # → single_f on dissolution
men   = simulate(..., own_stakeholder="m")   # → single_m on dissolution
```

- `own_stakeholder=None` (default) reproduces the exact prior "first declared leg" routing — **byte-identical** for every existing caller.
- New helper `_select_own_leg(legs, own_stakeholder)`. No cross-row gather, no matcher, no new per-subject array — those belong to linked mode (below).

### Synthetic vs the general two-sided (linked) split

[...]

PR #409, opened 2026-07-21
Collective / gated regimes + edge-fold of IID shocks

hmgaudecker:
## What this PR adds

This PR lets pylcm represent collective life-cycle decisions in which several
people choose one household action but retain distinct utilities, continuation
values, and outside options. It also adds two complementary ways to handle
uncertainty without unnecessarily enlarging stored value functions: folding a
purely transitory IID shock and declaring a correlated `JointTransition`.

The existing singleton-regime workflow remains the default.

### Collective choice

A regime becomes collective by declaring a `CollectiveUtility` in the slot it
already has for utility. For each feasible state-action cell, pylcm constructs
one action value per stakeholder, `Q_s(x, a)`. The household chooses one common
action

```text
a*(x) = argmax_a sum_s lambda_s Q_s(x, a),
```

and stores every stakeholder's value `Q_s(x, a*(x))` at that same choice. This
is the key distinction from solving separate optimization problems for the two
partners.

Everything is declared in a slot the regime already has, so there is exactly one
way to write a collective regime:

```python
Regime(
    functions={
        "utility": CollectiveUtility(
            utilities={"f": u_f, "m": u_m},
            objective=ParetoObjective(weights={"f": power}),  # omit for equal weights
        )
    },
    constraints={
        "participation_f": ValueDependentConstraint(
            predicate=participation_f,
            references={
                "V_alone_f": ProjectedRegimeValue(
                    regime="single_f", projection={"wealth": half_of_wealth}
                )
            },
        )
    },
    ...
)
```

The `utilities` keys are the stakeholders, in insertion order, and that order
fixes the trailing axis of `V` and of every published array. A stakeholder whose
body is `None` delegates to a `utility_<s>` entry arriving from the model level.

Value constraints may read each stakeholder's `Q_s`, which is what supports
limited-commitment applications where household decisions must satisfy
individual rationality rather than only maximize a household objective. When no
action satisfies all participation constraints, the cell carries a dissolution
flag; `solve(..., return_dissolution_flags=True)` publishes those flags and
`simulate(period_to_regime_to_dissolution_flags=...)` takes them back.

### Consent, separation, and fallback regimes

`ValueDependentTransition` declares a transition whose continuation depends on a
Boolean economic condition — mutual consent before entering a couple regime, or
dissolution when a participation constraint fails. It goes in the `transition`
slot, keyed by target:

```python
transition={
    "couple": ValueDependentTransition(
        probability=MarkovTransition(stays_married),
        gate=no_dissolution,
        routes={"f": StakeholderRoute(target_stakeholder="f", fallback=alone_f)},
        gate_references={
            "V_alone_f": ProjectedRegimeValue(
                regime="single_f", projection={"wealth": half_of_wealth}
            )
        },
        off_grid="pointwise",
    )
}
```

**The key is always the gate-open target.** A dissolution edge is keyed by the
*continuing* collective regime under `gate = ~D_target`; keying it by the
singleton would send both partners there whenever the couple stays together.

`routes` is keyed by source stakeholder, and each route owns four destinations:
the open regime (the dict key) and role (`target_stakeholder`), and the closed
regime and role (`fallback.regime`, `fallback.stakeholder`). If the gate is open
the leg reads the appropriate stakeholder value in the target regime; if closed
it reads the declared same-period fallback, usually that stakeholder's single
regime. The ordinary regime-transition draw still determines which edge is
attempted. Solve and simulation use the same routing declarations, including
target-specific state projections and stakeholder roles.

Simulation follows a fixed-size cohort of rows. A separation does not create two
new rows; `own_stakeholder` says which stakeholder each simulated row follows
after stakeholder-specific routing.

## One declaration surface

An earlier revision of this branch shipped **two** public spellings for the same
regime: the declarations above, and the lowered form the engine runs after
pylcm takes them apart — `stakeholders`, `pareto_objective`, `value_constraints`,
`same_period_refs`, `gated_edges` as `Regime` arguments, plus a public
`GatedEdge` class.

Publishing both would have made the equivalence and collision rules between two
spellings part of the long-term public API. The lowered vocabulary has been
removed from the public surface:

- the five slots are now `field(init=False)` derived values. They survive as
  *outputs* of lowering, which is the supported way to ask what a declaration
  produced, but they cannot be passed to the constructor or to `replace`;
- lowering no longer consumes `functions`, `constraints` and `transition`, so the
  declaration objects stay where the author wrote them, and the engine reads
  three derived views instead (`decomposed_functions`, `decomposed_constraints`,
  `decomposed_transition`);
- `GatedEdge` moved to `_lcm.gated_edge` and left `lcm.__all__`.

This costs no user. An AST walk over every `*.py` under `tests/` — counting
`ast.keyword` nodes named after one of the five slots, 0 parse failures at each
ref — finds **0** occurrences at the branch point and **7** at the head, none of
them to `Regime(...)`. The lowered vocabulary never existed before this branch,
so removing it before release breaks no promise that was ever made. The seven
survivors are engine vocabulary — five `_MockRegime` (which bypasses `__init__`),
one `build_pareto_weights`, one `dataclasses.replace` on a canonical engine
regime — and each is pinned by name in a census test that fails if an unlisted
one appears.

Closing the lowered form also required widening the declarations to cover three
configurations that had been reachable only in lowered form: a gated edge inside
a `Phased` transition, a `Phased` per-stakeholder utility, and per-stakeholder
utility bodies arriving from the model level.

### Folded IID shocks

An IID process can set `fold=True` when its realization is observed, used in the
current decision, and then discarded. pylcm still solves the choice problem at
every quadrature node and then averages the resulting values,
`E_epsilon[max_a Q(x, epsilon, a)]`, but does not store the shock as a value-
function axis.

This is deliberately restricted to cases where the reduction is exact: a fully
specified, non-persistent IID process under linear expectation, solved by grid
search, with no later transition or same-period value read depending on the
discarded realization. A folded shock whose conditioning state can move between
periods is refused, because the fold would otherwise price a distribution the
declared transition does not use. Unsupported combinations fail during model
construction.

### Correlated next-state innovations with `JointTransition`

Separate `MarkovTransition` declarations imply separate draws. That is wrong
when one innovation jointly determines several next-period states—for example,
a partner offer carrying correlated earnings and health characteristics.

`JointTransition` declares one finite-support lottery on a particular
source-target edge and several output laws that share the same support row.
Backward induction forms the expectation over that common draw inside the
source action value before maximizing, while simulation samples one row and
uses it for every output. The latent realization is transition-local: it is not
a model state, initial condition, simulation column, or stored value-function
axis. `Phased` can give solving and simulation different perceived and realized
joint laws while preserving the same structural support.

An output may land on a stochastic process's own nodes as well as an ordinary
grid, which is how correlated innovations reach a grid pylcm discretized rather
than one discretized by hand. Probability vectors are never silently normalized:
a lottery that does not return one probability vector per source cell is an
error, raised in the params-bound preflight.

## Transition architecture

The branch removes the former `TransitionLaws` metadata representation. Ordinary
deterministic and Markov state laws, stochastic-process laws, explicit process
entry, interpolation bases, and `JointTransition` kernels now lower to one
target-specific `TargetTransitionPlan`. Solve, simulation, validation,
diagnostics, and solver-scope checks consume that same plan.

Regime-transition probabilities remain a separate concern: they select the
target regime, whereas a target transition plan determines the target states
conditional on that target. Both public grammars are canonicalized by reachable
target, but combining target selection and conditional state evolution into one
public object would blur rather than remove this distinction. This supersedes
the internal-representation question raised in #423 without adding a wrapper
around `Regime.transition`.

## Correctness and supported scope

The implementation also adds the less visible machinery needed to make these
features compose safely:

- solve/simulation phase closure, target-period and age-specialized-grid
  handling, and target-specific parameter namespaces;
- explicit ownership of transition outputs and rejection of overlapping or
  incomplete producers;
- zero-mass and extended-real-safe arithmetic, avoiding `0 * -inf -> NaN` in
  regime mixtures, stakeholder aggregation, and gated fallbacks;
- Pareto weights that stay representable under normalization, and that are
  judged even when a weight reads a carried state, whose solve value comes from
  its imputation rather than from a grid axis;
- deterministic routing under ahead-of-time compilation and batched
  simulation; and
- fail-closed checks for solver or feature combinations whose economic timing
  is not implemented.

Collective regimes currently use `GridSearch` and linear expectation and do not
combine with folded shocks or EV1 taste shocks. Transition-local
`JointTransition` lotteries likewise use `GridSearch`; EGM-family solvers reject
them explicitly. A gated edge on a coarse or terminal transition is unsupported
and refused: a gate is a route, and a transition naming no target has no route
to fall back to.

The branch includes end-to-end tests of the documented public workflow and
class-level regression coverage for collective choice, participation and
dissolution, gated routing, folded-shock timing, joint-transition correlation,
parameter ownership, float32/float64 edge cases, and solve/simulation/AOT
parity.

## Review status

Nine bounded external mathematical/code-review rounds are complete. The definitive
Round-9 closure audit returned `CLEAN / CLOSED`: F1-F9 are `verified_fixed`, both
required profiles pass, no finding remains open, and `next_action` is `done`.

The final candidate certificate is an exact route-local direct-data-flow proof over all
six GridSearch corridors (ordinary and taste-shock singleton/collective solve and
simulation). It seals the 43-source dependency frontier and proves that the exact `Q_arr`
and `F_arr` produced by `Q_and_F` reach the full reducer and publication path without a
candidate-changing transformation. The synchronized closure suites reject all 66/66
semantic mutations and all 176/176 full re-anchored mutations, including the historical
MT9 and MT10 escapes.

The follow-up for review 5064008752 is complete at `90fe284f`. It preserves ordinary
DAG ancestors when carried-state Pareto weights are folded into solve-time imputation,
retains narrow reachability for `Phased` per-target transitions, and updates the public
`Model` docstrings to name `ValueDependentTransition` rather than the lowered internal
edge type. This repair head also incorporates #407 at `1c9f1d22` and re-anchors the source
inventory and direct-flow seals to the combined production tree. After #407 landed on
`main` at `b803c810`, the tree-preserving post-land merge produced current head
`723f9a26`; the inventory and unified verifier remain green. PR #409 was then
squash-merged into `main` at `e7e4a83f`.

Focused verification on the combined topology is green: float32 has 193 passed and 5
skipped; float64 has 198 passed; neither profile has failures or errors. The unified
static verifier passes, and the synchronized self-test rejects all 176/176 direct-flow
mutations.

Round-9 closure evidence at `1bff3e16`:

- candidate certificate: 93/93 passed at float32 and 93/93 at float64;
- full float32 profile: 6,787 passed, 316 skipped, 7 xfailed, 0 failed/errors;
- full float64 profile: 7,083 passed, 55 skipped, 7 xfailed, 0 failed/errors;
- `prek run --all-files`: all 34 hooks passed without modifying files.

Round 9 changes only the four certificate/test paths; no production `src/` path changes.
The closure state passes the strict audit gate with zero warnings or blockers.

A Windows portability follow-up at `3272af34` makes the sealed-source hashes
independent of checkout newline convention. The certificate now canonicalizes UTF-8
text before hashing, so CRLF checkouts match the LF inventory while any non-newline
source change remains detectable. The expanded candidate-certificate battery passes
95/95 at both float32 and float64; inventory generation, the unified contract/policy
verifier, all 176 synchronized mutations, and an actual `core.autocrlf=true` clone also
pass. This follow-up changes three test/tooling paths and no production `src/` path.

## Stack and related issues

This PR was stacked on `feat/continuous-outer` and merged after #407. The
collective, gated-edge, folded-shock, and transition-plan work touches the same
Bellman, transition, parameter, and simulation boundaries, so it is kept on the
single audited branch rather than split after the fact.

- Closes #401 (stakeholder-specific dissolution routing).
- Closes #402 (fold IID utility shocks out of stored values).
- Supersedes the representation question in #423; see **Transition
  architecture** above.


hmgaudecker:
## Head-vs-base failure diff — this branch causes 8 failures, all in `tests/test_ndimage.py`

`collective-regimes` @ `5c0b829`: **76** failures. `feat/continuous-outer` @ `db00f64`: **68**.
Caused by this PR: **8**. Inherited: **68**. Fixed by this PR: 0.

CI could not answer this. `main.yml` runs `pixi run -e tests-cpu tests` with no `-rf`, so
no failure is ever named — the ubuntu job printed 53 `F` marks and zero test names — and
that job additionally dies at ~96% with "the runner has received a shutdown signal". So
this was measured locally, chunked, one fresh process per chunk.

### Caused by this PR — 8, and they are not cosmetic

All eight are `tests/test_ndimage.py`, which this PR does not modify. It modifies
`src/_lcm/regime_building/ndimage.py`, swapping

```python
weighted_value = _multiply_all(weights) * contribution
# for
weighted_value = zero_safe_weighted_term(weight_product, contribution)
```

Verified directly rather than inferred from the logs — same file, same command, two trees:

| tree | result |
| --- | --- |
| `feat/continuous-outer` @ `db00f64` | **30 passed** |
| `collective-regimes` @ `5c0b829` | **8 failed, 22 passed** |

There are two distinct causes.

**(a) `map_coordinates` now returns a WRONG GRADIENT at every on-node coordinate — 1 test.**

`zero_safe_weighted_term(w, v)` is `w * where(w == 0, 0, v)`. Masking the value with a
hard select is sound when `w` is a *constant* — a Pareto weight, a regime-transition
probability, a quadrature weight. An interpolation **corner weight is not constant**: it
is a differentiable function of the very coordinate being differentiated. At an
exactly-integer coordinate one corner weight is exactly `0`, `where` zeroes that corner's
value, and the term becomes `w * 0` — whose derivative is `w'(c) * 0 = 0` instead of
`w'(c) * v`. The gradient loses precisely the corner whose weight is changing.

Minimal witness, `grid[i] = i²` so the correct slope on `[i, i+1]` is `grid[i+1] - grid[i]`:

```
  c=1.0   value=1.000   grad=-1.000   on_node=True     (correct: 3.0)
  c=1.5   value=2.500   grad= 3.000   on_node=False    correct
  c=2.0   value=4.000   grad=-4.000   on_node=True     (correct: 5.0)
  c=2.5   value=6.500   grad= 5.000   on_node=False    correct
  c=3.0   value=9.000   grad=-9.000   on_node=True     (correct: 7.0)

  d/dw [zero_safe(w, 7.0)] at w=0.00 -> 0.0   (should be 7.0)  <-- the hard select
  d/dw [zero_safe(w, 7.0)] at w=0.25 -> 7.0
  d/dw [zero_safe(w, 7.0)] at w=1.00 -> 7.0
```

Off-node gradients are exact; on-node ones come back as `-grid[c]`, i.e. only the
surviving corner's `-1 * v_lo` term. Values are unaffected — this is gradient-only, which
is exactly what makes it easy to miss. `test_gradients[map_coordinates]` catches it:
`-10.0` where `2.0` is expected.

**(b) The helper rejects integer weights — 7 tests.**

```
BeartypeCallHintParamViolation: parameter weight="JitTracer(int32[7])"
violates type hint FloatND | ScalarFloat | float
```

The previous expression was a bare multiply with nothing to violate. All seven are the
`int32` parametrization of `test_map_coordinates_against_scipy` /
`test_map_coordinates_round_half_against_scipy`; the float variants pass.

### Suggested direction

(b) is an annotation/dtype-coverage fix. (a) is the substantive one and I do not think it
should be fixed by loosening the test: the zero-safe contract as written is a statement
about *values*, and `map_coordinates` needs it to hold for *derivatives* too. Either
`map_coordinates` keeps the plain multiply (its corner weights are never `±inf`-valued,
so it may not need zero-safety at all), or `zero_safe_weighted_term` needs a
custom-JVP/`stop_gradient`-free formulation that preserves `∂/∂w` at `w == 0`. Worth
checking whether any other caller passes a weight that is differentiable in the argument
of interest.

### Inherited from the base — 68

Unchanged from `db00f64`; none of these are this PR's doing.

| file | n |
| --- | ---: |
| `tests/test_solution_on_toy_model_stochastic.py` | 24 |
| `tests/test_solution_on_toy_model_deterministic.py` | 8 |
| `tests/simulation/test_subject_batching.py` | 7 |
| `tests/solution/test_n_nbegm.py` | 6 |
| `tests/solution/test_n_nbegm_finite_baseline.py` | 6 |
| `tests/test_processes.py` | 5 |
| `tests/solution/test_negm_bequest.py` | 3 |
| `tests/solution/test_negm_serviceflow.py` | 3 |
| `tests/solution/test_egm_carry.py` | 1 |
| `tests/solution/test_egm_discrete.py` | 1 |
| `tests/solution/test_n_nbegm_continuous.py` | 1 |
| `tests/solution/test_n_nbegm_epstein_zin.py` | 1 |
| `tests/solution/test_nbegm_distributed_carry_sharding.py` | 1 |
| `tests/test_regression_test.py` | 1 |

The seven `test_subject_batching.py` failures are worth calling out: they sit in
`tests/simulation`, which this PR modifies heavily, so they were my prior candidate for a
genuine regression. They fail identically on the base.

### Method, and what is NOT covered

One process per chunk (peak RSS resets between them); `-q -n 2 --tb=no -rf`, with
`### EXIT <chunk> = $?` recorded after each so a chunk killed by the memory ceiling shows
up as *unmeasured* rather than silently counted as clean.

> **The four `tests/test_mahler_yum_*.py` files are UNMEASURED on both sides** — that
> chunk exited `143` (SIGTERM from the memory cap) on head *and* base, even at `-n 1`.
> Symmetric, so it is unlikely to hide an asymmetry, but I have not measured it and am
> not claiming it is clean.

The base was measured in a **fresh** detached worktree at `db00f64`, not in the shared
one: that worktree had uncommitted work in `src/_lcm/egm/upper_envelope/query.py`, the
same area as most of the inherited failures, so measuring there would have compared this
PR against someone's work in progress. `_lcm.__file__` was checked to resolve inside each
tree, since a linked pixi env can otherwise resolve the package to a different branch.

hmgaudecker:
## Fixed — `27bc15f`, and the failure set is now identical to the base

All 8 regressions from the diff above are closed. `tests/test_ndimage.py` is back to
**30 passed**, matching `feat/continuous-outer` @ `db00f64`.

**The gradient one.** `zero_safe_weighted_term` now masks only where the value is
**non-finite**, instead of on every zero-weight node. That keeps the property the helper
exists for while restoring `∂/∂w` at `w == 0`:

- zero-mass safety unchanged — `zs(0, ±inf)` and `zs(0, nan)` are still exactly `0.0`,
  and `zs(1, -inf)` is still `-inf`;
- `d/dw [zs(w, 7.0)] == 7.0` at `w = 0`, `0.25` and `1.0` alike (it was `0.0` at `w = 0`);
- **values are bit-identical** to the previous form — for finite `v`, `0 * v == 0`
  whether masked or not. Checked over 10,000 random cells including zero-weight slots
  with `-inf` on each of them, and then at suite scale: `tests/solution` returns the
  **same 23 failures, node for node**, as before the change. So this is gradient-only
  and carries no numerical risk to the EGM family.

The select still sits on an *operand* of the multiply, so the FMA-contractibility
argument in the module docstring is unaffected (docstring updated to say why the mask is
restricted).

**The 7 annotation ones.** Parameters and return were `FloatND`-only; `map_coordinates`
accepts `FloatND | IntND`, so an integer weight product raised
`BeartypeCallHintParamViolation` and then `...ReturnViolation`. Widened.

New `tests/regime_building/test_zero_safe_gradients.py` pins all three legs — safety,
derivative, value-invariance — plus the `map_coordinates`-on-a-quadratic case that
started this.

### Also in this push — `9f75cc6`

Comparing every `X` / `X_collective` builder pair (there are exactly two in `src/_lcm`)
turned up a second instance of the class the branch-2 audit's F1 repaired:
`get_Q_and_F_terminal_collective` had no `next_state_names` parameter while its singleton
twin does and both call sites thread it. That argument feeds
`_fail_if_unproduced_next_state_is_read`, whose test is `read_names & next_state_names`,
so against the default `frozenset()` the guard could never fire on the collective path —
the model failed later with a bare missing-argument error instead of the named one.

Severity is **diagnostic, not numerical**, and I checked that rather than assuming it:
`next_state_names` reaches `_get_U_and_F` and is used *only* for the guard call, taking no
part in graph construction. Worth noting the corollary for F1 itself — of the four
arguments that repair restored, `flow_transitions` and `continuation_functions` are the
load-bearing ones and `next_state_names` was the diagnostic member; this is that member,
still dropped one function away.

### Current state of this branch vs its base

| | head `9f75cc6` | base `db00f64` |
| --- | ---: | ---: |
| failures | 68 | 68 |
| caused by this PR | **0** | — |

Same caveat as before: the four `tests/test_mahler_yum_*.py` files remain **unmeasured on
both sides** (that chunk hits the memory ceiling even at `-n 1`), so this is "no
difference in what was measured", not "clean everywhere".

hmgaudecker:
@/tmp/claude-1000/-home-hmg-econ-dev-pylcm-lcm-replications-src-lcm-reps-ecksteinCareerFamilyDecisions2019/da153776-cb2a-4e8a-9fc3-5b37c71b1f06/scratchpad/pr409-r3.md

hmgaudecker:
Cascaded `feat/continuous-outer` in (`f318eb0`), local verification on the merged tree:

- `tests/solution` — **1250 passed, 3 skipped, 2 xfailed, 0 failed** (run as four file chunks; see below)
- rest of suite — **1923 passed, 0 failed**, 7 errors

The 38 DataFrame-shape failures this branch was carrying are **gone**. They came from `d0d655c` adding `nested_policy_fallback` without regenerating the pickled fixtures; `610b6f9` on #407 is the fix, and cascading it in resolved them. Nothing in the merge itself needed conflict resolution beyond the eight in `regime_building/processing.py`, resolved by provenance, plus a collective-twin signature widening in `Q_and_F.py` that `ty` caught and the suite did not.

### The 7 remaining errors are #407's, not this branch's

`tests/test_mahler_yum_continuous_outer.py` and `tests/test_mahler_yum_implicit_pilot.py` error at **setup** with `Out of memory allocating 158915772464 bytes` — **148 GiB**.

- Both files exist only on `feat/continuous-outer` and `collective-regimes`, **not on `main`**.
- Reproduced **byte-identically** (same allocation to the byte) on `origin/feat/continuous-outer`'s own tip in a detached worktree, with `_lcm.__file__` asserted to resolve into that worktree. So the merge did not introduce it.
- It is not contention or an `-n` artifact: a single test at `-n 0` with no neighbours asks for the same 148 GiB, which exceeds the whole machine.

**This would hit CI too.** Both are `@pytest.mark.slow`, and CI phase 3 is `pytest tests --ignore=tests/solution -m slow -n 2` — the same command shape, against a 16 GB runner. It is currently masked: Ubuntu dies at ~99% of phase 1 (`-m "not slow"`) with `The runner has received a shutdown signal` — all dots, zero test failures — so the `&&` chain never reaches phase 3, and macOS passes only because it deselects `slow` wholesale. Fixing the phase-1 memory problem would expose this one.

This looks like the class an in-flight external audit already has open (workspace proportional to `n_rows * n_cells * max_runs` causing supported-CI OOMs, `escalation_required`), so it is flagged here rather than re-diagnosed.

### Note for anyone verifying locally

Do not try to phase `tests/solution` with `-m "not slow"` — it collects **nothing**, because the entire directory is `slow`. Chunk by file instead (`ls tests/solution/*.py | split -n l/4`, one pytest invocation per chunk); a single full-directory run gets OOM-killed at ~96% with no summary line.

timmens (commented):
Very nice feature! A few comments though:

## Blocking: silent economic or routing errors

### Make off-grid value-dependent routing consistent between solve and simulation

Solve evaluates the gate on target-grid nodes and then interpolates the resulting open/fallback continuation. Simulation interpolates the gate's value operands at the realized target state and then evaluates the gate. These operations do not commute and can change the source-period optimal action.

I reproduced this through the public `Model.solve()` and `Model.simulate()` API. A source chooses between a safe action worth `1` and an action leading to an off-grid target state `x=0.6`. On the target grid, the gated continuation interpolates to `1.8`, so solve assigns the risky action a discounted value of `1.71` and chooses it. At the realized state, simulation obtains target value `2.2` and reference value `2.3`, closes the gate, and routes to a fallback worth `0`. The resulting path is `src -> fallback`: the solver rejects a guaranteed value of `1` for a path that realizes `0`.

Could solve and simulation evaluate the gate and branches in the same order? Restricting value-dependent transitions to on-grid target states would also give a coherent narrower contract. The current documentation that off-grid gating is approximate does not communicate that the approximation can reverse the preceding optimal action.

### Carry stakeholder identity with each simulated row

The call-wide `simulate(own_stakeholder=...)` argument creates two related problems.

First, it must be valid for every active collective source, including sources with no simulated rows. In a public-API MRE containing a couple with roles `("f", "m")` and a team with roles `("owner", "worker")`, every possible value fails even though the simulated population contains only a couple member.

Second, a singleton-to-collective `EdgeLeg` already declares `target_stakeholder`, but simulation discards that identity. I reproduced a person entering marriage through `target_stakeholder="f"` and later encountering a closed dissolution gate. Calling simulation with `own_stakeholder="f"` produces `single_f_0 -> married_1 -> single_f`, while changing only the call-wide argument to `"m"` produces `single_f_0 -> married_1 -> single_m`.

Could simulation store stakeholder identity per row, initialize it from `target_stakeholder` on collective entry, and use it for later routing? A per-source mapping would solve the disjoint-vocabulary failure, while retaining the row-level identity would also prevent a later argument from changing the person's economic identity.

### Draw target-local folded shocks during simulation

A target-local IID process with `fold=True` is integrated correctly during solution but is never realized when simulation enters that target. I reproduced this through the public API with a source choosing between an outside payoff of `0.5` and a job worth `1 + bonus_shock`, where the folded shock has mean zero. Solve correctly stores the expected job value as `1` and chooses the job. In the simulated job period, `bonus_shock` remains `NaN`, the reported action is `0`, and the value is `NaN`.

Could simulation retain the ordinary process-entry draw for a folded target while solve continues to integrate its axis out of the stored value? Please also extend the existing folded-target simulation test to assert the target period's realized shock, action, and value; it currently checks only the source-period action and value.

### Use state-conditioned probabilities when folding a shock

`fold=True` is accepted together with a `StateConditioned` IID process, but the fold uses one constant probability vector derived from the widest conditional distribution. Every value of the conditioning state therefore receives the same folded value. In a public-API MRE with `sigma=0.05` under low uncertainty and `sigma=1.0` under high uncertainty, the correct optimal values are `0.1` and approximately `0.304`. PyLCM stores approximately `0.304` for both. This can reverse upstream decisions; against an outside option of `0.2`, only the high-uncertainty prospect should be selected.

Could the fold apply the process's category-specific probability row at each value of the conditioning state? If this requires broader changes to the fold reducer, the public combination of `fold=True` and `StateConditioned` should be rejected until it is supported.

### Validate simulate-phase joint probabilities over carried states

The joint-transition preflight evaluates both phase variants on `regime.solution.state_action_space`. A simulate-phase probability function that reads a carried-only state, declared as `Phased(solve=imputation, simulate=Grid)`, therefore cannot be evaluated because that state has no solution grid axis. The preflight logs that the simulate phase was not numerically validated and continues even at `log_level="debug"`.

In a public-API MRE, the simulate-phase function returns `[1.2, -0.2]` as a function of carried `experience`. Both `solve(log_level="debug")` and `simulate(log_level="debug")` succeed, and simulation samples the first joint-support node using these invalid probabilities. Could the simulate-phase preflight evaluate over the simulation grids, including carried-only states? At minimum, debug mode should refuse to execute a joint lottery whose probability function could not be validated.

### Enforce Boolean gate outputs

The public contract requires a gate to return a Boolean condition, but an ordinary callable returning a numeric value is accepted. JAX treats every nonzero value as true in `where`, so a gate returning `0.25` silently opens the edge for every row rather than representing a 25% probability or raising an error. This changes both the solved continuation and the simulated route without any diagnostic. Could the gate boundary require a Boolean output dtype and reject numeric outputs with an error naming the edge and returned dtype?

## Blocking: model-validation contradictions

### Discover transition parameters from the specific edge that consumes them

A gated-edge parameter is currently classified as engine-provided whenever its name matches a state in any unrelated regime, a variable of the source regime, or a reference key belonging to another edge. It is then omitted from `get_params_template()`, although the edge evaluator does not receive that value from the engine. For example, a gate `gate(V_target, marriage_bonus)` becomes impossible to call when an unrelated regime declares a state named `marriage_bonus`: omitting the parameter produces a missing-argument error, while supplying it is rejected as unknown.

Could parameter discovery construct the non-parameter set separately for each edge from that edge's actual target states and injected value/reference operands?

### Reject free parameters in `SamePeriodRef` projections at model construction

The public contract says that a regime-level `SamePeriodRef.projection` cannot introduce free parameters, but this is not validated. For example, a projection `project_wealth(wealth, divorce_cost)` is accepted, while `divorce_cost` is omitted from `get_params_template()`. Solving without it then reports a missing argument, and supplying it is rejected as an unknown parameter, so the accepted model has no valid parameter assignment.

Could model construction inspect each projection signature and reject every argument that is not supplied by the declaring regime, naming the reference, projection, and unsupported argument in the error?

### Remove the discrete-action requirement for collective regimes

Why must every collective regime contain a `DiscreteGrid` action? The collective readout already handles an empty action-axis tuple, and the ordinary maximization path covers continuous actions. After removing only this validation check, continuous-only and zero-action collective models solve, simulate, and produce the expected stakeholder values. The current restriction therefore forces terminal regimes with no decision to declare a dummy discrete action and rejects continuous-only collective problems that the engine can already execute.

Could we remove the check and add public tests for both cases?

## Non-blocking: public API and scope discussions

### Colocate collective and value-dependent behavior with the primitives they modify

Could we reconsider whether these features need separate top-level `Regime` fields? The current API distributes collective choice across `stakeholders`, `weights`, and specially named utility functions; value-dependent feasibility across `value_constraints` and `same_period_refs`; and value-dependent routing across `transition` and `gated_edges`. This creates multiple sources of truth and requires names and target keys to be synchronized across separate mappings.

Would it be cleaner to represent each extension inside the primitive it modifies? For example:

```python
Regime(
    functions={
        "utility": CollectiveUtility(
            utilities={"f": utility_f, "m": utility_m},
            objective=household_objective,
        ),
    },
    constraints={
        "participation_f": ValueDependentConstraint(
            predicate=participation_f,
            references={"outside_option": ...},
        ),
    },
    transition={
        "next_couple": ValueDependentTransition(
            probability=MarkovTransition(next_couple_probability),
            gate=no_dissolution,
            fallbacks={"f": ..., "m": ...},
        ),
    },
)
```

In particular, I would favor consolidating `transition` and `gated_edges` into a `ValueDependentTransition`. `GatedEdge` exposes internal graph terminology and leaves one semantic transition split across two declarations. `SamePeriodRef` also seems less than ideal: the name does not tell users that it evaluates another regime's current-period value at mapped state coordinates, and inside an edge fallback the same projection also writes the simulated destination state. We considered `ProjectedRegimeValue`, but it is unclear whether that would be more understandable to users. The exact helper names can remain open for discussion.

The important properties are a small top-level `Regime` vocabulary, one local source of truth for each declaration, and ordinary function-parameter discovery for the household objective.

### Make Pareto weights part of the model's parameter interface

`Regime.weights` currently accepts only fixed Python floats, which are closed into the compiled model and do not appear in `get_params_template()`. Consequently, estimating even a constant bargaining weight requires rebuilding and recompiling the model, while state-dependent or promised-weight specifications are inexpressible without encoding weights as separate regimes. Since the collective numerical core already accepts broadcastable weight arrays, could we expose at least constant Pareto weights through `params` before stabilizing this public API? It would also be useful to decide whether DAG- or state-dependent weights belong in the intended collective-model scope.

### Support the ordinary repeated-IID use case before stabilizing `fold=True`

The current persistence guard rejects an IID shock whenever the same process state is present in a reachable next-period regime, including a self-transition. This excludes the ordinary use case in which a shock is independently redrawn every period. The test documentation explicitly identifies this as the central repeated-IID use case while limiting the implementation to shocks that appear for one period and then disappear.

This seems more conservative than the model semantics require. PyLCM already knows that the state is an `_IIDProcess`, and `fold=True` says that the target's stored value has already integrated out that axis. Under the existing restrictions on linear expectation, fixed distribution parameters, GridSearch, and downstream reads of the realized shock, an incoming continuation should be able to read the already-folded target value without constructing a `next_<shock>` interpolation coordinate; simulation would continue to draw a fresh shock each period. There is already a narrow version of this treatment for folded-only targets in `_strip_folded_axes_for_scalar_targets`.

Could we either support this repeated-IID case before making `fold=True` public, or keep the option experimental until it is supported? The current one-period-only scope requires dedicated transient regimes and excludes the main source of the intended repeated value-function memory saving.

### Allow beliefs and realized fallback laws to differ

The projection in an `EdgeLeg.fallback` currently performs two phase-specific jobs: during solution it locates the fallback value the agent expects, while during simulation it writes the realized state in the fallback regime. Regular and joint transitions support `Phased(solve=..., simulate=...)`, but gated fallbacks require one shared projection. For example, an agent may expect to retain 50% of household wealth after divorce while the simulated settlement gives 40% after costs.

Could the value-dependent transition API support separate solve- and simulate-phase fallback laws, either through `Phased` or by separating the fallback-value mapping from the realized state transition?

### Support participation constraints in terminal collective regimes

Terminal collective regimes currently cannot declare `value_constraints` or `same_period_refs`. This rules out an economically meaningful final-period participation or dissolution decision, even though the decision only needs current-period stakeholder utilities and same-period outside-option values. Users must introduce an artificial later regime or period to represent it. Is this restriction required by the implementation, or could terminal collective regimes evaluate the same value-aware feasibility mask using their terminal stakeholder action values?

## Non-blocking follow-ups

### Include dissolution flags in the planned solution object

Issue [#428](https://github.com/OpenSourceEconomics/pylcm/issues/428) already tracks returning and consuming one complete `SolutionResult`. This PR adds dissolution flags as another artifact produced by solve and consumed by simulation, currently through an additional Boolean-controlled tuple shape. Could dissolution flags be added explicitly to the scope of #428 so that the eventual solution object keeps values, replay policies, and dissolution information from the same solve together?

The compatibility metadata should validate exact period/regime coverage and age-resolved array shapes for values and dissolution flags; the current rank-only value check accepts a one-element array for a four-node state grid and silently changes the simulated Bellman value.

### Include gate evaluators in the ahead-of-time compilation contract

`Model(n_subjects=N)` is documented as compiling all simulation functions for the declared batch shape, but `compile_all_simulation_phases` collects only action maximizers, next-state functions, and regime-transition probability functions. The new per-subject `simulate_gate_evaluator` remains a lazy `jax.jit` callable and compiles when a gated edge is first encountered during simulation. The ahead-of-time gated-edge test checks only that action-maximization functions were replaced by compiled programs, so it does not detect this gap.

Could gate evaluators also be lowered and installed during the ahead-of-time step, with a test that inspects them? If they are intentionally outside the ahead-of-time scope, the `n_subjects` documentation should identify which simulation functions remain lazily compiled.

timmens (commented):
Thanks for the substantial follow-up. I reviewed the current head
`494a3a5c7` from the last head we walked through, `e41762d0c`, and checked the
new GitHub discussion. The new commits resolve the disconnected-role requirement
and large-float Pareto normalization. The projected-landing changes also close
the earlier solve/simulation routing inconsistency. I have removed those comments.

The focused current-head batteries pass: 143 passed, 0 failed, 0 errors, and
0 skipped. I also reran the public-API counterexamples below at the current head.

## Blocking correctness issues

### Resolve carried states through their solve imputation in Pareto weights

A Pareto weight may read a regime state. A carried state is declared as
`Phased(solve=imputation, simulate=Grid)`, so its solve value comes from the
imputation and has no solve-grid axis. A collective model whose weights read such
a state builds successfully, then `solve(log_level="debug")` raises
`KeyError: 'power'` in `check_pareto_weights`, which looks only in the solve state
grid.

Could the weight evaluator and its precondition resolve carried states through
their solve-side imputation? Please cover a case where the imputed bargaining
power selects one stakeholder's preferred action while the seeded simulation
value differs, then assert the solved and simulated choices.

### Preserve time-t conditioning when a folded shock spans periods

`StateConditioned` defines the next shock distribution from its conditioning
state at time `t`. A folded target currently stores a value averaged with the row
selected by its own `s[t + 1]` cell. An incoming continuation therefore uses
`s[t + 1]` to select the distribution.

At the current head, I started everyone in the low-risk category and let the
period-0 action set period-1 risk. The unfolded model chooses low risk and has
`V0 = 0.2`. With `fold=True`, the same model chooses high risk and has
`V0 = 0.412952`. Simulation then draws the period-1 shock from the low-risk row,
as required by the time-0 conditioning state, and realizes value `0.2`. The
folded solve has priced a distribution that the declared transition does not use.

Could the fold retain the time-t conditioning row needed by incoming
continuations, or reject `fold=True` when its conditioning state can move? A
folded-versus-unfolded policy test with a switching conditioning state would pin
the timing contract.

### Map joint outputs onto stochastic-process nodes

`JointTransition.outputs` accepts target-state laws without restricting the
target grid type. A process-backed output is a natural way to represent correlated
innovations. With a target-only `NormalIIDProcess`, a two-node joint draw
`[-0.1, 0.1]` at equal probabilities builds successfully, then `solve()` raises
`TypeError: Indexer must have integer or boolean type, got float32`.

The joint output currently gets `PhysicalCoordinate()`, while process value
functions are stored on process nodes. Could joint output values be mapped onto
the target process grid through the existing interpolation-basis machinery? If
this composition is outside the first release, please reject process-backed joint
outputs during public validation and document the restriction.

## Non-blocking follow-ups

### Align `ValueDependentTransition.probability` with the transition grammar

The field accepts `UserFunction | MarkovTransition` and says it accepts what a
plain target transition entry accepts. Lowering places it in a per-target cell,
where the grammar requires `MarkovTransition`. A plain callable therefore passes
the declaration type and immediately fails `Regime` validation.

Could the field require `MarkovTransition`, or wrap a plain probability function
automatically, so the annotation and documentation match the resulting grammar?

### Validate zero-argument callable Pareto weights

`check_pareto_weights` skips evaluators with no argument names. At the current
head, callable weights returning `(-1, 2)` and `(0, 0)` both pass a debug solve.
The second declaration leaves the objective identically zero and selects the
first action by tie-breaking.

Could every callable declaration be evaluated, including zero-argument ones?
The same check could require one scalar per cell so an accidental vector return
cannot broadcast across an action axis.

### Reject undeclared event axes in joint probabilities during preflight

The preflight checks the final probability-axis length and row sums. A
zero-argument probability function returning a valid `2 x 2` matrix for a
two-node support passes those checks. Solve later raises the raw JAX error
`Axis must be specified when shapes of a and weights differ`, because the extra
leading axis has no declared source variable.

Could preflight compare the complete result shape with the source cell shape plus
the support axis? A zero-argument function should return exactly one probability
vector.

## API discussion

### Publish one declaration surface

`CollectiveUtility`, `ValueDependentConstraint`, and
`ValueDependentTransition` give each extension a clear home. `Regime` also
publicly accepts the lowered `stakeholders`, `pareto_objective`,
`value_constraints`, `same_period_refs`, and `gated_edges` fields, while
`GatedEdge` remains exported from `lcm`. This makes the equivalence and collision
rules between two declaration forms part of the long-term public API.

Is the lowered vocabulary intended as a public compatibility surface? Since both
forms are introduced here, this is a good point to keep the lowered representation
engine-facing and let tests construct it through an internal helper.

I am leaving the solution-artifact concern to issue #428, where dissolution flags,
exact period/regime coverage, and age-resolved shapes are now explicitly tracked.

hmgaudecker:
Cascaded `feat/continuous-outer` in again (`9b511cc2`), and the branch is now
`MERGEABLE` against it after being `CONFLICTING`. The only conflict was the
`_replay_continuous_outer_decision` docstring in `simulation/simulate.py`: #407's
`d191f850` replaced the single replay candidate with a scored candidate bank, so its
description is the accurate one, kept alongside this branch's sentence that inference
must refuse whenever any fallback entry is `True`.

### Round 6 came back: eight of nine closed

F1-F8 are `verified_fixed`. F9 stays open, and it is worth being precise about what it
is: a **certification** finding about the certificate that establishes zero
represented-candidate search error, not a demonstrated wrong answer from the solver. No
production omission has been exhibited in any round, and both certified sources are
byte-identical between rounds 6 and 7 (`git diff --name-only 4178176a b2d93944 -- src/`
is empty).

Round 6 named two residual escapes. Both were reproduced here from the shipped
artifacts before anything was changed, rather than taken on the reviewer's word.

1. **The compiled policy was blind to half the inventory.** The
   `profile-contract.policy.json` shipped inside `bundle-round6.zip` contains **0**
   occurrences of `candidate_sources` against 4 of `candidate_source`. The cause is
   structural, not a stale edit: the bundler compiles the contract down to one `source`
   and one `source_digest` per profile, so a second certified source declared under
   `additional_sources` cannot reach the policy the protocol enforces.

2. **One fixed candidate ordering hid an omission class.** The mask sweep was complete
   over all 63 nonempty feasibility neighbourhoods but ran at a single ascending
   ordering `Q = [1,2,3,4,5,6]`. Under a strictly ascending ordering the winner of any
   mask is its highest feasible index, so candidate 0 wins only the singleton mask and
   only candidate 5 is ever the full-support winner. An omission of any other candidate
   conditional on full support therefore changed nothing the certificate observed.

### What changed

The contract's primary `source`/`source_digest` pair now names the generated inventory
and the SHA-256 of its bytes, with every certified source under `additional_sources`.
Because each source's digest is a record inside that inventory, changing either
source's bytes moves the inventory file, moves the anchor, and makes a stale policy
fail the join. That is the one shape a compiled policy can carry that still binds a set.

The certificate now sweeps the ordering as well as the mask: `rank_vectors` generates
one strict total order per candidate, each won by that candidate, and all four routes —
singleton and collective, solve and simulate — enumerate every ordering against every
nonempty mask against a scalar argmax that refuses ties and empty support. The AST gate
requires both matrices and both loops, so deleting the sweep is refused rather than
silently green.

Two controls ship with it, each requiring all five of its perturbations to be rejected:
`mt7` seeds a byte change in each certified source, the anchor replaced by each, and an
absent anchor; `mt8` collapses the ordering matrix and removes the sweep from each route
independently.

### Evidence at `b2d93944`

Full suite on one 16-core node, pinned to the commit, fresh JAX compile cache per leg
(float64 `0 -> 34877` entries, float32 `0 -> 33209`), SLURM `COMPLETED` in `00:33:22`
with exit code `0:0`:

| precision | tests | passed | failures | errors | skipped | time |
| --- | --- | --- | --- | --- | --- | --- |
| 64 | 7090 | 7028 | 0 | 0 | 62 | 1193.6 s |
| 32 | 7055 | 6732 | 0 | 0 | 323 | 779.8 s |

Test and skip counts are identical to round 6 at both precisions, which is expected:
this round changed the content of four tests, not their number. The candidate
certificate is 50 passed / 0 failed at both precisions, and `prek --all-files` is clean
(34 hooks, 0 failed).

### One thing that moved the wrong way

The protocol's own anchor checker now reports `all_structural_markers_present: false`
where round 6 reported `true`. This is a direct consequence of the anchor change: that
tool greps the file named by `candidate_source` for full-grid enumeration markers, and
that file is now a JSON inventory. Nothing about the production sources changed. The
enumeration route is instead resolved through the syntax trees of *both* certified
sources by the certificate's structural half, which is strictly more than a marker grep
over one file — but the tool prints `false`, and that is recorded rather than
paraphrased away.

The earlier OOM report in this thread is from 2026-07-29 and describes a tree many
commits behind; treat this comment as superseding it.

hmgaudecker:
@timmens Final close-out: the current head is `723f9a26` and the external correctness audit is now definitively `CLEAN / CLOSED` after nine rounds. All nine findings are `verified_fixed`; there are no open findings, blockers, or pending decisions.

The last residual item was certificate architecture, not a demonstrated production error. The final form proves exact `Q_arr`/`F_arr` transport through all six ordinary/taste-shock singleton/collective solve/simulation corridors, seals the 43-source helper/reducer frontier, and rejects 66/66 semantic plus 176/176 synchronized full mutations (including the historical MT9/MT10 escapes).

Round-9 verification at `1bff3e16`: candidate certificate 93/93 at each precision; full float32 6,787 passed / 316 skipped / 7 xfailed; full float64 7,083 passed / 55 skipped / 7 xfailed; zero failures/errors; all 34 pre-commit hooks passed.

The follow-up head `3272af34` fixes a Windows-only certificate portability defect: sealed UTF-8 source text is now hashed after newline canonicalization, so a CRLF checkout no longer produces false mismatches while non-newline mutations remain detected. Its expanded certificate battery passes 95/95 at both precisions; inventory generation, the unified contract/policy verifier, all 176 synchronized mutations, and a real `core.autocrlf=true` clone pass as well. The follow-up changes only three test/tooling paths and no production `src/` path.

This supersedes my Round-7 status comment above.

The follow-up to review 5064008752 was implemented at `90fe284f` and is included in this head. It preserves ordinary DAG
ancestors in carried-state Pareto imputation, retains narrow reachability for `Phased`
per-target transitions, and replaces the two public `Model` docstring references to the
lowered edge type with `ValueDependentTransition`. It also incorporates #407 at
`1c9f1d22` and re-anchors the certificate to the combined tree. After #407 landed on
`main` at `b803c810`, the tree-preserving post-land cascade produced `723f9a26`; the
inventory and unified verifier remain green. PR #409 was then squash-merged into
`main` at `e7e4a83f`. Focused combined-tree
verification is green: float32 193 passed / 5 skipped; float64 198 passed; zero failures
or errors in either profile. The unified verifier passes and all 176/176 synchronized
direct-flow mutations are rejected.

timmens (approved):
Thanks for the substantial follow-up. I reviewed the current head `1bff3e16b` from
`494a3a5c7`, reran the previous-finding battery, and checked the declaration redesign
and candidate certificate. The earlier joint-transition, callable-probability,
zero-argument-weight, and duplicate-declaration-surface concerns are resolved.

I am approving conditionally on the required CI becoming green. Windows currently
fails because the candidate certificate hashes raw checkout bytes, so CRLF line endings
change the source seals; the benchmark job is still running. The remaining review
points below are non-blocking, and I am happy for you to decide whether they belong in
this PR or a follow-up.

## Non-blocking correctness follow-ups

### Preserve ordinary DAG dependencies of carried-state Pareto weights

A Pareto weight may now read a carried state, which fixes the direct case. The carried
state's solve imputation is also a first-class regime function, so it may read another
function in the regime DAG. In a small public model I used `weight_f(power)`,
`impute_power(power_signal)`, and a normal regime function named `power_signal`. Model
construction succeeds, then `solve(log_level="debug")` raises
`KeyError: 'power__power_signal'`.

`_compose_carried_imputations` includes the carried imputations alone and consequently
reclassifies the ordinary function read as a parameter. Could the weight evaluator
compose the reachable solve-function DAG while preserving the existing
function/parameter distinction? A regression test can assert the resulting stakeholder
values `(3, 0)`. This is a specialized, immediately visible failure with a simple user
workaround, so I do not consider it a merge blocker.

### Preserve narrow reachability for `Phased` target mappings

The folded-conditioner guard appears to lose the declared target set when the regime
transition is `Phased`. I reproduced this with two disconnected paths:
`safe_entry -> folding` holds `risk_type` fixed, while
`moving_entry -> sideways` changes it. `moving_entry.transition` is
`Phased(solve={"sideways": ...}, simulate={"sideways": ...})`. Model construction
rejects the valid folding path and says `moving_entry` reaches `folding`.

`_reachable_regime_targets` recognizes a direct mapping and treats the outer `Phased`
object as a coarse transition over every regime. Could it resolve either phase first?
The phase grammar already guarantees matching forms and target keys. This requires a
rare combination of features and fails during model construction, so I also consider
it non-blocking.

## Small documentation cleanup

The single declaration surface looks much clearer to me. One small remnant remains:
the public `Model.solve` and `Model.simulate` docstrings describe dissolution in terms
of `GatedEdge`, while that class is now internal and users declare a
`ValueDependentTransition`. Could those two references use the public name and
vocabulary?

Focused local checks at this head: 64 previous-finding tests passed, 405 declaration
and integration tests passed, and all 93 candidate-certificate tests passed.

PR #416, opened 2026-08-03
Fix silently wrong Epstein-Zin results at high risk aversion; give recursive preferences their own model slots

### LINKED ISSUE #412: GridSearch bypasses PowerMean.aggregate and collapses finite lotteries to zero
## Summary

`PowerMean` implements a numerically stable `aggregate()` whose docstring documents
exactly one failure mode — and **nothing in `src/` ever calls it**. `Q_and_F` instead
executes the naive `inverse(Σ wᵢ · transform(vᵢ))` route that `aggregate()`'s own
docstring warns against, so a finite certainty equivalent collapses to `0` and the
Bellman argmax reverses.

Found by the round-4 external audit of #390 as finding **F6** (root cause R4). It is
filed separately because it is not a reachability/DC-EGM defect — it sits in the
certainty-equivalent layer and affects any `GridSearch` regime using `PowerMean`,
independently of that branch.

## The defect

`PowerMean.aggregate()` states the problem in its own words:

> The naive `inverse(sum(w · transform(v)))` overflows when `risk_aversion > 1` and `v` is
> near the borrowing constraint: the intermediate `v^(1-ra)` exceeds the dtype's range and
> the certainty equivalent collapses to zero or infinity.

`Q_and_F` builds precisely that expression — `ce.transform(...)` per value, a
probability-weighted sum, then a single `ce.inverse(...)` — at both aggregation sites (the
Bellman `Q` and the diagnostic intermediates). `aggregate` is never dispatched to:

```console
$ grep -rn "\.aggregate(" src/
$            # no matches
```

## Reproduction (verified locally, not merely relayed)

At `risk_aversion = 8` on an even two-point lottery, with a 60-digit `Decimal` reference:

| precision | lottery | exact CE | current path |
|---|---|---|---|
| float32 | (1e-8, 2e-8) | `1.102862741e-08` | `0.0` |
| float64 | (1e-50, 2e-50) | `1.102862741e-50` | `0.0` |

The mechanism, step by step in float32:

```
transform    v^(1-8) = v^-7  ->  [inf inf]      # (1e-8)^-7 = 1e56 > float32 max 3.4e38
weighted sum                 ->  inf
inverse      inf^(-1/7)      ->  0.0
```

This is not a tolerance question. The transform overflows the dtype before any summation
happens, so the information is destroyed at the first step and no downstream care recovers
it. It affects **scalar-only and stateful-only** targets alike.

## Impact

A collapsed continuation is not merely an inaccurate level — it reverses decisions. The
audit's mutation family reports **7/7 wrong actions** across risk-aversion and scale
combinations: the exact Bellman action is `1` in every case and the current JIT `Q` chooses
`0`.

The trigger is not exotic. `risk_aversion > 1` is the normal Epstein–Zin range, and small
positive continuation values are exactly what occurs near a borrowing constraint — the
region where the policy function matters most.

## Suggested repair

Give `PowerMean` a stable partial-aggregation path across **both** the stochastic-node and
regime lotteries, then invert once, and mirror it in the diagnostics so `Q` and the
diagnostic intermediates cannot disagree. The anchored log-domain design already exists in
`aggregate()`; the work is threading a partial-sum form of it through `Q_and_F`'s two
accumulation loops rather than inventing anything.

Note the ordering constraint that the same round established for F5: under a nonlinear
certainty equivalent the transform must be applied *before* every expectation, over
stochastic state transitions and over regime transitions alike. Any fix must preserve that
while fixing the stability.

## Acceptance criteria

- Every finite mathematical `PowerMean` stays finite, for scalar-only, stateful-only, and
  mixed shock-by-regime lotteries.
- Action ordering is invariant to representable positive rescalings of the continuation
  lottery.
- Values match a high-precision or stable-log oracle across scale, risk aversion, JIT,
  float32 and float64.
- Diagnostics match `Q`.
- Existing certainty-equivalent tests stay green.

## Verification status

Independently reproduced before filing:

- `PowerMean.aggregate` exists and documents this exact failure mode.
- `grep -rn "\.aggregate(" src/` returns **no callers**.
- The exact CE values were recomputed here with `decimal` at 60 digits and match the
  audit's numbers to all printed digits.
- The float32 overflow chain was reproduced directly.

Not independently re-run here: the 7/7 action-reversal mutation family (the audit's own
execution).

---

## Artifacts

From the round-4 closure audit of #390. The full archive lives at
`whole-branch-round4/pro-bundle/reply/whole-branch-round4-closure-audit-artifacts.zip`;
the F6-relevant files are inlined below so the issue stands alone.

<details><summary><code>oracle_round4.py</code></summary>

```python
"""Independent scalar-target, process-entry, and power-mean oracles."""
from __future__ import annotations

from decimal import Decimal, getcontext
import math

getcontext().prec = 100


def stateless_action_q(*, utility: float, probability: float, value: float) -> float:
    return utility + probability * value


def geometric_mean(values: tuple[float, ...], weights: tuple[float, ...]) -> float:
    return math.exp(sum(w * math.log(v) for v, w in zip(values, weights, strict=True)))


def stable_power_mean(
    values: tuple[float, ...], weights: tuple[float, ...], risk_aversion: float
) -> float:
    """High-precision weighted power mean, stable for tiny positive values."""
    vals = tuple(Decimal(str(v)) for v in values)
    ws = tuple(Decimal(str(w)) for w in weights)
    ra = Decimal(str(risk_aversion))
    exponent = Decimal(1) - ra
    if exponent == 0:
        # Decimal has ln/exp in current Python.
        return float(sum(w * v.ln() for v, w in zip(vals, ws, strict=True)).exp())
    logs = tuple(v.ln() for v in vals)
    anchor = min(logs) if exponent < 0 else max(logs)
    total = sum(
        w * (exponent * (log_v - anchor)).exp()
        for w, log_v in zip(ws, logs, strict=True)
    )
    return float((anchor + total.ln() / exponent).exp())


def iid_expectation(values: tuple[float, ...], weights: tuple[float, ...]) -> float:
    return sum(v * w for v, w in zip(values, weights, strict=True))


def main() -> None:
    assert stateless_action_q(utility=-1.0, probability=1.0, value=10.0) == 9.0
    assert abs(geometric_mean((1.0, 9.0), (0.5, 0.5)) - 3.0) < 1e-15
    pm = stable_power_mean((1e-8, 2e-8), (0.5, 0.5), 8.0)
    assert 1e-8 < pm < 2e-8
    iid = iid_expectation((2.0, 4.0, 8.0), (0.25, 0.5, 0.25))
    assert iid == 4.5
    assert max(range(2), key=lambda i: (4.0, iid)[i]) == 1
    assert max(range(2), key=lambda i: (4.0, 0.0)[i]) == 0
    print("oracle stateless Q=9; geometric CE=3; tiny power mean=", pm, "; IID E=4.5; process action exact/omitted=1/0")


if __name__ == "__main__":
    main()
```

</details>

<details><summary><code>test_round4_power_mean_contract.py</code></summary>

```python
"""Static contract check for PowerMean aggregation in the GridSearch Bellman path."""
from pathlib import Path

CURRENT = Path('/mnt/data/whole_branch_r4/input/pro-bundle/project')


def test_power_mean_documents_stable_aggregate_but_q_does_not_dispatch_to_it():
    ce = (CURRENT / 'src/_lcm/certainty_equivalent.py').read_text()
    qaf = (CURRENT / 'src/_lcm/regime_building/Q_and_F.py').read_text()
    assert 'def aggregate(' in ce
    assert 'The naive `inverse(sum(w · transform(v)))`' in ce
    assert '.aggregate(' not in qaf
    assert 'ce.transform(' in qaf and 'ce.inverse(' in qaf
```

</details>

<details><summary><code>test_round4_power_mean_mutations.py</code></summary>

```python
"""Mutation family for stable scalar PowerMean aggregation and decisions."""
from __future__ import annotations

from pathlib import Path
from types import MappingProxyType

import jax
jax.config.update('jax_enable_x64', True)
import jax.numpy as jnp
import numpy as np
import pytest

from oracle_round4 import stable_power_mean
from q_and_f_loader import PowerCE, load_q_and_f

CURRENT = Path('/mnt/data/whole_branch_r4/input/pro-bundle/project')

CASES = [
    (jnp.float32, 8.0, 1e-8),
    (jnp.float32, 12.0, 1e-5),
    (jnp.float32, 20.0, 1e-3),
    (jnp.float64, 8.0, 1e-50),
    (jnp.float64, 12.0, 1e-30),
    (jnp.float64, 20.0, 1e-20),
    (jnp.float64, 50.0, 1e-8),
]


def _q(module, dtype, low_probs, high_probs):
    return module.get_Q_and_F(
        flat_param_names=frozenset(),
        functions=MappingProxyType({'H': lambda utility, E_next_V: utility + E_next_V}),
        constraints=MappingProxyType({}),
        period_targets=(),
        scalar_targets=('low', 'high'),
        transitions=MappingProxyType({}),
        stochastic_transition_names=frozenset(),
        compute_regime_transition_probs=lambda **kw: MappingProxyType({
            'low': jnp.asarray(low_probs, dtype=dtype),
            'high': jnp.asarray(high_probs, dtype=dtype),
        }),
        regime_to_v_interpolation_info=MappingProxyType({}),
        certainty_equivalent=PowerCE(),
    )


@pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
    values = (scale, 2.0 * scale)
    probs0 = (0.9, 0.1)
    probs1 = (0.5, 0.5)
    ce0 = stable_power_mean(values, probs0, risk_aversion)
    ce1 = stable_power_mean(values, probs1, risk_aversion)
    assert ce1 > ce0 > 0.0
    utility0 = 0.4 * (ce1 - ce0)
    exact_q = np.asarray([utility0 + ce0, ce1])
    assert int(np.argmax(exact_q)) == 1

    module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
    q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
    got = jax.jit(lambda: q(
        next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
        U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
        certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
    )[0])()
    np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
    assert int(jnp.argmax(got)) == 1
```

</details>

<details><summary><code>run_round4_counterexample_matrix.py</code></summary>

```python
"""Print decisive round-4 counterexample outcomes from the audited source."""
from __future__ import annotations

from pathlib import Path
from types import MappingProxyType

import jax
jax.config.update("jax_enable_x64", True)
import jax.numpy as jnp
import numpy as np

from oracle_round4 import stable_power_mean
from q_and_f_loader import PowerCE, load_q_and_f
from test_round4_process_target_contract import _current_target_partition

CURRENT = Path('/mnt/data/whole_branch_r4/input/pro-bundle/project')


def q_for(dtype, probs0, probs1):
    module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'matrix_{dtype}_{probs0}_{probs1}')
    return module.get_Q_and_F(
        flat_param_names=frozenset(),
        functions=MappingProxyType({'H': lambda utility, E_next_V: utility + E_next_V}),
        constraints=MappingProxyType({}),
        period_targets=(),
        scalar_targets=('low', 'high'),
        transitions=MappingProxyType({}),
        stochastic_transition_names=frozenset(),
        compute_regime_transition_probs=lambda **kw: MappingProxyType({
            'low': jnp.asarray([probs0[0], probs1[0]], dtype=dtype),
            'high': jnp.asarray([probs0[1], probs1[1]], dtype=dtype),
        }),
        regime_to_v_interpolation_info=MappingProxyType({}),
        certainty_equivalent=PowerCE(),
    )


def main():
    carry, scalar = _current_target_partition()
    print(f'coarse_process_target carry={carry} scalar={scalar} exact_iid_E=4.5')
    cases = [
        (jnp.float32, 8.0, 1e-8),
        (jnp.float32, 12.0, 1e-5),
        (jnp.float32, 20.0, 1e-3),
        (jnp.float64, 8.0, 1e-50),
        (jnp.float64, 12.0, 1e-30),
        (jnp.float64, 20.0, 1e-20),
        (jnp.float64, 50.0, 1e-8),
    ]
    probs0, probs1 = (0.9, 0.1), (0.5, 0.5)
    wrong = 0
    for dtype, gamma, scale in cases:
        values = (scale, 2 * scale)
        ce0 = stable_power_mean(values, probs0, gamma)
        ce1 = stable_power_mean(values, probs1, gamma)
        utility0 = 0.4 * (ce1 - ce0)
        expected = np.asarray([utility0 + ce0, ce1], dtype=np.dtype(dtype))
        q = q_for(dtype, probs0, probs1)
        got = np.asarray(jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({
                'low': jnp.asarray(values[0], dtype=dtype),
                'high': jnp.asarray(values[1], dtype=dtype),
            }),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(gamma, dtype=dtype),
        )[0])())
        exact_action = int(np.argmax(expected))
        got_action = int(np.argmax(got))
        wrong += got_action != exact_action
        print(
            f'{np.dtype(dtype).name} gamma={gamma:g} scale={scale:.1e} '
            f'exact={expected.tolist()} got={got.tolist()} '
            f'actions={exact_action}/{got_action}'
        )
    print(f'wrong_actions={wrong}/{len(cases)}')


if __name__ == '__main__':
    main()
```

</details>

<details><summary><code>test-results-power.log</code></summary>

```
===== power =====
FFFFFFF                                                                  [100%]
=================================== FAILURES ===================================
___ test_finite_power_mean_and_action_order_survive_scale[float32-8.0-1e-08] ___

dtype = <class 'jax.numpy.float32'>, risk_aversion = 8.0, scale = 1e-08

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.1028628e-08
E       Max relative difference among violations: 1.
E        ACTUAL: array([3.512929e-10, 0.000000e+00], dtype=float32)
E        DESIRED: array([1.050169e-08, 1.102863e-08], dtype=float32)

test_round4_power_mean_mutations.py:66: AssertionError
__ test_finite_power_mean_and_action_order_survive_scale[float32-12.0-1e-05] ___

dtype = <class 'jax.numpy.float32'>, risk_aversion = 12.0, scale = 1e-05

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.0649938e-05
E       Max relative difference among violations: 1.
E        ACTUAL: array([2.214982e-07, 0.000000e+00], dtype=float32)
E        DESIRED: array([1.031769e-05, 1.064994e-05], dtype=float32)

test_round4_power_mean_mutations.py:66: AssertionError
__ test_finite_power_mean_and_action_order_survive_scale[float32-20.0-0.001] ___

dtype = <class 'jax.numpy.float32'>, risk_aversion = 20.0, scale = 0.001

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 0.00103715
E       Max relative difference among violations: 1.
E        ACTUAL: array([1.26377e-05, 0.00000e+00], dtype=float32)
E        DESIRED: array([0.001018, 0.001037], dtype=float32)

test_round4_power_mean_mutations.py:66: AssertionError
___ test_finite_power_mean_and_action_order_survive_scale[float64-8.0-1e-50] ___

dtype = <class 'jax.numpy.float64'>, risk_aversion = 8.0, scale = 1e-50

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.10286274e-50
E       Max relative difference among violations: 1.
E        ACTUAL: array([3.512929e-52, 0.000000e+00])
E        DESIRED: array([1.050169e-50, 1.102863e-50])

test_round4_power_mean_mutations.py:66: AssertionError
__ test_finite_power_mean_and_action_order_survive_scale[float64-12.0-1e-30] ___

dtype = <class 'jax.numpy.float64'>, risk_aversion = 12.0, scale = 1e-30

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.06499383e-30
E       Max relative difference among violations: 1.
E        ACTUAL: array([2.214982e-32, 0.000000e+00])
E        DESIRED: array([1.031769e-30, 1.064994e-30])

test_round4_power_mean_mutations.py:66: AssertionError
__ test_finite_power_mean_and_action_order_survive_scale[float64-20.0-1e-20] ___

dtype = <class 'jax.numpy.float64'>, risk_aversion = 20.0, scale = 1e-20

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.03715494e-20
E       Max relative difference among violations: 1.
E        ACTUAL: array([1.26377e-22, 0.00000e+00])
E        DESIRED: array([1.018198e-20, 1.037155e-20])

test_round4_power_mean_mutations.py:66: AssertionError
__ test_finite_power_mean_and_action_order_survive_scale[float64-50.0-1e-08] ___

dtype = <class 'jax.numpy.float64'>, risk_aversion = 50.0, scale = 1e-08

    @pytest.mark.parametrize(('dtype', 'risk_aversion', 'scale'), CASES)
    def test_finite_power_mean_and_action_order_survive_scale(dtype, risk_aversion, scale):
        values = (scale, 2.0 * scale)
        probs0 = (0.9, 0.1)
        probs1 = (0.5, 0.5)
        ce0 = stable_power_mean(values, probs0, risk_aversion)
        ce1 = stable_power_mean(values, probs1, risk_aversion)
        assert ce1 > ce0 > 0.0
        utility0 = 0.4 * (ce1 - ce0)
        exact_q = np.asarray([utility0 + ce0, ce1])
        assert int(np.argmax(exact_q)) == 1
    
        module = load_q_and_f(CURRENT / 'src/_lcm/regime_building/Q_and_F.py', f'qaf_pm_mut_{dtype}_{risk_aversion}_{scale}')
        q = _q(module, dtype, [probs0[0], probs1[0]], [probs0[1], probs1[1]])
        got = jax.jit(lambda: q(
            next_regime_to_V_arr=MappingProxyType({'low': jnp.asarray(values[0], dtype=dtype), 'high': jnp.asarray(values[1], dtype=dtype)}),
            U_arr=jnp.asarray([utility0, 0.0], dtype=dtype),
            certainty_equivalent__risk_aversion=jnp.asarray(risk_aversion, dtype=dtype),
        )[0])()
>       np.testing.assert_allclose(np.asarray(got), exact_q.astype(np.dtype(dtype)), rtol=8e-5, atol=0)
E       AssertionError: 
E       Not equal to tolerance rtol=8e-05, atol=0
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference among violations: 1.01424639e-08
E       Max relative difference among violations: 1.
E        ACTUAL: array([4.837544e-11, 0.000000e+00])
E        DESIRED: array([1.006990e-08, 1.014246e-08])

test_round4_power_mean_mutations.py:66: AssertionError
=========================== short test summary info ============================
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float32-8.0-1e-08] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.1028628e-08
Max relative difference among violations: 1.
 ACTUAL: array([3.512929e-10, 0.000000e+00], dtype=float32)
 DESIRED: array([1.050169e-08, 1.102863e-08], dtype=float32)
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float32-12.0-1e-05] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.0649938e-05
Max relative difference among violations: 1.
 ACTUAL: array([2.214982e-07, 0.000000e+00], dtype=float32)
 DESIRED: array([1.031769e-05, 1.064994e-05], dtype=float32)
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float32-20.0-0.001] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 0.00103715
Max relative difference among violations: 1.
 ACTUAL: array([1.26377e-05, 0.00000e+00], dtype=float32)
 DESIRED: array([0.001018, 0.001037], dtype=float32)
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float64-8.0-1e-50] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.10286274e-50
Max relative difference among violations: 1.
 ACTUAL: array([3.512929e-52, 0.000000e+00])
 DESIRED: array([1.050169e-50, 1.102863e-50])
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float64-12.0-1e-30] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.06499383e-30
Max relative difference among violations: 1.
 ACTUAL: array([2.214982e-32, 0.000000e+00])
 DESIRED: array([1.031769e-30, 1.064994e-30])
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float64-20.0-1e-20] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.03715494e-20
Max relative difference among violations: 1.
 ACTUAL: array([1.26377e-22, 0.00000e+00])
 DESIRED: array([1.018198e-20, 1.037155e-20])
FAILED test_round4_power_mean_mutations.py::test_finite_power_mean_and_action_order_survive_scale[float64-50.0-1e-08] - AssertionError: 
Not equal to tolerance rtol=8e-05, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.01424639e-08
Max relative difference among violations: 1.
 ACTUAL: array([4.837544e-11, 0.000000e+00])
 DESIRED: array([1.006990e-08, 1.014246e-08])
7 failed in 0.48s
EXIT_CODE=1
```

</details>

<details><summary><code>test-results-power-contract.log</code></summary>

```
.                                                                        [100%]
1 passed in 0.02s
```

</details>

<details><summary><code>counterexample_matrix.log</code></summary>

```
coarse_process_target carry=() scalar=() exact_iid_E=4.5
float32 gamma=8 scale=1.0e-08 exact=[1.0501688407771326e-08, 1.1028627788789436e-08] got=[3.5129288367130584e-10, 0.0] actions=1/0
float32 gamma=12 scale=1.0e-05 exact=[1.0317690794181544e-05, 1.0649938303686213e-05] got=[2.2149824019379594e-07, 0.0] actions=1/0
float32 gamma=20 scale=1.0e-03 exact=[0.0010181984398514032, 0.0010371549287810922] got=[1.2637702639040072e-05, 0.0] actions=1/0
float64 gamma=8 scale=1.0e-50 exact=[1.0501688088354014e-50, 1.102862741485982e-50] got=[3.5129288433720444e-52, 0.0] actions=1/0
float64 gamma=12 scale=1.0e-30 exact=[1.0317690906484091e-30, 1.064993825700262e-30] got=[2.2149823367902014e-32, 0.0] actions=1/0
float64 gamma=20 scale=1.0e-20 exact=[1.0181983857297784e-20, 1.0371549403296514e-20] got=[1.2637703066582003e-22, 0.0] actions=1/0
float64 gamma=50 scale=1.0e-08 exact=[1.0069900715716124e-08, 1.0142463869673272e-08] got=[4.8375435971432627e-11, 0.0] actions=1/0
wrong_actions=7/7
```

</details>

hmgaudecker:
Fixes #412.

Epstein-Zin models could return **silently wrong answers** — no error, no `NaN`, just
a different optimal policy than the one your model implies. This PR fixes that, and
gives the two objects that define a recursive preference their own places in the model
specification.

## What could go wrong in your results

If you used `GridSearch` with a `PowerMean` certainty equivalent and risk aversion
above 1, the solver could compute a certainty equivalent of exactly zero for a lottery
whose true value was perfectly ordinary. The value function stayed finite and
plausible-looking, so nothing alerted you — but the agent then chose a **different
action** than your model prescribes.

It bit hardest exactly where Epstein-Zin models are usually interesting: **high risk
aversion, with continuation values near the borrowing constraint**, and where one
branch of the lottery carries small probability. A joint probability of `1e-4` is
completely ordinary once the lottery is spread across regime transitions and
stochastic states, and that was already enough to exceed the accuracy budget in single
precision.

The same defect appeared a second time, one level out, in the Epstein-Zin aggregator
itself (details below). Both are fixed.

**If you ran Epstein-Zin models with risk aversion above 1, results are worth
re-checking.** Expected-utility models are unaffected — their arithmetic is unchanged.

## Why it happened

The certainty equivalent for a power mean is

$$CE = \left(\sum_r w_r \, v_r^{\,1-\gamma}\right)^{1/(1-\gamma)}$$

The old code computed this literally, in that order. With risk aversion $\gamma > 1$
the exponent $1-\gamma$ is negative, so a small continuation value $v_r$ raised to it
becomes an enormous number — large enough to exceed the biggest number the computer
can represent, *before* the summation that would have brought it back to a sensible
scale. Once that happens the information is gone, and the final step returns zero.

The repair is the standard trick, applied properly: factor out the most extreme term
before doing anything else, and work with logarithms in between. Written that way,
every intermediate quantity stays between 0 and 1, so nothing can blow up. The result
is mathematically identical — it is the same formula, evaluated in an order that a
finite-precision machine can actually follow.

Two details made this more than a one-line change:

- **The whole lottery has to be assembled first.** You cannot factor out the extreme
  term if you have already collapsed each regime separately, because the extreme term
  is a property of the *joint* lottery. The engine now hands the complete lottery —
  every stochastic state of every reachable regime, weighted by its transition
  probability — to a single aggregation step. (The economics is untouched: the
  transform is still applied before every expectation, as recursive utility requires.)
- **The sum needs two forms, not one.** One arrangement is accurate when risk aversion
  is close to 1 (where the power mean approaches a geometric mean); the other is
  accurate when the lottery puts very little weight on its extreme branch. Neither
  covers both, so the code picks per lottery. Using only one silently loses the case
  it does not cover.

## The same bug, one level out

The Epstein-Zin recursion contains **two** power means, and the first version of this
PR fixed only the inner one. The outer aggregator

$$W = \left[(1-\beta)\,u^{\rho} + \beta\,CE^{\rho}\right]^{1/\rho},
\qquad \rho = 1 - 1/\psi$$

is *also* a weighted power mean — over $(u, CE)$ with weights $(1-\beta, \beta)$ — and
it was still evaluated literally. It carried the same two failure modes:

| case | returned | correct |
| --- | --- | --- |
| $\psi = 0.125$, both arguments `1e-50` (float64) | `0.0` | `1e-50` |
| $\psi = 0.125$, both arguments `1e-8` (float32) | `0.0` | `1e-8` |
| $\psi \approx 1$, two actions (float32) | picks **A** | correct choice is **B** |

The first two are the overflow above. The third is the near-unit-exponent case: at
$\psi = 1$ the CES form collapses to Cobb-Douglas, and evaluating it literally near
that point loses enough precision to reverse the ranking of two actions.

Since both objects are the same mathematical thing at different exponents, they now
share one implementation. A parameter range that one survives, the other survives too.

### Two new slots

The certainty equivalent and the Koopmans aggregator now sit in matching, explicitly
named slots on both `Regime` and `Model`:

```python
Model(
    regimes=...,
    certainty_equivalent=PowerMean(),        # default: LinearExpectation()
    koopmans_aggregator=CESAggregator(),     # default: LinearAggregator()
)
```

Declare them at the model level and they reach every non-terminal regime. Declare one
on a regime instead and **every** non-terminal regime must declare it — mixing the two
is rejected, because the silent regimes would look like they had opted out of something
they never saw. Terminal regimes have no continuation and receive neither.

### The aggregator moves out of `functions`

Previously the aggregator lived in `functions` under the name `H`, which was never
quite true — `functions` otherwise means "callables whose arguments pylcm resolves for
you", and this one needed nine special cases across the engine to keep it out of that
machinery. It now has its own slot, and `functions` has no exceptions.

`Phased(solve=..., simulate=...)` still works on the slot, so phase-varying
aggregators — the naive beta-delta agent, for instance — are unaffected.

### Renames

The symbol follows the paper's notation ($W$ for the Koopmans aggregator):

| before | after |
| --- | --- |
| `H_linear`, `H_epstein_zin` | `LinearAggregator()`, `CESAggregator()` |
| `lcm.temporal_aggregation` | `lcm.koopmans_aggregation` |
| `E_next_V` (continuation argument) | `CE` |
| `template[regime]["H"]` | `template[regime]["koopmans_aggregator"]` |

`CE` is now a reserved name: a state, action, function, or derived categorical called
`CE` would be shadowed inside the aggregator and nowhere else, so it is rejected when
the model is built rather than causing confusion later.

## Two smaller correctness fixes

**Expected utility now normalises by the probability mass it actually received.** The
fast path for `LinearExpectation` accumulated $\sum_r p_r E[V_r]$ where the definition
is $\sum_r p_r E[V_r] / \sum_r p_r$. These agree when the transition probabilities sum
to exactly one, and transition validation accepted anything within a loose tolerance.
At the top of that tolerance the two disagreed enough to reverse the optimal action.
Models whose probabilities really do sum to one are bit-for-bit unaffected.

**Subclasses of `LinearExpectation` are no longer ignored.** The engine takes a cheaper
route for the linear case. That route is now selected by exact type, so if you subclass
`LinearExpectation` to state a different quantity, your version actually runs.

## Accuracy, measured

Relative error against a 180-digit reference, for a two-point lottery `(1e-3, 1.0)` at
risk aversion 8, as the bad branch becomes less likely:

| P(bad branch) | before (float64) | after (float64) | before (float32) | after (float32) |
| --- | --- | --- | --- | --- |
| 1e-2 | 0 | 0 | 1.7e-07 | 3.1e-07 |
| 1e-4 | 1.6e-14 | 3.5e-16 | **2.4e-05** | 7.6e-09 |
| 1e-6 | 4.1e-12 | 3.6e-16 | **1.9e-03** | 2.0e-07 |
| 1e-8 | 7.2e-10 | 1.3e-16 | **infinite** | 4.2e-07 |
| 1e-12 | 3.2e-06 | 0 | **infinite** | 1.1e-07 |
| 1e-17 | **infinite** | 4.1e-16 | — | — |

Read the bold entries as: a 2.4% error returned *silently*, then an infinite one.
Accuracy budgets are `1e-12` (double precision) and `5e-5` (single).

The old form degraded by roughly one correct digit per decade of probability, because
it computed a quantity of the form $1 + \tilde{D}$ where $\tilde{D}$ was itself very
close to $-1$ — the classic catastrophic cancellation.

## Performance

Faster than what it replaces, on the Epstein-Zin benchmark model:

| | float32 | float64 |
| --- | --- | --- |
| new aggregator vs. the literal CES form | **-79.5%** | **-15.7%** |
| expected-utility solve, vs. merge base | **-24.6%** | **-13.1%** |

Solving with a `PowerMean` certainty equivalent costs more than before (+9% single,
+36% double warm time) — that is the price of evaluating it correctly, and it is
recorded as an accepted cost. Compilation is up across all arms and stays well inside
budget in absolute terms — an order of magnitude under the 5 s ceiling on every arm
of the benchmark model.

## Missing regime mass now fails loudly

If a regime transition puts probability on a regime that is not active in the next
period, that probability used to be dropped and the rest renormalised — so the value
function came back finite, plausible, and *independent of the missing mass*. A model
whose survival probability ranged from 0.999999 to 0.000001 could return bit-identical
values. Runtime validation caught it, but validation is controlled by `log_level`, and
`log_level="off"` skips it entirely — which is exactly what an estimation loop runs.

The check now lives in the arithmetic, where no log level can skip it: if the mass a
regime's continuation represents is more than `1e-3` away from one, the continuation
becomes `NaN`. Deliberately a coarse backstop against a wrong model rather than a
numerical check — smaller mass errors are still the `debug`-level validator's business,
and models whose probabilities really do sum to one are bit-for-bit unchanged, verified
by checksum over the whole solved value function at production grid sizes in both
precisions.

**This can turn a model that used to solve into one that returns `NaN`.** Two shapes
reach it:

- transition probabilities that do not sum to one within `1e-3`;
- a non-terminal regime that is still active in a period where *none* of its targets
  is — it emits unit mass toward regimes that cannot receive it, so its continuation
  carries no mass at all. Previously that returned a plausible utility-only value.

In both cases the model was already misspecified; only the reporting changed.

**Run every model once at `log_level="debug"` with your real parameters** before
trusting its output. After that, `"off"` is a reasonable choice inside an estimation
loop. More generally: unless a failure is *very* obvious, do not reason about it from a
run at anything below `"debug"` — this defect is the worked example, since nothing in
the results shows where the missing mass went. This is documented in the debugging
guide, the solving guide, and the introductory example.

## Verification

- Both power means checked against 60- and 180-digit arithmetic across risk aversions,
  elasticities of intertemporal substitution (including $\psi$ exactly 1 and either
  side of it), and discount factors at 0 and 1.
- Action-ordering tests: the optimal choice is compared against an exact reference, not
  just the value.
- Scale-equivariance of solve and simulate at both precisions.
- The expected-utility fast path is pinned by differential tests against the definition
  it claims to compute.
- Independently audited over nine review rounds against a formal numerical contract,
  closing `clean`.


timmens (commented):
I like the solution, very nice! I have mainly two blocking comments before approval:

1. You write `_get_compute_E_next_V`, but it should be `_get_compute_certainty_equivalent`, right? This would also entail renaming the public `E_next_V` argument of `H` to `CE` or something like that. 
2. Several functions allow `certainty_equivalent = None` which default to the standard expectation. I would just make this explicit.

timmens (approved):

timmens:
This looks very good overall. One aspect of the public API still feels uneven to me: the Koopmans aggregator is supplied as a function such as `W_linear`, while the certainty equivalent is supplied as a named specification object such as `LinearExpectation()` or `PowerMean()`.

I understand the internal reason: the Koopmans aggregator participates in the function DAG, whereas the certainty equivalent describes how an entire continuation lottery is reduced. Still, users are currently exposed to that implementation distinction.

I see two possible improvements.

The more economics-oriented interface would treat the two components as one recursive preference specification:

```python
Model(..., preferences=ExpectedUtility())  # default
Model(..., preferences=EpsteinZin())
```

with an advanced compositional escape hatch:

```python
preferences=RecursivePreferences(
    koopmans_aggregator=my_W,
    certainty_equivalent=QuasiArithmeticMean(...),
)
```

Internally, this could lower to the existing two components without changing the numerical architecture. It would also make the standard pairings explicit:

- expected utility: additive Koopmans aggregator + linear expectation;
- Epstein–Zin: CES Koopmans aggregator + power-mean certainty equivalent.

A smaller compromise would be to retain the existing two slots but make the built-in Koopmans aggregators callable specification objects as well:

```python
koopmans_aggregator=AdditiveAggregator()
certainty_equivalent=LinearExpectation()
```

and:

```python
koopmans_aggregator=CESAggregator()
certainty_equivalent=PowerMean()
```

I would prefer `AdditiveAggregator` and `CESAggregator` over `LinearAggregator` and `EpsteinZinAggregator`: these names describe exactly what the outer component computes, without implying that `CESAggregator()` alone constitutes Epstein–Zin preferences. Custom callables and `Phased(...)` aggregators could remain supported.

I would also retain the name `koopmans_aggregator`. “Temporal aggregator” makes the contrast with lottery aggregation intuitive, but `koopmans_aggregator` is precise, economist-facing, and already used consistently in this PR.

This is not only cosmetic. For example, `W_epstein_zin` together with `LinearExpectation()` is not the expected-utility recursion; expected utility requires `W_linear` together with `LinearExpectation()`. A more uniform API—or ideally named preference presets—would make such interpretations less likely.

One related notebook correction: the Epstein–Zin recursion collapses to expected CRRA utility when

$$
1-\gamma=\rho=1-\frac{1}{\psi},
$$

hence when $\gamma=1/\psi$, not when $\gamma=\rho$ as currently stated.

hmgaudecker:
Thank you — makes perfect sense. As discussed, I've taken the smaller compromise: `koopmans_aggregator=LinearAggregator()` alongside `certainty_equivalent=LinearExpectation()`, and `CESAggregator()` alongside `PowerMean()`. `koopmans_aggregator` stays, for the reasons you give; custom callables and `Phased(...)` are unaffected, and that path now has the end-to-end test coverage it was missing.

For the record, we went with `Linear` rather than `Additive` since "additive" reads as *time-additive* in this literature, but `AdditiveAggregator()` with `PowerMean()` is Kreps–Porteus, not additive preferences. Basically the same trap elsewhere over from the one you're pointing at. `LinearAggregator` claims only linearity in `(utility, CE)`, and is exactly the `ρ = 1` member of the family `CESAggregator` parameterizes: the forms differ only by the normalization `(1, β)` vs `(1 − β, β)`, and since every shipped certainty equivalent is homogeneous of degree 1, `Ṽ = (1 − β)V` satisfies the one recursion iff `V` satisfies the other — same policies, levels differing by that factor.

There is a `KoopmansAggregator` base, deliberately thin: no `param_names` property mirroring `CertaintyEquivalent`, because an aggregator's *signature* is the sole declaration of what it consumes. That is what lets a parameter name matching a regime function be wired from the DAG rather than the params template.


Two further changes on the branch while I was in here, both with tests:

- Regime-transition mass that goes missing — probability placed on a target that isn't active next period — used to be divided straight back out by the normalization on every aggregation route, so the value function came back finite, plausible, and *independent of the missing mass*: a model at total mass 0.977 and one at 1.0 returned bit-identical values. `validate_transitions` caught it, but `log_level="off"` skips validation, which is exactly what an estimation loop runs. The check now lives in the arithmetic — a represented mass more than `1e-3` from one turns the continuation into NaN, at every log level. Deliberately a coarse backstop against a wrong model rather than a numerical check; smaller mass errors remain the `debug`-level validator's business. Well-formed models are bit-unchanged, verified by checksum over the whole solved value function at production grid sizes in both precisions.
- A `Series`-valued parameter indexed inside a callable-object aggregator now resolves like the function form: the engine reads the subscript order from `__call__` when the object's own source can't be read.

hmgaudecker:
@timmens, this is good to be merged from my side. Do you want to have a last check?

Note that I reorganised the ci workflows to make them easier to stop via the api in a granular fashion (=broke up main into its components and made them required checks for merging). No changes to content, except that we run ty on the docs notebooks now, caught a few things.

timmens:
Looks good to me. Feel free to merge!

PR #418, opened 2026-08-04
Price entry into a stochastic process the source does not carry, and never enlarge a probability the format cannot hold

hmgaudecker:
A regime whose state is a stochastic process can be entered without the source regime
carrying that process. The continuation over such a target is weighted by the process's
own probabilities, and no event in that continuation is ever weighted by a probability
the model did not specify.

## Entry into a target-only process

On the base branch such an entry is averaged uniformly over the process's nodes: the
quadrature weights are built and then discarded, because stochasticity is decided from a
model-wide set of transition names built from the *source* regime's states, and a process
the source does not carry is absent from that set by construction.

Here, which axes of a continuation are stochastic is decided per `(target, next_state)`
pair, from the synthesized function set: a law is stochastic exactly when a weight
function was built for it. The engine can therefore neither price a law by a weight that
was never built, nor build a weight that nothing consumes.

The magnitude at stake is not small. With a Gauss-Hermite 3-node normal — probabilities
`(1/6, 2/3, 1/6)` — and payoff `shock²`, the branch point prices entry at `1.0`, the
probability-weighted value, against `2.0` for the unweighted mean over the nodes. Every
test of this behaviour uses a nonuniform law with a nonlinear or asymmetric payoff;
`UniformIIDProcess` and linear payoffs on symmetric grids serve as negative controls,
since the weighted and unweighted means coincide there and such a case is evidence of
nothing.

## Declared entry

A **declared entry** names one physical value: the agent arrives at the target holding a
particular level of that process's state. That is a different object from an intrinsic
draw, which is a genuine lottery. The two coincide under expected utility and diverge
under any nonlinear aggregator, so the engine keeps them distinct.

A declared value is read by interpolating the target's value function at the physical
coordinate. Cost is independent of the entered processes' node counts: a model declaring
entry into three 25-node processes — a 15,625-point product surface — is indistinguishable
in runtime, compile time and memory from one entering a single 25-node process.

## What the weights guarantee

An expectation is only as good as its weights, and each of the following would otherwise
produce a plausible number rather than a complaint.

**An impossible event never destroys a possible one.** A law evaluated at a node the draw
gives zero probability may name anything, including a value outside the target's grid,
which the interpolator reports as undefined. Probability says such a node contributes
nothing whatever it names; floating-point arithmetic says `0 × undefined` is undefined and
propagates it through the whole expectation. Zero-probability nodes are neutralised at the
engine boundary, before any certainty equivalent sees them, so the guarantee holds for a
user-written aggregator as much as for the ones shipped here — and inside each shipped
aggregator too, so calling one directly gives the same answer as routing through the
engine.

**Weights that are not probabilities are refused, not read as impossible events.** A
liveness test of the form "weight > 0" silently discards negative weights, which are
misspecifications rather than unlikely events: a transition placing `1.5` on one target
and `−0.5` on another sums to one and would otherwise be published as a well-formed
answer. Unit mass and non-negativity are both required, and both are arithmetic rather
than diagnostics, so they hold with logging off.

**An event too rare for the format to hold is still priced at its own size.** A joint
node's probability is the product of one factor per stochastic dimension. Each factor can
sit comfortably inside the representable range while the product falls below it — in
single precision, two probabilities of about `5×10⁻²⁰` — at which point the hardware
returns exactly zero, and zero is how the engine spells "cannot happen".

Preserving that product as a subnormal number does not work, and not for want of range: a
subnormal that exists only as an intermediate cannot be classified inside a fused region
at all. In one compiled function the same expression answers "this is not zero" when the
boolean is the output and selects the substituted branch when it is only a predicate, and
an optimisation barrier does not change it. Nothing the code can test about the encoded
float survives the fusion.

So the scale travels **beside** the number rather than inside it. A joint product leaves
as a pair `(w, shift)` standing for `w · 2⁻ˢʰⁱᶠᵗ`, with `w` normal and the shift chosen
where the product is formed; a consumer that normalises by its own total can ignore the
shift entirely, and one that concatenates weights from several producers reconciles them
first — which lifts each toward the largest scale present and lets an entry that cannot
reach it keep the residual of its own, so no entry is read against a scale it never
moved to. A weighted mean depends on its weights only through their ratios and
a power of two rescales them exactly, so nothing is approximated by this. A node of
probability `2⁻¹²⁸` against a continuation of `2¹²⁶` in single precision prices at its
exact quarter.

Where the plain float interface is still the right one — a single joint weight handed to a
caller — the underflow test is read off the **integer exponent** the product would need,
never off the encoded float.

**A weight is judged against the value the mean actually averages.** A certainty
equivalent with a nonlinear transform — one of the shipped ones under a non-unit exponent,
or a transform a user wrote — applies that transform to the values, and a perfectly
ordinary value can map to an undefined or infinite one. Under `g(v) = 1/v` a continuation
of zero transforms to an infinity, and a rare node standing there carries the answer;
deciding from the supplied value instead reads that node as impossible and returns an
ordinary number, ranking an alternative worth half of it the wrong way round. The same
ordering makes a value outside the transform's own domain — a negative continuation under
`√·` — stay undefined however rare the node is, rather than quietly dropping out. Values
are substituted only where a node genuinely cannot occur, the transform is applied, and
only then is it decided what a weight too small to name was standing against.

**A state where no action is feasible does not destroy the states beside it.** Backward
induction writes `−∞` for such a state, and the next step reads that array back by
interpolation. An interpolation corner of weight exactly zero multiplied by `−∞` is
undefined, and it travels through the sum into every neighbouring read — in two dimensions
a single infeasible state would take out the whole block around it. The corner is
neutralised on the value, as elsewhere in the engine. The test is for a represented zero
rather than for positivity, which matters here specifically: this interpolator
extrapolates rather than clamping, so outside the grid its corner weights are legitimately
negative and discarding them would truncate an extrapolated read rather than drop a null
event.

**Next-period values are visible only where they exist.** Visibility of `next_<state>` is
decided per consumer and per phase, so a function outside a law of motion cannot have that
name bound as a free parameter and solve against a user-supplied constant standing where a
computed continuation belongs.

## Contract users must satisfy

These make the solver less forgiving than the baseline, deliberately — the project has no
legacy models to protect:

- transition probabilities must form a genuine distribution. A specification that does not
  yields undefined values rather than a plausible number, at every logging level;
- `next_<state>` is readable only inside a law of motion. Within-period quantities need
  their own names.

## Limits

- **A lottery whose weights span more binades than the exponent field can express is
  priced only approximately, and that residual is accepted.** One shared scale cannot
  make two such weights normal at once, whatever the scale.

  Reconciling several such weights does not itself lose anything: an entry that cannot
  reach the shared scale keeps the residual of its own, so `w · 2⁻ˢ` is preserved
  exactly whether or not the cap binds, and nothing is either understated or enlarged
  there. The approximation appears one step later, and only for a consumer that has to
  name the weights as plain numbers rather than reduce them in logarithms: it has
  nowhere to put that residual, so the entries come down onto the largest scale every
  one of them reaches and an entry further below it than the format can express
  underflows to zero. That is understatement, the direction the contract allows, and
  never enlargement. Against `±∞` it is not an approximation but a lost answer, so the
  weight is floored at the smallest **normal** magnitude instead — normal rather than
  merely nonzero, because a subnormal operand is precisely what the backend flushes.

  This is a property of a fixed-width exponent field rather than of the implementation,
  and the maintainer accepts it as a declared limit of the approach: **it is not grounds
  to block.** What remains blocking is any lapse *within* the class — a weight enlarged
  beyond what its own scale states, a sign lost so an invalid negative weight reads as a
  null event, an infinity dropped so an unreachable-value state prices as finite, a
  represented zero treated as anything but a null event, or a weight omitted while its
  contribution can still change the sum it enters. Each of those is a defect, not a
  rounding limit.
- **A process whose law is parameterised at runtime cannot be entered.** Entry weights are
  evaluated inside the *source's* problem, which reads only the source's parameters, so a
  law the target parameterises at runtime has no value the source can read. Restating it
  under the source would create a second knob free to disagree with the grid the target's
  value function is built on — the defect this branch closes.
- **Intrinsic entry is bounded by conditioning; declared entry is not.** Arriving with no
  handoff at all uses the target's own unconditional law, and that is the whole
  distribution only for a process whose law depends on nothing. It is therefore available
  for a construction-fixed IID process and refused for a persistent one, whose next draw
  depends on a previous value the source would have to supply. Declared entry carries no
  such bound — naming a physical value works for a persistent target too, because it
  resolves by interpolation rather than by conditioning.
- **Target qualification lives in the representation, not the grammar.** A single state
  with a stochastic law toward one target and a deterministic law toward another is not
  constructible.
- **Carried processes match by name alone.** A source whose shock lives near zero may hand
  to a target whose same-named shock lives near a thousand, and the model builds and solves
  with no diagnostic at `log_level="debug"`. This behaves identically on the base branch,
  so it is reported rather than repaired here; whether a shared name should assert
  compatibility is a contract question for the maintainer.

## Evidence

Oracles are independent of the code under test: Gauss-Hermite nodes from
`numpy.polynomial.hermite.hermgauss`, exact rational arithmetic for the
representable-probability witnesses. A weight the format cannot hold as a normal number is
checked through the scale it carries rather than through its encoding, because the backend
answers a bit-level test of a subnormal differently depending on the surrounding fusion —
and the witnesses are swept across the whole range below the normal one rather than
sampled at the top of it, because the top is where every candidate rule looks harmless.

The interpolation guard is pinned by a decision witness as well as a definedness one: over
random reads spanning a grid's interior and its extrapolated edges, no `argmax` moves.
Agreeing to a few units in the last place is not sufficient for a value function that
feeds every subsequent comparison.

Every guarantee above is pinned by a mutation: the mechanism enforcing it is deleted and a
public test must then fail.

Suites are green at both precisions. Runtime, compile time and peak memory are measured
per workload at both precisions against declared budgets, with an untouched control
workload carried alongside so machine noise is distinguishable from a real regression.

## What the guarantees cost

Warm execution is the metric this branch is judged on, and it is met. On the GPU
benchmark, `aca-baseline` runs at 1.08× main and `aca-baseline-debug` at 1.00×; every
other workload is within noise except the irregular-grid solve-and-simulate at 1.17×.
That figure was not always there. Carrying a scale beside every joint probability
initially cost a factor of two on `aca-baseline`, and two reductions bought it back:
specialising the target-node expectation away from the general exponent-balanced kernel,
then replacing JAX's `ldexp` — which XLA has no native form of, and which lowers through
a general decompose-and-exponentiate graph — with a downward-only edit of the exponent
field.

Two costs remain over their triggers and are accepted rather than met, on the
maintainer's judgement:

- **cold compile time**, between 1.24× and 3.53× depending on workload. This is the graph
  the guarantee costs: a probability the format cannot hold as a normal number travels as
  a normal number and a scale, and every consumer reads it that way. Compile is paid once
  per model shape, and warm solve time did not move with it;
- **peak memory**. On `aca-baseline` the device figure is 1.03× while the host figure is
  1.50× (6.39 GB to 9.56 GB). Smaller workloads show larger device ratios on small
  absolute bases — 8 MB to 34 MB on the precautionary-savings solve — and the
  solve-and-simulate pair rises from 566 MB to 1.68 GB.


hmgaudecker:
@timmens -- finally. there might be a bit of a performance regression, I am currently bisecting that one. If it is real, I'd prefer to tackle it elsewhere.

timmens (approved):
Very nice PR! No comments.

PR #427, opened 2026-08-23
DOC Reorganize documentation around modeling decisions

hmgaudecker:
### What problem do you want to solve?

The existing documentation mixed task-oriented guidance, mathematical explanation,
examples, exact API contracts, and development internals. That made the User Guide too
deep for new users and left the EGM solver family without a coherent authoring journey.

This PR reorganizes the book into six distinct chapters:

- Getting Started
- User Guide
- Concepts & Methods
- Examples
- Reference
- Development

It makes EGM, DCEGM, NEGM, NBEGM, and NNBEGM first-class solver families; adds a
problem-first specialized-solver authoring path; separates solver prerequisites such as
conditions, case pieces, piecewise-affine schedules, and margin declarations into
cross-linked Reference pages; and adds curated specialized consumption-saving and
collective-regime examples.

The change also:

- adds an exact public-API coverage index and guard;
- clarifies certified-envelope payload, ordering, NaN-refusal, and failure contracts;
- documents the shipped NBEGM batching/admission semantics without claiming memory
  controls that do not exist;
- removes obsolete top-level NBEGM and benchmarking pages in favor of the new structure;
- adds mutation-sensitive coverage for public NBEGM partition-request routing.

The documentation and supporting contracts went through five bounded external review
rounds. The final review is closure-clean with no remaining Serious or Moderate finding.

### Todo

- [x] Target the stacked #409 branch (`collective-regimes`).
- [x] Merge the definitive #409 tip (`723f9a26`; review repairs at `90fe284f`, #407 landed on `main` at `b803c810`).
- [x] Land #409 on `main` at `e7e4a83f` and cascade that tip into this branch at `6fb4c87e`.
- [x] Build and cross-reference-check the complete book.
- [x] Verify public API coverage, curated examples, documentation contracts, and NBEGM
  routing in float64 and float32.
- [x] Complete the external documentation audit.
- [ ] No corresponding issue to close.

### Verification

Documentation verification at `810be995`, after merging the Round-9 #409 closure head
`1bff3e16`. Commit `85330260` then merged the #409 three-file Windows certificate
newline-portability follow-up. Commit `6dc03386` merged the #409 repair head `90fe284f`, including review
5064008752 repairs and #407 at `1c9f1d22`. After #407 landed on `main` at
`b803c810`, commit `771e174f` cascades the tree-preserving #409 post-land head
`723f9a26`. PR #409 then landed on `main` at `e7e4a83f`; commit
`6fb4c87e` bridges that squash back into this branch. The bridge is tree-identical to
the verified pre-land head, and its per-step checks pass:

- Post-land `prek run ty --all-files`: passed.
- Post-land collective/Pareto behavioral smoke: 33/33 passed at float64.
- Post-land `pixi install -e tests-cuda12 --locked`: passed with no lockfile change.
- Post-land source-inventory check and unified verifier: passed.
- `prek run --all-files`: exit 0, 34 hooks, every one `Passed`, no file modified.
- `pixi run -e docs build-docs`: exit 0, 70 pages, no MyST or cross-reference errors.
- Post-merge documentation/API/certificate battery: 75 passed at float64, with no failures, errors, or skips.
- Public-API coverage of the source docs: across 71 source `.md`/`.ipynb` files
  (excluding `_build`), every name imported from `lcm` in a fenced Python block or
  notebook code cell resolves against `dir(lcm)`. This checks `from lcm import ...`
  specifically; it does not cover attribute access after a bare `import lcm`, submodule
  imports, or API named only in prose.
- The lowered collective vocabulary (`GatedEdge`, `stakeholders=`, `pareto_objective=`,
  `value_constraints=`, `same_period_refs=`, `gated_edges=`) no longer appears in `docs/`
  or `src/lcm_examples/`. Two references survived the merge and were removed in a
  follow-up commit: the public-API index still listed `lcm.GatedEdge`, and the
  collective-regimes reference still described declaring an edge in either spelling. The
  index entry was a genuine failure of
  `tests/test_docs_public_api_coverage.py::test_public_api_index_covers_exactly_the_lcm_exports`,
  which compares the documented set against `lcm.__all__` in both directions; that test
  is green again (5 passed, 0 failed). `prek` does not cover it, which is why the merge
  passed the hook gate with the test red.
- The exact #409 head passed the complete float32 and float64 profiles with zero failures/errors;
  this branch then adds documentation-only changes, whose affected documentation, example, API,
  and routing seams were exercised directly.
- Focused verification of the final combined production topology is green: float32 193
  passed / 5 skipped; float64 198 passed; zero failures or errors in either profile.
  The unified verifier passes and the synchronized self-test rejects all 176/176
  direct-flow mutations.
- A fresh search of `docs/` and `src/lcm_examples/` found no stale replay semantics or
  public use of the lowered edge vocabulary. The remaining `GatedEdge` mention explains
  that it is engine-internal, so no documentation change is required for this merge.

### Merge resolution worth noting

Merging #409 into this branch conflicted in `docs/user_guide/defining_models.md`, where
this branch had replaced the inline `JointTransition` block with a pointer to
`reference/transitions.md#joint-transitions` while #409 had grown that block.

The resolution keeps this branch's pointer, but the pointer's target did not yet cover
four things #409 had added, so they were moved into the reference page rather than
dropped: what `support`, `probabilities` and output laws may read, together with the
params-bound preflight and the guarantee that probability vectors are never silently
normalized; the `params[source][target][kernel][...]` parameter paths; and outputs onto
a stochastic process, including the node-basis reading and the `NaN` a value outside the
support yields.


timmens (commented):
## Blocking

### Solver selection and specialized authoring

#### Add a plain-EGM eligibility gate to the solver-selection guidance

The solver-selection map currently treats the absence of a discrete choice as
sufficient for selecting plain EGM. PyLCM's `EGM` class has additional restrictions:
it requires exactly one continuous state and action, permits no passive or process
states, and requires resources to equal the liquid state. A smooth consumption-saving
model with stochastic income or a genuine resources function therefore follows the
current flowchart to `EGM` and is then rejected, while `DCEGM` supports that model even
when the discrete-action set has only one element.

Could we add an explicit plain-EGM eligibility gate and update the DCEGM descriptions
consistently in Getting Started, the User Guide, Methods, and Reference? It would also
help to explain that PyLCM's class named `EGM` implements a deliberately narrow subset
of the models commonly associated with classic EGM.

## Medium

### Solver selection and specialized authoring

#### Add a small, complete NBEGM authoring example

The specialized-solver documentation contains complete examples for plain EGM and
NEGM, while NBEGM is covered through isolated reference snippets and the much larger
Mahler and Yum NNBEGM model. A reader therefore cannot see the complete minimal
specification in one place.

Could we add an executable NBEGM example that assembles the boundary, both
`@lcm.piece` formulas, output or resources function, `LiquidMargin`, lower-bound
constraint, and solver configuration into one model? This would make the relationship
between the individual declarations and the final solver contract visible.

#### Make the NEGM example exercise its outer adjustment margin

The nested authoring example chooses `illiquid_investment = 0` at every one of the 160
liquid-wealth and illiquid-wealth grid cells. Investment reduces liquid resources
one-for-one, while the illiquid stock has only a small utility weight, so the outer
adjustment branch is never economically relevant. The existing test confirms that the
model solves and simulates without asserting any action values.

Could we calibrate the example so that some states choose positive investment and
others choose no adjustment, then add concrete assertions for both outcomes? That
would make the example demonstrate the outer comparison and the `no_adjustment`
alternative it is intended to teach.

#### Correct nonlinear certainty-equivalent support

The Epstein-Zin notebook states that `certainty_equivalent` is supported only with
`GridSearch`. The current capability contract also supports a nonlinear certainty
equivalent with `NBEGM` and with `NNBEGM` through its inner solver. `EGM`, `DCEGM`, and
`NEGM` reject it.

Could we correct the notebook and state this capability in the central solver table as
well? The current statement can lead users to select `GridSearch` unnecessarily for
models that the endogenous-grid solvers support.

### NBEGM public contracts

#### Update NBEGM parameter documentation to match the current algorithms

Two public NBEGM parameter descriptions refer to removed implementation details.

- `jump_read="bridged"` says that stochastic rows remain available before expectation
  folding. The current solver interpolates all stochastic rows before taking
  expectations, so that pre-fold representation no longer exists.
- `envelope_arithmetic="certified"` describes double-double arithmetic and abstention
  with `NaN` for ambiguous comparisons. The current certified implementation uses
  exact-affine arithmetic with deterministic total ordering.

Could we update these public docstrings and the related internal comments to describe
the current algorithms? The present wording gives users an inaccurate model of the
settings when interpreting or diagnosing jump and envelope behavior.

## Low

### NBEGM public contracts

#### Remove the nonexistent `arithmetic` argument from `_degenerate_inversion`

The `_degenerate_inversion(marginal, consumption)` docstring documents an `arithmetic`
argument that is absent from the function signature. Could we remove that copied
argument description?

## Non-blocking general comment

### Keep solver capabilities consistent across documentation layers

The new documentation structure works very well overall. The separation into Getting
Started, User Guide, Methods, Examples, Reference, and Development gives each page a
clearer purpose, while the executable examples and public-API coverage make the
documentation more trustworthy.

One systemic risk is that solver capabilities and eligibility rules are stated
independently in Getting Started, the User Guide, Methods, Reference, and some
notebooks. The plain-EGM eligibility guidance and nonlinear certainty-equivalent
support already differ between those descriptions. Could we treat the solver
capability table as the canonical source and, in a future PR, derive or validate the
higher-level decision guidance against the same capability declarations?

timmens (approved):
Approved. All follow-up comments below are non-blocking.

## Medium

### Finite outer search describes the wrong quantity as grid-snapped

`docs/reference/outer_search.md:18`

`FiniteOuterGrid.grid` contains outer post-decision targets. Simulation recovers the
declared outer action by exactly inverting the post-decision map. For the supported map
`target = old + 2 * action`, selecting target `1` when `old = 0` yields action `0.5`,
so the recovered action can lie between action-grid nodes. Could this say that the
selected post-decision target is grid-snapped and that the action is recovered exactly?

### Describe `refined_grid_factor` as envelope-row headroom

`docs/reference/solvers.md:107`

`refined_grid_factor` sizes the NaN-padded envelope rows. It provides storage headroom
for ownership changes, and a row that needs more slots is reported as overflow and
NaN-poisoned. The density of the policy read-out grid stays unchanged. Could this
describe the field as envelope-row headroom so users tune the correct numerical
resource?

### Qualify the introductory Bellman equation

`docs/methods/dynamic_programming.md:7`

The displayed `u + beta E[V]` recursion describes the default `LinearAggregator()` plus
`LinearExpectation()` specification. pylcm also supports the general
`H(u, CE(V'))` recursion through, for example, `CESAggregator()` and `PowerMean()`.
Could this either qualify the equation as the default linear case or display the general
aggregator/certainty-equivalent form used on the preferences page?

### Clarify joint-probability validation and normalization

`docs/reference/transitions.md:103`

The statement that joint-transition probabilities are never silently normalized does
not hold when validation is disabled. With probabilities `[0.2, 0.3]`,
`solve(log_level="off")` completes and the default certainty equivalent returns the
mass-normalized mean. In a two-node witness with values `[0, 3]`, the result is `1.8`:
`(0.2 * 0 + 0.3 * 3) / (0.2 + 0.3)`. Could this explain that `debug` rejects invalid
mass, `warning` and `progress` warn and continue, and `off` skips the check? Any path
that continues into aggregation normalizes the mass it receives.

### State the deterministic factory contract for age specialization

`docs/user_guide/age_specialized.md:39`

The list of user obligations is missing that `build(age)` must be deterministic and
side-effect-free. Model construction resolves the same age multiple times. A minimal
counter-based factory was called four times for age 20, and the installed utility
captured the fourth result. A stateful factory can therefore validate one object and
install another. Could this third correctness obligation be stated here and in the
linked Reference section?

### Document all persistence artifacts and the consuming save contract

`docs/user_guide/solving_and_simulating.md:350`

`SimulationResult.save()` writes four sibling artifacts. `arrays/` contains the
per-subject raw-result tree, while the solution arrays are stored separately in
`V_arr/`; the current three-item list omits `V_arr/` and assigns its sharded-value
behavior to `arrays/`. `save()` also consumes the in-memory result by clearing its value
function arrays and compiled regimes, so callers needing further access must reload.
Could this section match the four artifacts and state that consuming behavior?

### Validate the rendered target of public API links

`tests/test_docs_public_api_coverage.py:51`

This check strips each fragment and verifies only that the destination file exists, so
it misses MyST resolving a duplicate implicit heading ID to a different page. In the
current built site, `lcm.Model` resolves to Tiny Example, the continuous-grid links to
the User Guide grids page, the state-transition links to the User Guide transitions
page, and `NEGM`/`NNBEGM` to the Nested EGM methods page. Could the canonical pages use
unique explicit labels and could this guard assert the rendered target source as well
as file existence?

## Low

### Show the keyword-only standalone persistence calls

`docs/reference/runtime_and_results.md:67`

Both standalone persistence functions are keyword-only, so the displayed
`save_solution(solution, path)` and `load_solution(path)` spellings raise `TypeError`
when copied. Could this show the callable forms
`save_solution(period_to_regime_to_V_arr=..., path=...)` and
`load_solution(path=...)`?

### Include qualified constructor calls in the documentation sweep

`tests/test_docs_constructor_kwargs.py:47`

The AST walk accepts only `ast.Name`, so it checks `Model(bogus=...)` while silently
skipping `lcm.Model(bogus=...)`. Several prominent reference examples use qualified
constructors, including `lcm.Model`, `lcm.AgeGrid`, `lcm.LiquidMargin`, and
`lcm.OuterContinuousMargin`, which leaves their keyword arguments outside this guard.
Could the collector resolve supported qualified names so the claimed documentation-wide
contract covers those examples?

### Remove control characters from the joint-transition docstring

`src/lcm/regime.py:173`

There are two embedded DC4 (`U+0014`) control characters around “including carried-only
simulation states.” These can render unpredictably in generated API documentation and
source viewers. Could they be replaced with ordinary punctuation?

PR #429, opened 2026-08-28
Fix taste shocks in simulate AOT compilation

timmens:
## Summary

- supply the per-subject taste-shock PRNG-key input while lowering simulation argmax functions for ahead-of-time compilation
- add a public-API regression test asserting identical seeded discrete choices under lazy and AOT simulation
- let the taste-shock toy-model fixture opt into `Model.n_subjects`

## Problem

A model with both `n_subjects` and regime-level taste shocks failed before simulation while lowering the AOT argmax executable:

```text
ValueError: Expected arguments: [..., 'taste_shock_key', ...], missing: {'taste_shock_key'}
```

Runtime simulation already supplied `taste_shock_key`. The AOT argument template omitted the corresponding shaped key.

## Verification

- `pixi run -e tests-cpu pytest tests/simulation/test_simulate_taste_shocks.py tests/simulation/test_simulate_aot.py tests/test_taste_shocks.py tests/test_random.py -q` (35 passed)
- commit-time hooks, including Ruff formatting, Ruff checks, and `ty` (passed)
- repository policy suite reached the long-running solver matrix with no failures before the focused verification run was concluded; CI will run the full matrix

### DIFF
--- a/src/_lcm/simulation/compile.py
+++ b/src/_lcm/simulation/compile.py
@@ -345,10 +345,19 @@ def _build_argmax_args(
     subject_states = _subject_shape_arrays(
         base.states, n_subjects=n_subjects, sharding=subject_sharding
     )
+    taste_shock_kwargs = {}
+    if regime.has_taste_shocks:
+        _, taste_shock_keys = generate_simulation_keys(
+            key=jax.random.key(0),
+            names=["taste_shock"],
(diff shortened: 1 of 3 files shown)


hmgaudecker (approved):
Good catch, thanks!

PR #430, opened 2026-08-31
Make the keyword-only convention executable

### LINKED ISSUE #421: Make the keyword-only convention explicit, then sweep the functions that predate it
The keyword-only rule from #373 — "as soon as a function takes more than one argument, it must be keyword-only" — is **not written down anywhere**: not in `AGENTS.md`, not in `CLAUDE.md`, not under `.ai-instructions/`. It exists only as a review comment, yet it is deeply followed: **76% of `src/` complies** (618 functions fully keyword-only vs 199 with 2+ positional).

Surfaced by a whole-branch review of #390. The scope was settled in a grilling session; this issue records the decision and the measurements so it can be executed without re-litigating it.

## The rule, as decided

> A function takes its arguments keyword-only as soon as there are two.

It binds **everywhere** — engine code and model functions alike. There are exactly **two** exemptions:

1. **A callback whose caller is a library that passes positionally.** `lax.scan` bodies `(carry, x)`, `custom_jvp` rules `(primals, tangents)`, pytree `unflatten(aux, children)`. Not a judgement call — Python raises otherwise. Verified: `lax.scan` with a keyword-only body raises `TypeError`.
2. **A module that implements an arithmetic and nothing else.** `_lcm/egm/upper_envelope/double_double.py` — operator surrogates (`two_sum`, `two_prod`, `dd_add`, `dd_mul`, `dd_quotient`, …) keep the literature's spelling. Declared once in the module docstring, and such a module may contain **only** operators, so it stays auditable rather than sprinkled.

Two carve-outs that were **considered and rejected**, so they do not get re-proposed:

- *"No hand-written call site"* (library callbacks and DAG functions are never called by a human). Rejected: readability matters just as much deep in the codebase, and the criterion smuggles in "internal code counts less". It also turned out to be factually unforced — **`dags` accepts keyword-only functions** (verified), so DAG model functions were never blocked from complying.
- *"Public API only"*. Rejected: every site in question is under `_lcm/`, so it would exempt the entire population and make the rule vacuous.

## Scope of the sweep

**1101 functions across 236 files**, after removing both exemptions:

| Location | Functions |
| --- | ---: |
| `src/` | 184 |
| `tests/` | 917 |

`tests/` breaks down as:

| Kind | Count | Note |
| --- | ---: | --- |
| ordinary test helpers | 615 | straightforward |
| model functions in `tests/test_models/` | 160 | consumed by `dags` |
| **pytest test functions** | **142** | **needs care — see below** |

### The 142 pytest functions need individual attention

These take fixtures and `@pytest.mark.parametrize` arguments. pytest's binding for fixtures and for parametrized arguments is not uniform, so do **not** apply a blanket `*,` here — check that each still collects and runs. If pytest binding turns out to forbid it, that is a third exemption of the same kind as (1): the caller is a library that passes positionally.

## Why this is not on #390

#390 is a 313-file PR under an ongoing Pro computational-method audit, with #400 stacked directly on it and further PRs behind that. A 1101-function sweep there would:

- move the audit's candidate-set structural digest, forcing the round-7 bundle to be rebuilt;
- have to be replayed through every descendant PR in the cascade;
- bury the review signal on the actual DC-EGM work.

**#390 carries only** the written rule plus the three sites the review flagged:

- `_lcm/solution/diagnostics.py::_states_for_period(regime, state_action_space, period)`
- `_lcm/egm/upper_envelope/segment_envelope.py::_line_value(low, high, x_query, endog_grid, ordinate)`
- `_lcm/simulation/simulate.py::grid_position(name, values)`

## How to execute

1. Wait until #390 and its stack have landed.
2. Branch from `main`.
3. One mechanical commit, no behaviour change. Convert signature *and* every call site together.
4. Skip anything matching exemption (1) or living in a module declared under (2).
5. Handle the 142 pytest functions in a separate commit, verified by collection.
6. `prek run --all-files` and the full suite must be green; **no test may change its assertions** — if one does, the sweep changed behaviour and is wrong.

A `ruff` rule would be better than a one-off sweep if one exists that expresses this; worth checking before hand-rolling. Otherwise consider a small AST check in the pre-commit config so the rule stays enforced instead of drifting again.

hmgaudecker:
Closes #421.

### What problem do you want to solve?

The repository convention that functions become keyword-only once they have two named arguments was documented but not mechanically enforced, and 1,101 pre-existing functions still predated it.

This PR completes the repository-wide migration and keeps it closed:

- adds an AST-based pre-commit/CI checker with actionable diagnostics;
- audits exact positional-callback exemptions and the declared double-double arithmetic module;
- converts production, examples, docs, notebooks, benchmarks, test helpers, and pytest functions together with their call sites;
- preserves mixed-signature coverage for the argument adapters;
- re-anchors the candidate certificate to the mechanically changed source bytes.

The work is split into the convention/checker, the production sweep, and the separate test sweep requested by the issue.

### Validation

- all pre-push hooks passed, including Ruff, formatting, ty, the keyword-only checker, notebook structure, docs formatting, workflow validation, and certificate-related static hooks;
- pytest collection succeeded with 7,187 selected tests and no collection errors;
- focused production, callback, adapter, checker, and documentation tests passed;
- the documentation build executed all 70 pages successfully;
- candidate source inventory regeneration/check and the unified semantic verifier passed.

The full test matrix and long mutation sweep are intentionally left to this PR's CI runs rather than duplicated locally.

### Todo

- [x] Target `main` and use an issue-specific branch.
- [x] Add `Closes #421`.
- [x] Convert production and pytest code in separate commits.
- [x] Let the full CI matrix complete.


.pre-commit-config.yaml
```diff
@@ -150,6 +150,12 @@ repos:
         language: script
         types:
           - jupyter
+      - id: keyword-only-convention
+        name: functions with two named arguments are keyword-only
+        entry: pixi run -e type-checking python -m tests.ci.keyword_only_convention
+        language: system
+        types:
```
timmens: [P2, non-blocking] This hook only receives files classified as Python, so the Markdown fences and notebook cells migrated by this PR remain outside the enforcement boundary. A focused implementation can dispatch by suffix: parse `.py` directly, load `.ipynb` with the standard-library JSON parser and inspect code cells, extract `python` and `py` fenced blocks from `.md`, then use `types_or` for `python`, `jupyter`, and `markdown`. Preserve diagnostics such as `page.md:137` and `page.ipynb:cell 8:line 12`. The current corpus is compatible with this approach: all 113 notebook code cells and 144 Python Markdown fences parse successfully.

tests/ci/keyword_only_convention.py
```diff
@@ -0,0 +1,311 @@
...
+    attached_marker_lines = {
+        marker_line
+        for definition in definitions
+        for _exemption, marker_line in [
+            _exemption_for(definition=definition, source_lines=source_lines)
+        ]
+        if marker_line is not None
+    }
+    violations = []
+    for line_number, line in enumerate(source_lines, start=1):
```
timmens: [P3, non-blocking] The raw line scan treats marker-looking text inside a module docstring as a real exemption and reports `KWO003`. For example:

```python
"""Documentation can show the exemption syntax.

# keyword-only-exempt: library-callback=jax.lax.scan
"""

def compliant(*, left: int, right: int) -> int:
    return left + right
```

Using `tokenize.generate_tokens` to collect standalone `COMMENT` tokens, then sharing that mapping between exemption attachment and orphan detection, avoids the false positive. A regression test with this module docstring would pin the behavior. I verified that this token-based change keeps the nine checker tests green and makes the reproducer pass.

AGENTS.md
```diff
@@ -954,12 +954,19 @@ There are exactly two exemptions:
...
+   ```
+
 1. **A module that implements an arithmetic and nothing else** — operator surrogates
    keep the spelling their operation is known by (`two_sum(a, b)`,
    `dd_mul(left, right)`). State it once in the module docstring; such a module may
    contain *only* operators, so the exemption stays auditable rather than sprinkled per
-   function.
+   function. The declaration is `Keyword-only exemption: arithmetic-only module.`
 
 Functions predating this rule are being converted separately; see issue #421.
```
timmens: [P3, non-blocking] Please remove this sentence when the migration lands. This PR completes the conversion described by issue #421, so the note will immediately be stale.

src/lcm/persistence.py
```diff
@@ -100,9 +100,7 @@ class SimulateSnapshot:
 
 
 def load_snapshot(
-    path: Path,
-    *,
-    exclude: Sequence[str] = (),
+    *, path: Path, exclude: Sequence[str] = ()
```
timmens: [mini, non-blocking] Consider retaining `load_snapshot(path, *, exclude=())`. Its ordinary user-facing call supplies one primary path, and keeping that argument positional is convenient and consistent with the existing examples.

src/lcm/ages.py
```diff
@@ -193,8 +194,11 @@ def get_periods_where(
 
         """
         _convert: Callable[[object], int | float] = int if self._is_integer else float  # ty: ignore[invalid-assignment]
+        # Active predicates are name-bound DAG functions; the unary age-grid
+        # boundary therefore adapts either positional or keyword-only signatures.
+        positional_predicate = allow_args(predicate)
```
timmens, line 199: [mini, non-blocking] `AgeGrid.get_periods_where(np.isfinite)` previously worked and now raises while `allow_args` inspects the NumPy ufunc. Consider falling back to the original callable when signature inspection raises `TypeError` or `ValueError`, with a regression assertion that this example returns `(0, 1, 2)`.

timmens (approved):
Approved. The internal callback and numerical rewiring is sound, and the test-certificate sweep preserves its behavioral checks. One small documentation follow-up: if the strict public signatures remain, please update `docs/reference/case_pieces.md` to use `case_boundary(condition=..., kind="jump")`, and update the `load_snapshot` examples in `docs/user_guide/debugging.md`, `docs/reference/runtime_and_results.md`, and the generated `REPRODUCE.md` in `src/_lcm/persistence/io.py` to pass `path=...`. All remarks in this review are non-blocking.

.pre-commit-config.yaml
hmgaudecker: Implemented in 22a74c1f. The hook now uses types_or for Python, Jupyter, and Markdown; dispatches by suffix; checks notebook code cells and python/py Markdown fences; and reports notebook cell/line or document line coordinates. Indented CommonMark fences are covered by regression tests.

tests/ci/keyword_only_convention.py
hmgaudecker: Implemented in 22a74c1f. Exemptions are now collected only from standalone tokenize COMMENT tokens, and the same mapping drives attachment and orphan detection. The module-docstring reproducer is covered by a regression test.

AGENTS.md
hmgaudecker: Removed in 22a74c1f.

src/lcm/persistence.py
hmgaudecker: Implemented in 22a74c1f. load_snapshot retains a positional primary path and keyword-only options. A narrow audited primary-argument marker keeps the executable convention explicit, and the persistence test now exercises the positional call.

src/lcm/ages.py
hmgaudecker: Implemented in 22a74c1f. allow_args returns the original callable when signature inspection raises TypeError or ValueError, and it no longer forwards defaulted parameters eagerly. The AgeGrid regression confirms np.isfinite returns periods (0, 1, 2) on this NumPy/Python build.

PR #431, opened 2026-08-31
Establish NB-EGM correctness foundations

hmgaudecker:
## Summary

- add an explicit NBEGM.probe_schedule contract with the safe every_solve default
- support first_solve for parameter-invariant model declarations and never for externally validated models
- keep parameter-dependent preconditions solver-owned while the engine supplies every draw
- test scheduling, invalid later draws, skipped checks, and first-solve failures

## Stack

- depends on #430 (codex/issue-421-keyword-only)
- this branch is one commit above that base
- follow-up: nbegm-kernel-replay-diagnostics

## Validation

- pre-push hooks pass
- focused stacked validation: 26 tests passed across probe scheduling, kernel attribution, and period replay


src/_lcm/solution/preconditions.py
```diff
@@ -8,10 +8,13 @@
 run in `Solver.validate`, which the engine calls while building kernels at
 `Model` construction.
 
-Such a solver publishes its checks as `SolutionKernels.param_checks`. `Model`
-calls `check_solver_params` for every solve because the evaluated functions can
-change with every parameter draw; a precondition established for one draw says
-nothing about another.
+Such a solver publishes its checks as `SolutionKernels.param_checks`, and `Model`
```
timmens, line 11: [P3] Please update the companion `ParamCheck` contract documentation to match this paragraph. `src/_lcm/engine.py` still says that the engine calls each check on the first solve and never again, and `src/_lcm/solution/contract.py` makes the same first-solve promise. Four NB-EGM docstrings also describe their probes as first-solve-only. The implementation calls every published check for every parameter draw, while each probe owns its schedule. Keeping that invariant consistent across these locations will prevent future solver implementations from assuming that the engine provides the cache.

src/_lcm/solution/nbegm.py
```diff
@@ -257,16 +257,33 @@ class NBEGM(OneMarginSolver):
...
+    parameter values and synthesizing the states and actions they sweep. A DAG
+    that cannot be differentiated that way leaves the precondition unverified:
 
     - `"reject"` — refuse to solve; the per-interval EGM preconditions must be
       machine-verified.
     - `"assume_declared"` — warn and solve; the model author asserts the budget's
       within-interval affinity and every liquid-reading law's interval-constancy,
       to be validated empirically (e.g. full-model brute-agreement gates).
     """
+    probe_schedule: Literal["first_solve", "every_solve", "never"] = "every_solve"
```
timmens, line 271: [P3] `probe_schedule` also controls `_fail_if_flow_not_single_power` on the Epstein-Zin ride-along path, while this field documentation and the new test module enumerate only budget affinity and interval constancy and describe `"never"` as asserting both preconditions. Please name all three scheduled probes, clarify that `"first_solve"` and `"never"` also relax the single-power-flow check, and add focused coverage for that path. A parameterized flow such as `q(c) = c + B` can pass at the first draw with `B = 0` and leave the supported single-power class at a later draw. The safe default rejects that change; the two opt-out schedules intentionally admit it, so the public contract should make the full scope of the assertion explicit.

timmens (approved):
Good PR. The solver-owned scheduling boundary is clear, the safe default preserves per-draw validation, and the focused tests cover the intended behavior. I left two non-blocking comments to make the opt-out contract complete.

src/_lcm/solution/preconditions.py
hmgaudecker: Fixed in 83b60c3f. The ParamCheck contracts in the engine, SolutionPhase, and SolutionKernels now state that every published check receives every parameter draw, while the check owns its evaluation schedule and cached verdict. I also aligned the four stale NB-EGM docstrings and the companion test documentation.

src/_lcm/solution/nbegm.py
hmgaudecker: Fixed in 83b60c3f. The public probe_failure/probe_schedule documentation now names budget affinity, interval constancy, and single-power flow, and explicitly says that first_solve and never relax all three checks. Added an Epstein-Zin ride-along regression with q(c) = c + B: B = 0 passes initially, a later nonzero B is rejected by every_solve, and is intentionally admitted by first_solve and never. The focused module passes 10/10 tests.

PR #432, opened 2026-08-31
Add NB-EGM kernel replay diagnostics

hmgaudecker:
## Summary

- attribute executed kernels to regime, age, period, and discrete branch cardinalities when LCM_LOG_KERNEL_ATTRIBUTION is enabled
- capture one selected regime-period's actual kernel inputs through LCM_CAPTURE_PERIOD and replay that kernel independently
- expose replay_period through the persistence API
- batch NB-EGM derivative probes so diagnostic preconditions do not compile once per probe point
- test attribution, capture selection, exact replay, gated-edge continuation inputs, and invalid targets

## Stack

- depends on #431 (nbegm-correctness-foundations)
- this branch is one commit above that base

## Validation

- pre-push hooks pass
- 26 focused tests pass across probe scheduling, kernel attribution, and period replay


src/_lcm/solution/backward_induction.py
```diff
@@ -480,6 +485,26 @@ def _run_period_kernel(
     """
     period_kernel = regime.solution.period_kernels[period]
 
+    # Captured before the period-specific state axes are substituted below, so a
+    # replay re-enters this function at the top and repeats every step of it.
+    capture_kernel_inputs(
```
timmens, line 498: `replay_period` re-enters `_run_period_kernel`, so this capture hook runs again when the shell still has `LCM_CAPTURE_PERIOD` set. A writable capture is silently replaced. A read-only capture, such as a downloaded CI or cluster artifact, cannot be replayed:

```python
monkeypatch.setenv("LCM_CAPTURE_PERIOD", "working_life@1")
model.solve(params=params, log_level="off")

capture = tmp_path / "working_life@1"
capture.chmod(0o555)
replay_period(directory=capture)
```

The final call raises `PermissionError` while `_save_pkl` creates its temporary file. The same read-only artifact replays successfully after unsetting `LCM_CAPTURE_PERIOD`. Please resolve the selector once at solve entry and pass explicit capture control into the kernel funnel, with replay disabling capture. This also lets a malformed selector fail before the full AOT compilation.

timmens (changes requested):
The replay lifecycle has one blocking issue: replay can recapture and mutate or fail on its own input while the capture selector remains set.

src/_lcm/solution/backward_induction.py
hmgaudecker: Fixed in 444097e7. The selector is now resolved once at solve entry, before any AOT compilation, and the resolved host-side target is passed explicitly through the period-kernel funnel. replay_period() passes capture_target=None, so replay cannot recapture or mutate its input even when LCM_CAPTURE_PERIOD remains set. The control is absent from CoreProgram, compilation keys, kernel_kwargs, and the serialized payload. Added regressions for read-only replay with the selector still set and malformed-selector rejection before compilation; the focused replay plus certificate suite passes 15/15. I also regenerated sources.json and the hand-maintained backward-induction seal.

timmens (approved):

Issue #428, opened 2026-08-25
Return and consume one complete solution object


hmgaudecker:
## Motivation

Value arrays are not always a complete solution. Some solvers publish additional
artifacts that are required to reproduce their solved policy during forward
simulation—for example, NNBEGM's ordered joint candidate bank and exact discrete
candidate identities.

The current API exposes this incompletely:

```python
values, policies = model.solve(params, return_simulation_policy=True)
```

An automatic `simulate()` call can carry those policies internally when it solves
afresh. The documented separate workflow, however, has historically accepted only
`period_to_regime_to_V_arr`; passing saved values without the corresponding replay
artifacts can therefore select actions from a different candidate set. A bounded
explicit-policy argument and fail-closed validation are suitable for the solver PR,
but they leave two parallel objects that callers must keep aligned themselves.

This extends the model/solution separation discussed in #219. That issue’s original
stateless-model and runtime-grid goals were completed by #235; the concrete need here is
new: a public solution object that retains solver-specific replay artifacts.

## Proposed direction

Introduce a public result type, provisionally `SolutionResult`, which is returned by
`Model.solve()` and consumed directly by `Model.simulate()`:

```python
solution = model.solve(params)
result = model.simulate(
    params=params,
    initial_conditions=initial_conditions,
    solution=solution,
)
```

At minimum it should contain:

- value functions;
- solver-published forward-replay artifacts, keyed by period and regime;
- solver diagnostics needed to explain policy construction, approximation, fallback,
  and convergence outcomes;
- enough structural metadata to validate that artifacts match the model and solution
  layout that consumes them;
- an explicit representation of artifacts that were intentionally not requested or
  persisted.

The artifact payload must remain solver-extensible. The generic engine should transport
it opaquely and ask capability questions; it should not branch on concrete solver
classes.

## Required behavior

1. A separate `solve()` then `simulate(solution=...)` workflow reproduces the automatic
   solve-and-simulate workflow, including solver-specific candidate identities and tie
   rules, and exposes the same solver diagnostics as automatic simulation.
2. Simulation fails clearly before the forward loop if a supplied solution lacks an
   artifact required by one of its solver routes. It must not silently substitute a
   foreign action-grid optimization.
3. Value-only access remains straightforward for comparisons, diagnostics, and warm
   starts. Callers should not need to deserialize large replay banks merely to read
   values.
4. Complete solutions can be persisted and restored without losing categorical
[...]

Issue #422, opened 2026-08-11
Make the `Solver` contract usable from outside the package


hmgaudecker:
`TwoAssetEGM` is being lifted out of pylcm and supplied back as an external specialized
solver, hosted with the paper that publishes its results (copied out at
`ea32bc78`, with a manifest). That move turns a question we have been able to avoid into
one we have to answer: **is `Solver` an extension point for code outside this package, or
only for code inside it?**

Today it is the second. `docs/explanations/adding_a_solver.md` documents the contract and
the design rule is right — the backward-induction loop asks capability questions, never
`isinstance`, and adding a solver must not add a branch to `backward_induction.py`. But
an out-of-tree solver cannot actually be written against the public surface.

Three gaps, in increasing order of difficulty.

### Gap 1 — `KernelResult` is not public

An external solver must *construct* a `KernelResult` (`_lcm/solution/contract.py`), since
it is the sole boundary between solver-specific execution and the generic loop.
`lcm.solvers` exports `Solver`, `SolverBuildContext` and `SolutionKernels`, but not
`KernelResult`, so the only way to build one is `from _lcm.solution.contract import
KernelResult` — reaching into the private package.

`PeriodKernel` needs no equivalent fix: it is a `Protocol`, so plain closures and small
frozen dataclasses satisfy it structurally without importing anything.

Pure promotion. No design content.

### Gap 2 — private helpers a real solver leans on

`TwoAssetEGM` was the honest test case, and it reached for:

- `_lcm.solution.continuation_target`: `target_period_grid`, `_period_to_continuation_target`,
  `_union_fixed_params`, `_union_free_params` — the last three underscore-private inside
  an already-private package
- `_lcm.engine.StateActionSpace`
- `_lcm.grids.ContinuousGrid`
- `_lcm.beartype_conf.REGIME_CONF`
- `_lcm.typing`: `EconFunction`, `EconFunctionsMapping`, `FlatParams`, `RegimeName`

Each needs a public home, a public equivalent, or a documented decision that external
solvers do without it. This is the bulk of the work and the part where "what does a
solver legitimately need at build time?" actually gets decided — `SolverBuildContext`
already exists to bundle exactly that, so some of these may belong on it rather than in a
new public module.

### Gap 3 — the continuation aliases name concrete EGM types

`contract.py`:

```python
type ContinuationPayload = EGMCarry
type SimulationPolicy = EGMSimPolicy
```

[...]

PR #433, opened 2026-08-31
Converge GridSearch execution on one CoreProgram graph and solve() on one SolutionResult

hmgaudecker:
## Summary

This PR establishes the production execution-planning slice around GridSearch and
converges the two interfaces the maintainer review asked for:

- **one native `CoreProgram` graph per GridSearch period kernel** is the sole authority
  for each core's function, argument builder, execution requirements and exact value
  reads, output roles, execution disposition, and reason. Eager, JIT, AOT, liveness,
  layout, and period replay materialize and resolve that same graph through one seam;
  only the final compilation step differs. Unmigrated kernels (EGM, DC-EGM, NEGM, NB-EGM,
  NNBEGM) cross exactly one centralized fail-closed legacy adapter. GridSearch's parallel
  `core` / `unwrapped_core` / `streamed_core` callables, the `enable_jit=False` dense
  bypass, and the `CoreProgramAware` / `_TargetValueAccessAware` / `OutputLayoutAware`
  protocols are removed;
- **`Model.solve()` returns `SolutionResult`** and is the sole public solve entry point.
  `solve_result()`, mapping/tuple returns, `return_simulation_policy`,
  `return_dissolution_flags`, and the split simulation inputs are removed without a
  compatibility bridge; `simulate(solution=...)` consumes the complete result under the
  model-authoritative replay boundary;
- the **Round-2 Pro findings F4 and F5** are implemented as explicit production
  dispositions: singleton EV1 selects `deliberately_dense:ev1_canonical_reduction_order`
  (blockwise grouping changes the canonical float reduction and can flip an upstream
  winner) and collective hard max selects
  `deliberately_dense:collective_resource_regression` (the paired row regressed every
  measured resource surface). Their blockwise implementations stay as private references;
- streamed ordinary singleton hard max keeps its eligible value-dependent,
  fixed-distributed-state co-map, and folded compositions, complete candidate support, and
  the existing feasibility, NaN, tie, and signed-zero semantics;
- model-authoritative replay validation (R1), M0 transfer/liveness planning, and P2
  source-coordinate authentication from the earlier checkpoints are unchanged;
- the diagnostics fold no longer assumes every regime's value array shares one device
  placement (a planned single-device layout is committed; a mesh-sharded neighbour is not
  combinable with it without an explicit scalar transfer).
- a gated edge into a **stateless** target is now gated in solve: the source's
  continuation applies the gate and the leg fallbacks to the target's folded channel
  stack, so a closed gate pays the projected fallback. The dense reduction used to reduce
  silently over the channel axis and always paid the target's own value; the streamed
  program refused the non-scalar cell, which is how CI exposed it.

The transition ledger records the centralized adapter, the two deliberate-dense gates,
and every other bridge with its exact retirement condition.

## Exact anchors

- base: `98be11fbe42f92b32237fea7961379c3a1923dc0`
- Round-2 anchor (immutable, reviewed): `77ebdd72f93c8b8d07697deeb2c8d120fb600442`
- Round-4 bundle head: `0b7b9b98988bf013997cffbe4178e83ef1666dc5` (tree `b169bacb0d797ecc2ac55117ecc39c0538630332`)
- head and Round-5 bundle head: `37e24419` (tree `5e8e0f3e1ef79c458fda81f9e35ab53d00e940da`); differs from the Round-4 bundle head only in `benchmarks/asv/_gpu_mem.py` and `benchmarks/test_gpu_mem.py` (lazy `resource` import for the Windows CI leg)

## Review status

Round 1 closed F1--F3 (result authority). Round 2 audited the immutable Q anchor and
returned F4/F5, which this head implements natively rather than by applying the returned
patch to the superseded dual-interface architecture. Round 4 verified F1--F4 at
`0b7b9b98` and left F5 open for the exact-head paired resource matrix. Round 5 supplied
that matrix at this head (below) with every trigger disposed and is the bounded closure
review before maintainer review.

## Validation at the head

Observed on the exact head on Linux x86_64, CPU backend (`pixi run -e tests-cpu`, Python 3.14.7, jax 0.11.1):

| Check | Result |
| --- | --- |
| 42 focused modules (execution, output layout, liveness, value transfer, streaming production path, period replay, `SolutionResult` authority, distributed, docs coverage, paired harness, solver selection, plus the modules touched since the previous checkpoint), `-n 2` | 606 passed, 21 skipped, 0 failed |
| `tests/test_grid_search_candidate_certificate.py --precision=32` | 96 passed |
| `tests/test_grid_search_candidate_certificate.py --precision=64` | 96 passed |
| `tests/candidate_certificate/verify.py --repo-root . --self-test` | pass; 354/354 mutations rejected; mutation digest `292027e6…`, inventory digest `f5226c9c…` |
| `tests/candidate_certificate/generate_sources.py --repo-root . --check` | inventory matches (51 derived source paths) |
| `prek run --all-files` | 35 hooks passed |

The Windows CI job failed at `0b7b9b98` on a module-level `import resource` in `benchmarks/asv/_gpu_mem.py`; the import is lazy since `37e24419` and the job passes at that head.

The full repository/platform/precision matrix is delegated to CI.

## Deliberate boundaries

No general runtime or peak-memory claim. The width cap of 64 remains a ledgered
bootstrap. The exact-base-to-exact-head paired matrix at this head (Marvin job 27258781,
one A100 80 GB, four CPU devices for the co-map row, float32, four AB/BA repeats; 20/20
pairs bitwise-equal and layout-parity passed, 40/40 workers, zero collectives) measures
the tradeoff this PR makes; median head/base ratios over the four repeats:

| Scenario | warm solve | cold AOT compile | host RSS | compiler peak | device peak |
| --- | --- | --- | --- | --- | --- |
| singleton-hard-max | 1.57 | 1.86 | 1.05 | 0.054 | 0.062 |
| singleton-ev1 (EV1 dense by disposition) | 1.82 | 1.20 | 1.04 | 1.000 | 1.000 |
| collective-gs-vd (collective dense by disposition) | 1.29 | 1.63 | 1.07 | 1.000 | 1.265 (48 KB to 61 KB) |
| distributed-co-map (4 CPU) | 1.38 | 2.24 | 1.07 | 1.003 | n/a |
| folded-hard-max | 1.42 | 2.20 | 1.04 | 0.237 | 0.257 |

The warm and compile costs on every row and the collective device peak are accepted as
the memory-for-runtime trade of the streamed action product; the Round-5 completion
record disposes each trigger by name. One forward item for the tile-policy slice: on the
EV1 row the dense EV1 kernel sets the program's peak, so streaming the model's
retirement kernels bought no peak reduction at a 1.82x warm cost; a plan that cannot
lower the program's peak should not stream. The Round-2 matrix at `77ebdd72` (warm
1.26x--3.05x) is closure history.

### The benchmark-pr comparison is diagnosed; reviewers need not investigate it

The ASV comparison on this PR shows execution-time regressions of 1.2x to 1.7x on every
GridSearch solve row and 3.6x on Mahler-Yum, plus higher peak GPU memory on the ACA
(+29%) and Mahler-Yum (+8%) rows. Compile-time rows are not a review concern.

- The runtime cost is confined to the streamed GridSearch solve. Simulate-only rows are
  flat (Precautionary Savings Simulate 1.02, Iskhakov GridSearch Simulate 0.99) and
  DC-EGM Solve is 1.04.
- Mechanism: `initial_core_tile_widths` streams every non-trivial action product at the
  largest power of two at or below `min(64, extent - 1)`, on any device and regardless
  of whether memory is tight. Mahler-Yum (50 saving points times 3 labor states times
  the effort categories) and Iskhakov (500 consumption points) therefore run each period
  program as a 5 to 8 step scan whose blocks XLA cannot fuse into the dense kernel, with
  a max/argmax merge per block. On the T4 benchmark runner that is 3.6x; on the A100
  matrix above it is 1.3x to 1.8x.
- Streaming only lowers a peak where the dense action product was the peak
  (Precautionary Savings Solve, 34 MB to 4 MB). On the other rows the peak is set by
  arrays streaming does not touch, so those rows pay the runtime and gain no memory.
- The fix is a library change, not a benchmark parameter: a memory-aware width policy
  that selects dense when the product's dense working set fits a declared budget and
  otherwise the largest width that fits, plus an `action_block_width` override on
  `GridSearch`. Every ASV model fits dense on the T4, so the rows return to base
  runtime while production-scale models keep streaming. This is the ledgered
  "bootstrap width cap" item and is deferred to the final PR on this stack.
- The ACA +29% and Mahler-Yum +8% GPU peaks are not explained by streaming (the
  solve-only ACA bridge measured device peak 0.38 head/base at benchmark size). They
  will be attributed with a memory profile split at the solve/simulate boundary in that
  same final PR.

One paired ACA measurement exists for this head and is recorded here as a bounded
observation, not a claim: benchmark-sized ACA model (1 subject, 2 preference types),
base `98be11fb` vs head `0b7b9b98`, harness `e35a7336`, one A40 (non-exclusive node),
executed float64 (every parity array reports `float64`), 2 repeats in AB/BA order,
GridSearch, no NB-EGM.

| scenario | warm head/base | device peak head/base (GB) | cold solve head/base | parity max abs |
| --- | --- | --- | --- | --- |
| aca-a6-c16 | 1.383, 1.346 | 1.296 / 3.414 = 0.380 | 1.66 | 5.7e-13 |
| aca-a6-c64 | 0.984, 0.994 | 1.215 / 13.307 = 0.091 | 1.33 | 5.1e-13 |

The c16 warm penalty reproduces the earlier `d141e968` measurement (1.343) on the
converged head; its mechanism is not measured. Device peak bytes are identical to that
earlier run on both sides. Head-side `hlo_instruction_count` is 1.80x and
`hlo_text_bytes` 1.70x the base. The fp64 bridge rows are not comparable to an fp32
frontier.

One production-sized ACA measurement exists, cold legs only so far: one warm-start MSM
criterion evaluation (full solve, simulate, moments) on the production grid (assets 24,
aime 12, consumption 70, wage residual 5, hcc 3x5, savings 200, `pref_type` distributed
one type per GPU), three A40s on a non-exclusive node, executed float32 read from the
artifact (`jax_enable_x64: false`), base pylcm `98be11fb` vs head `0b7b9b98` with the ACA
arms differing only by dropping the `period_to_regime_to_V_arr=None` keyword.

| metric | base | head | head/base |
| --- | --- | --- | --- |
| criterion evaluation wall | 3668 s | 4155 s | 1.133 |
| device pool peak, GPU 0 / 1 / 2 | 1.31 / 1.28 / 1.28 GB | 0.69 / 0.67 / 0.67 GB | 0.528 / 0.524 / 0.524 |
| criterion value | 76491.45 | 76421.03 | rel. diff 9.2e-04 |
| residuals (624), max abs element-wise difference across arms | | | 5.64 |
| residuals (624), max abs element-wise difference between two head repeats | | | 6.45 |

The production fp32 run cannot test numerical parity: repeat evaluations of the same
revision differ by more than the two revisions differ from each other (criterion 5.1e-03
relative between the cold and warm head legs, against 9.2e-04 between head and base;
residual vectors 6.45 apart within the head arm against 5.64 across arms, with about
163 of 624 residuals bit-identical in both comparisons). Repeat evaluations differ because the
criterion's simulation draws a fresh random seed on every call (pylcm's documented
`simulate(seed=None)` default on both revisions; the ACA call site passes none), so the
row is a statement about the ACA harness, not about this PR. Device pool peak is stable across
the head repeats (0.69/0.66/0.66 GB cold, 0.69/0.66/0.66 GB warm), and the head warm leg
ran 4049 s against 4209 s cold, so the workload is execution-dominated rather than
compile-dominated.
No production runtime or parity claim is made from this row. The runtime cost is a
total: the kernel-attribution log names which regime-periods each compiled kernel served
(438 structural lines, none timed), so attributing the 13% needs a separate profiled
pair. Warm legs are pending.

Still dense: action products of extent at most one, co-mapped states combined with a
separate same-period or edge-value channel, EV1, collective hard max, and all
simulation-policy construction. Unsupported by the streamed program: collective EV1,
EV1 plus a folded state, collective hard max plus a folded state, EV1 without a discrete
action. Liveness records eligibility only; no release, donation, offload, submesh,
concurrency, or scheduler claim.

`CoreProgram` remains private. `SolutionResult` is in-memory only; durable persistence,
cross-process identity, and a stable external solver contract remain later capabilities.

## Next steps

1. Round-5 bounded closure review of this head (F5 matrix) :heavy_check_mark:, then maintainer review, then land.
1. #436 (the #434 simulate fix) is stacked on this branch and merges after it; the
   NB-EGM phase of #435 stacks on #436.
1. Everything after that is laid out in #435: production NB-EGM/NNBEGM execution in a
   fresh PR, then plain EGM / DC-EGM / NEGM / external-solver migrations, then
   persistence and conformance, then broad execution coverage. #435 also lists the
   bridges a reviewer will notice in this PR (legacy adapter, dense EV1/collective
   dispositions, eligibility-only liveness, no persistence) and which phase retires each.

Advances #422 and #428 without closing either.


src/lcm/model.py
```diff
@@ -597,6 +648,98 @@ def solve(
             return internal_result.value_functions, internal_result.dissolution_flags
         return internal_result.value_functions
 
+    @beartype(conf=PARAMS_CONF)
+    def solve_result(
```
timmens: Given that pylcm has no current users and the public API becomes a compatibility commitment only when the package is published, would it make sense to have `solve()` return `SolutionResult` directly and remove the parallel `solve_result()` workflow? This development window allows the return-type migration without a deprecation bridge.

`SolutionResult` is already a keyword-only, extensible container. Later work can add defaulted fields and methods for persistence, durable identity, continuation artifacts, and solver extensions without another return-type migration. Keeping both entry points creates two public workflows, duplicate documentation and adapters, and a planned API switch shortly before publication. One possible convergence is to update internal callers and tests to consume `solution.values` and the addressed artifact stores, then remove the legacy tuple-return flags in the same change. This is a nonblocking design discussion.

src/_lcm/execution/core_program.py
```diff
@@ -0,0 +1,611 @@
...
+    def __post_init__(self) -> None:
+        """Snapshot the declared axes and exact target-value reads."""
+        object.__setattr__(self, "streamable_axes", tuple(self.streamable_axes))
+        object.__setattr__(
+            self, "target_value_accesses", tuple(self.target_value_accesses)
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class CoreProgram:
```
timmens, line 166: `CoreProgram` currently repeats facts already exposed through `PeriodKernel.cores()`, `build_lower_args()`, and `OutputLayoutAware`. GridSearch also carries `core`, `unwrapped_core`, and `streamed_core`, while `enable_jit=False` bypasses the program and selects the dense implementation. The compiler therefore represents one mathematical operation through several synchronized sources.

Consider making each kernel expose `{name: CoreProgram}` as its sole program declaration. A program would own its raw function, argument builder, execution requirements, and output roles. The engine could resolve that same program for eager or AOT execution. One centralized adapter could translate unmigrated legacy kernels into unplanned programs, confining transitional compatibility to a single boundary. The dense GridSearch implementation could then remain as an independent test oracle. This is nonblocking because these interfaces are private and the transition ledger records their retirement.

benchmarks/grid_search_pair.py
```diff
@@ -0,0 +1,895 @@
+"""Run one immutable harness against exact base and head GridSearch checkouts."""
+
+from __future__ import annotations
```
timmens: This project requires Python 3.14 and its repository standards explicitly prohibit `from __future__ import annotations`. The same import is added in `benchmarks/grid_search_pair_worker.py`, `tests/test_grid_search_pair.py`, and `tests/candidate_certificate/direct_flow.py`. Please remove all four occurrences; Python 3.14 already provides the required annotation behavior. This is nonblocking.

timmens (commented):
Three nonblocking comments: an API design discussion, an internal architecture suggestion, and a repository-standards cleanup.

src/lcm/model.py
hmgaudecker: Sure! I think you're just too quick -- this should just be a transient thing during the migration.

benchmarks/grid_search_pair.py
hmgaudecker: Thanks! Done in 784238f9.

src/_lcm/execution/core_program.py
hmgaudecker: Excellent catch, will do!

src/lcm/model.py
hmgaudecker: Done in c4eb421c, on this PR. `Model.solve()` now returns `SolutionResult` and is the only public solve entry point; `solve_result()`, the mapping/tuple returns, both return-shaping flags, and the split simulation inputs are gone, with no deprecation bridge. `simulate(solution=...)` is the sole handoff and validates the result against the model's own descriptors rather than trusting metadata carried beside the arrays. Internal callers, tests, docs, examples, and notebooks consume `solution.values` and the addressed artifact stores. The round-4 and round-5 external audits verified the cutover at the current head (no parallel return path, one entry point), so this is closed rather than transient.

src/_lcm/execution/core_program.py
hmgaudecker: Done in c4eb421c, on this PR, and it is now the architecture. Each GridSearch period kernel publishes exactly one `{name: CoreProgram}` graph through `core_programs()`; the program owns the raw function, argument builder, execution requirements (including its exact target-value reads), output roles, and disposition. `core_program_graph()` is the single seam that reads a kernel, and eager, JIT, AOT, liveness, output layout, and period replay all resolve the same materialized program; a native kernel that still exposes `cores`, `core`, `unwrapped_core`, `streamed_core`, `build_lower_args`, or a parallel output-layout method is rejected at build. `enable_jit=False` no longer bypasses the program: it runs the same resolved function eagerly. Unmigrated kernels (EGM, DC-EGM, NEGM, NB-EGM, NNBEGM) cross one fail-closed legacy adapter that emits explicitly `LEGACY_UNPLANNED` programs, and the dense GridSearch evaluator survives only as an independent oracle in the tests. The fp32/fp64 candidate certificate binds the graph (354 synchronized mutations, including native-graph duplicate-authority and adapter-bypass classes). The remaining kernels migrate onto the same graph in #437, after which the adapter is deleted; the ledger records that retirement condition.

timmens (approved):
Reviewed the follow-up changes at this head. The public solve API and CoreProgram design discussions are fully addressed, the additional correctness fixes are covered by focused tests, and all required checks are green. The remaining cleanup and documented performance trade-offs are suitable for follow-up work.

PR #436, opened 2026-09-02
Read only the edge references of targets active in a period

### LINKED ISSUE #434: Simulation demands a landing value the next period never solved when a regime's gated edges have disjoint target activity windows
## Summary

Simulation raises `KeyError` when a regime has two gated edges whose **targets have
disjoint activity windows**. In any period where *some* edge needs its references, the
runtime asks for the landing value of *every* edge reference regime — including regimes
the next period never solved.

The solve is unaffected; only `simulate` fails.

## Version

`98be11fbe42f92b32237fea7961379c3a1923dc0`, which is current `main` at the time of
filing ("Add NB-EGM kernel replay diagnostics (#432)").

## Minimal reproducer

Four singleton regimes. `source` (ages 0–1) has two `ValueDependentTransition` edges:
to `near` (active at age 1) and to `far` (active from age 2). Their route fallbacks are
`source` and `far_fallback` respectively, and `far_fallback` is only active from age 2.

```python
"""Minimal reproducer: simulation demands a landing value the next period never solved.

`source` has two gated edges whose targets have DISJOINT activity windows.
In period 0 the runtime asks for the landing V of every edge reference regime,
including `far_fallback`, which is not active in period 1.
"""
import jax.numpy as jnp
from lcm import (
    AgeGrid, LinSpacedGrid, Model, ProjectedRegimeValue, Regime, StakeholderRoute,
    ValueDependentTransition, MarkovTransition, categorical,
)
from lcm.typing import BoolND, ContinuousAction, ContinuousState, FloatND, ScalarInt


@categorical(ordered=False)
class RegimeId:
    source: ScalarInt
    near: ScalarInt
    far: ScalarInt
    far_fallback: ScalarInt


W = LinSpacedGrid(start=1.0, stop=10.0, n_points=5)
C = LinSpacedGrid(start=0.5, stop=9.0, n_points=5)

def utility(*, consumption: ContinuousAction) -> FloatND:
    return jnp.log(consumption)

def utility_state(*, wealth: ContinuousState) -> FloatND:
    return jnp.log(wealth)

def next_wealth(*, wealth: ContinuousState, consumption: ContinuousAction) -> ContinuousState:
    return wealth - consumption + 1.0

def feasible(*, wealth: ContinuousState, consumption: ContinuousAction) -> BoolND:
    return consumption <= wealth

def keep_wealth(wealth: ContinuousState) -> ContinuousState:
    return wealth

def always_open(wealth: ContinuousState) -> BoolND:
    return wealth > -1.0

def p_near(age: int) -> FloatND:
    return jnp.where(age < 1, 1.0, 0.0)

def p_far(age: int) -> FloatND:
    return jnp.where(age >= 1, 1.0, 0.0)

def _edge(probability, fallback_regime):
    return ValueDependentTransition(
        probability=MarkovTransition(probability),
        gate=always_open,
        routes={"only": StakeholderRoute(
            target_stakeholder=None,
            fallback=ProjectedRegimeValue(regime=fallback_regime,
                                          projection={"wealth": keep_wealth}),
        )},
    )

common = dict(
    states={"wealth": W}, actions={"consumption": C},
    state_transitions={"wealth": next_wealth},
    functions={"utility": utility}, constraints={"feasible": feasible},
)

source = Regime(
    transition={"near": _edge(p_near, "source"), "far": _edge(p_far, "far_fallback")},
    active=lambda age: age < 2, **common,
)
def near_next(age: int) -> ScalarInt:
    return jnp.full(jnp.shape(age), RegimeId.far, dtype=jnp.int32)


near = Regime(transition=near_next, active=lambda age: age == 1, **common)
far = Regime(transition=None, active=lambda age: age >= 2,
             states={"wealth": W}, functions={"utility": utility_state})
far_fallback = Regime(transition=None, active=lambda age: age >= 2,
                      states={"wealth": W}, functions={"utility": utility_state})

model = Model(
    ages=AgeGrid(start=0, stop=3, step="Y"),
    regimes={"source": source, "near": near, "far": far, "far_fallback": far_fallback},
    regime_id_class=RegimeId,
    description="edge-reference activity repro",
)
print("MODEL BUILT")
print("edges of `source`:", sorted(model.user_regimes["source"].gated_edges))

PARAMS = {r: {"koopmans_aggregator": {"discount_factor": 0.95}} for r in ("source", "near")}

V, flags = model.solve(params=PARAMS,
                       log_level="debug", max_compilation_workers=1,
                       return_dissolution_flags=True)
print("SOLVE OK; regimes solved per period:",
      {p: sorted(r) for p, r in V.items()})
res = model.simulate(
    params=PARAMS, initial_conditions={
        "regime_id": jnp.array([int(model.regime_names_to_ids["source"])] * 2),
        "age": jnp.array([0.0, 0.0]), "wealth": jnp.array([5.0, 6.0]),
    },
    period_to_regime_to_V_arr=V, period_to_regime_to_dissolution_flags=flags,
    log_level="debug", max_compilation_workers=1, seed=0,
)
print("SIMULATE OK (bug not reproduced)")
```

## Observed

The solve succeeds and shows plainly that period 1 never solved `far_fallback`:

```
MODEL BUILT
edges of `source`: ['far', 'near']
SOLVE OK; regimes solved per period: {3: ['far', 'far_fallback'], 2: ['far', 'far_fallback'], 1: ['near', 'source'], 0: ['source']}
...
  File ".../_lcm/simulation/simulate.py", line 653, in _referenced_value_kwargs
    {ref: landing_V[ref] for ref in regime.edge_reference_regimes}
KeyError: 'far_fallback'
```

## Expected

Simulation completes. In period 0 the only edge that carries probability is the one to
`near`; the edge to `far` has probability 0 there, and its fallback's value is not
needed.

## Where it comes from

The two quantities the runtime pairs are computed on different granularities:

- **`edge_reference_periods` is per period** — derived from whether *that period's*
  compiled program takes the edge-ref argument
  (`_lcm/regime_building/processing.py`, `edge_reference_periods=frozenset(...)`).
- **`edge_reference_regimes` is one flat tuple per regime** —
  `_edge_reference_regimes(user_regime)` in `_lcm/regime_building/processing.py` takes
  no period and applies no activity filter; it is the union of the gate references and
  leg fallbacks over *all* of the regime's edges.

So the period guard fires correctly, and then the regime set it guards is the whole
union rather than the subset that period actually needs.

The AOT lowering does the same iteration at `_lcm/simulation/compile.py` but survives it,
because it builds zero templates from `regime_V_topology[ref]`, a topology that exists
for every regime regardless of activity. The runtime at `_lcm/simulation/simulate.py`
indexes the solved values and cannot. The comment there — "The same set the AOT lowering
consults, so a compiled program's pytree and this call's arguments name the same
channels" — holds only when every edge reference regime happens to be active in
`period + 1`.

## Why it cannot be worked around downstream

- Making the fallback regime active earlier is a false statement about the model. In our
  case the source is a working-age single woman (25–59) and the fallback is a retired
  regime (60–94).
- `transition` is a static per-regime mapping, so there is no way to declare an edge
  only in the periods where its target is active.

Any model whose regime has two edges with different target activity windows hits this.

Found while building De Nardi, Fella & Paz-Pardo (2024) in `lcm-zoo`, where a single
woman can marry into either a working-age couple or a couple in which she is already
retired; the two targets are active over disjoint age ranges.

hmgaudecker:
Stacked on #433 (base branch `codex/architecture-performance-plan`); merge after #433.

## Summary

Simulation raised `KeyError` for a regime with two gated edges whose targets are active over disjoint age windows (#434). The per-period guard that asks whether a period's decision program reads the edge-reference channel was paired with a per-regime union of every edge's gate references and leg fallbacks, so a period in which only one edge folds still demanded the other edge's fallback value from a period that never solved it.

The canonical regime's simulation phase now records `edge_reference_regimes_by_period`: for each period whose decision program reads the channel, the reference regimes of exactly the edges folding at `period + 1`. That is the same selection the period's `Q_and_F` is built from, so the program's pytree and the call's arguments name the same channels. The runtime call and the ahead-of-time lowering both read the mapping; the flat per-regime union stays for the solve side, which already filters by reachable targets.

## Test

`tests/simulation/test_gated_edge_disjoint_target_windows.py` is the issue's reproducer on the current API, at both `enable_jit` settings. With the source change reverted it fails with `KeyError: 'far_fallback'` inside `_referenced_value_kwargs`; with the change it passes, and the six existing gated-edge and simulate-router modules stay green (40 tests).

Closes #434.


tests/simulation/test_gated_edge_disjoint_target_windows.py
```diff
@@ -0,0 +1,169 @@
...
+            "near": near,
+            "far": far,
+            "far_fallback": far_fallback,
+        },
+        regime_id_class=RegimeId,
+        enable_jit=enable_jit,
+    )
+
+
+@pytest.mark.parametrize("enable_jit", [True, False])
```
timmens, line 145: [P2] Add coverage for the AOT simulation path

These cases omit the public AOT route: `enable_jit=True` uses lazy JIT, while `Model(..., n_subjects=2)` activates AOT compilation. With this exact regression model and `n_subjects=2`, simulation fails before `_build_argmax_args` can use the new period-specific mapping:

`KeyError: "The gated edge to regime 'near' has no simulate gate evaluator for period 2: 'near' is not active there, so it holds no value to fold. Compiled periods are [1]."`

`_collect_edge_gate_evaluators` requests every source landing period from every edge, so source periods `(0, 1)` request fold periods `(1, 2)` from the `near` edge even though its target is active only in period 1. Please intersect those landing periods with each edge's `simulate_gate_evaluators_by_period`, and add an AOT case here using `n_subjects`. Bypassing only evaluator precollection makes this reproducer complete with both subjects entering `near` in period 1.

timmens (changes requested):
One blocking AOT issue remains. Details inline.

timmens (approved):

tests/simulation/test_gated_edge_disjoint_target_windows.py
hmgaudecker: Done in 05e050b9.

`_collect_edge_gate_evaluators` now intersects the source's fold periods with each edge's `simulate_gate_evaluators_by_period` before lowering, so an edge whose target is active over a narrower window than its source is only lowered for the periods it actually carries an evaluator for. That mirrors what the router already does: `gated_routing.py` skips an edge whose target is not in `same_period_mappings` before it ever calls `simulate_gate_evaluator_at`, so nothing skipped here could have been demanded at simulate time.

One deliberate detail: only a *populated* mapping decides which periods to skip. An empty mapping still reaches the lookup, which reports it as the "never compiled" staging mistake `_select_period_callable` distinguishes — a blanket `continue` would have hidden that behind a silent no-op.

The regression test is now parametrised over `n_subjects in {None, 2}` (ids `lazy`/`aot`) crossed with `enable_jit`. Before the fix the two `aot` cases failed with exactly the KeyError you quoted and the two `lazy` cases passed; after it all four pass, and `tests/simulation` is at 415 passed / 2 skipped / 5 xfailed on CPU.

https://claude.ai/code/session_01Y7QCx9tkGD3TqwNw681QwD

Issue #435, opened 2026-09-02
Roadmap after #433: NB-EGM family execution, solver migrations, persistence, broad execution coverage


hmgaudecker:
This issue records where the execution-planning and solution-result work is going after #433 lands, so that a reviewer of #433 can see which loose ends are deliberate and which phase owns each of them. It condenses the maintainer's planning documents; the authoritative per-bridge list with retirement conditions is `docs/development/architecture_transition_ledger.md`.

## What #433 establishes

- One native `CoreProgram` graph per `GridSearch` period kernel is the sole authority for each core's function, argument builder, requirements, value reads, output roles, execution disposition, and reason. Eager, JIT, AOT, liveness, output layout, and period replay resolve the same graph. Unmigrated kernels (EGM, DC-EGM, NEGM, NB-EGM, NNBEGM) cross one fail-closed legacy adapter.
- `Model.solve()` returns `SolutionResult` and is the only public solve entry point; `simulate(solution=...)` consumes it under model-authoritative replay validation. There is no parallel return path and no deprecation bridge.
- Streamed action reduction is planned per program with an explicit disposition and reason. Singleton EV1 and collective hard max are dense by disposition (`deliberately_dense:ev1_canonical_reduction_order`, `deliberately_dense:collective_resource_regression`); their blockwise implementations stay as private references.
- The fp32/fp64 candidate certificate binds the native graph (seals, AST hashes, 354 synchronized mutations).

#433 makes **no** general runtime or peak-memory claim. Liveness records eligibility only: no physical release, donation, offload, submesh placement, concurrency, or scheduling is implemented. The action-width cap of 64 is a ledgered bootstrap.

## Remaining phases, in order

**Locked delivery shape (2026-09-03).** The phases below ship as three stacked PRs, each with its own Pro delta review and maintainer review:

1. **#437** (stacked on #436): phase 1 in full, then phase 2 (migrations of plain EGM, DC-EGM, NEGM; #422; retirement of the legacy adapter). They share one closure gate, adapter retirement, which fires only once every built-in solver has moved.
2. **Second PR**: phase 3 (persistence, lazy loading, the out-of-tree conformance solver; closes #428). The conformance solver is the acceptance test of the persistence and plugin boundary, so it is written against a boundary the migrations no longer move; it does not start before phase 1's retention specialization has landed.
[...]

PR #437, opened 2026-09-02
Publish every built-in solver as a native core-program graph and retire the legacy adapter

### LINKED ISSUE #422: Make the `Solver` contract usable from outside the package

hmgaudecker:
Closes phase 1 of #435 and delivers #422.

## Summary

Every built-in period kernel publishes a native core-program graph and returns a `KernelOutput`, and the backward-induction loop consumes that contract alone. There is no legacy adapter: `KernelResult`, `require_legacy_kernel_result`, `normalize_kernel_output`, the `UNPLANNED` layout and the post-hoc value-placement repair are gone. A published value lands in its declared placement or the solve fails, and `CoreExecutionDisposition` is exactly `PLANNED` and `DENSE`.

## Execution contract

- **`StateAxesLeading`** is the one parametrized output role. A leaf is placed on the named state axes' template placement, after a declared number of free leading axes, and every later axis is replicated; `dtype` and `shape` are checked when declared. `VALUE` and `DISSOLUTION_FLAG` remain as members. A role tree is any registered pytree with exactly one `VALUE` leaf.
- **`ProgramScope`** (`ANY`, `VALUES_ONLY`, `REPLAY`) is part of a `CoreProgram`'s specialization identity. A solve lowers only the programs its retention consumes, so a values-only solve never compiles a replay program.
- **A continuation is addressed by a versioned `ArtifactKey`** on the runtime-checkable `ContinuationArtifact` protocol, and the engine stores and rolls the payload without reading its fields. `Solver.required_continuation_keys` is a frozenset that model building matches against what every reachable target publishes, refusing the model with both regimes and the demanded version named.
- **Every regime declares one replay route**, `SimulationPhase.replay_route`, carrying a `ReplayMode` (`EXACT_REPLAY`, `VALID_RECOMPUTATION`, `UNSUPPORTED`), the exact payload class the solve retains, and the reader that consumes it. Simulation and the pre-simulation payload check dispatch on the route; a mismatched payload is refused with both classes named.

## Kernels

- **Plain EGM**: one dense program, `main`, reason `egm_one_row_no_product_axis`; carry roles `StateAxesLeading(())`.
- **DC-EGM**: one dense program, reason `dcegm_solver_owned_node_and_grid_batching`; carry and policy roles over the discrete and passive states; one exact `NEXT_REGIME_VALUE` access per stateful target.
- **NEGM**: `keeper` plus `outer_sweep`. The sweep over the durable margin is one compiled `jax.lax.map` program with `outer_batch_size` as its block width, binding the keeper's value and carry at runtime (ledgered, since a `CoreProgram` graph declares no intra-graph dependency).
- **NB-EGM ride-along**: one tile-local core publishing two planned programs, `main` (`VALUES_ONLY`: value and carry) and `replay` (adds the consumption policy and the branch banks). The expected-continuation stacks over the ride cells never exist as a complete array. The continuation is read once per class of branches that agree on every discrete action reaching it. `interval_batch_size` streams the continuation read and the candidate-envelope fold together: the standing winner re-enters each block under its global stored-link index, so every partition decides ownership by the same total order over the same identities, under both envelope arithmetics, and the step publishes each node's owner on request (`return_owner=True`). The certified kernel and both native payloads take an explicit `stable_index` operand for the final tie-break. Standalone NB-EGM (single liquid state) shares the plain EGM kernel.
- **NNBEGM**: a finite and an adaptive kernel selected by `outer_search`, each republishing the inner programs under `keeper:` and `adjuster:` prefixes. The adaptive policy is retained under `VALUES_AND_REPLAY` and omitted as `NOT_PERSISTED` under `ALL_PERSISTABLE_ARTIFACTS`.
- **GridSearch and the terminal-carry wrapper** return `KernelOutput` on the same channels; the wrapper adds the `EGM_CONTINUATION` channel and refuses a base kernel that already publishes it.

## Public surface (#422)

A solver can be written against `lcm.solvers`, `lcm.solver_api`, `lcm.typing` and `lcm.grids` alone. `lcm.solvers.__all__` has 51 names and `lcm.solver_api.__all__` 16, with nothing removed. `docs/reference/custom_solvers.md` states what a solver owes the engine and carries a runnable minimal solver that the test suite executes, with an assertion that it imports nothing from `_lcm`. The surface is importable and tested, not yet supported: persistence, durable cross-process identity, a conformance suite and a versioning policy belong to #428, and the page's Status section says so.

## Numerical guarantees

At default settings solved values are unchanged against the merge base. The block-width knobs (`outer_batch_size`, `interval_batch_size`, `cell_block_size`, `branch_batch_size`) are compiled vmap widths: ownership, feasibility and policy are invariant to them exactly, published levels to within a few units in the last place, and two candidates whose reads tie within that spacing are ordered by the records each width produced.

## Docs

`docs/reference/custom_solvers.md` is new. `docs/reference/public_api.md`, `docs/reference/solvers.md`, `docs/reference/outer_search.md`, `docs/reference/runtime_and_results.md`, `docs/user_guide/solving_and_simulating.md`, `docs/explanations/architecture.md`, `docs/methods/nested_egm.md`, `docs/methods/nonconvex_budgets.md`, `docs/methods/performance_scaling.md` and `docs/development/benchmarking.md` describe the state above. `docs/development/architecture_transition_ledger.md` carries the rows that stay open, including the NEGM sweep's runtime binding and the keyed continuation channel. `CHANGES.md` records the execution contract, the public solver contract and the tile-local NB-EGM execution.


src/_lcm/solution/kernel_output.py
```diff
@@ -56,27 +123,74 @@ def normalize_kernel_output(
...
-        continuation = raw_continuation
-
+        continuation = _pop_typed_artifact(
+            channel="continuations",
+            artifacts=continuations,
+            key=continuation_key,
+            expected_types=(ContinuationArtifact,),
+            regime_name=regime_name,
+            period=period,
+        )
```
timmens, line 133: Optional, non-blocking: consider checking that the returned continuation’s `artifact_key` matches `continuation_key`, so a mismatch reports the regime, period, and conflicting versions directly. JAX template matching already rejects this in the tested case when the key is static metadata; this would improve the diagnostic.

timmens (approved):
Looks good to merge. Reviewed the shared execution and artifact contracts, the built-in solver migrations, and focused numerical/replay checks at 19491b4. I found no blocking PR-introduced numerical defect within that scope. Two non-blocking points remain: the omission metadata below and the inline diagnostic suggestion.

### Non-blocking: use the declared continuation key in omission records

In [artifacts.py:112](https://github.com/OpenSourceEconomics/pylcm/blob/19491b4bf447e68fa9e457ebdace38c4653992f2/src/_lcm/solution/artifacts.py#L112), could this use `regime.solution.continuation_spec.artifact_key`? The custom counting solver introduced here publishes `tests.counter`, but its `SolutionResult.omissions` records `pylcm.egm.continuation`. The existing counting-solver fixture reproduces this under both `VALUES` and `ALL_PERSISTABLE_ARTIFACTS`: looking up `ArtifactRef(period=0, regime="alive", key=_COUNTER)` in the omissions fails. Value functions are correct; the public omission record names the wrong artifact.

src/_lcm/solution/kernel_output.py
hmgaudecker: Both points are now closed on #438 (`60932a76`).

The `artifact_key` cross-check is in `_normalize_kernel_output`, where the producing regime and period are still known; the message names the publication key and the payload's declared key. Your reading of the current behaviour was right — the mismatch is already rejected, but two frames further on, by the continuation template match in `_match_continuation_template_sharding`, as `ValueError: pytree structure error: different types at key path` with no cell coordinates. The new test fails with exactly that before the change and passes with the named `RuntimeError` after it.

The omission record also gained a guard, since it was previously resting on the rewrite's shape with nothing asserting it. `test_no_omission_names_an_artifact_the_model_never_declared` is the discriminating one: reintroducing the fixed-key block on top of the current code makes it fail at all three retentions, while `test_an_unkept_continuation_is_recorded_under_the_key_its_solver_declared` stays green, because the block *adds* a foreign key rather than removing the right one. So the second test is the witness and the first pins the positive half.

PR #438, opened 2026-09-05
ENH Persist complete solutions and add external solver conformance

### LINKED ISSUE #428: Return and consume one complete solution object

hmgaudecker:
## Summary

A `SolutionResult` is the labelled solution boundary: values, replay artifacts, metadata, and explicit omission reasons stay distinct, and a result nobody references any more releases its arrays.

- `SolutionResult.save(path=...)` / `save_solution(...)` atomically persist a complete labelled solution; `load_solution(path=...)` restores independently lazy values and artifacts. The versioned archive holds JSON metadata and checksummed numerical datasets only — no model, plugin class, callable, pickle, or executable code. A restored result saves again without a model: its payloads are re-read from the archive they came from and verified before they are written. `load_legacy_solution(path=...)` remains the explicit reader for old value-only HDF5 files.
- `SolutionMetadata` carries a durable mathematical model fingerprint plus exact solver, replay-route, artifact, solution-schema, and archive-format identities, so a restored result replays in a separately constructed compatible model. The fingerprint hashes every referenced array with its real rank and shape and sees through beartype guards a downstream package's own claw wraps around model functions. Compatibility is exact; incompatible versions fail clearly instead of migrating implicitly.
- The model is sealed at build. `Model(...)` digests its structure once and records every global and closure binding its callables read; rebinding one afterwards makes `solve()` and `simulate()` refuse with `ModelSealError`, and a model whose callables cannot be fingerprinted is refused with `ModelInitializationError`. Solution-relevant parameters are canonicalized once per solve and bound to the model and parameter fingerprints; execution-only and simulation-transition inputs stay out of the digest.
- A result is consumed according to its provenance. One the same model instance solved for the same canonical parameters is read by reference, so a simulate following a solve copies and re-validates nothing; a restored, unpickled, or foreign result is validated in full and materialized once per consuming model and parameter vector, and later simulations reuse that view. The consumed result stays reachable as `SimulationResult.solution` until the simulation result is saved.
- Persistence is selected per artifact authority. The authority a model declares is bound once per instance and support; the authority for solve-generated artifacts lives on the result. The adaptive NNBEGM policy carries its outer nodes as the candidate axis of its descriptor and is model-verifiable, so both NNBEGM policies persist; solver diagnostics persist under a descriptor generated from the payload and read back without a model. A present artifact declared `NOT_PERSISTED` becomes an explicit omission in the restored result. Retention selects computation per exact artifact address through `CoreProgram.retained_artifact_keys`, so kernels never assemble outputs the selected retention discards.
- `ValueStore` and `ArtifactStore` expose per-entry `LoadState`; metadata, omissions, coordinates, and whole-archive checksum verification never materialize payloads. Public mappings are admitted through one item traversal with every raw address checked before insertion.
- Every external solver declares how its decision is replayed: `SolutionKernels.replay_route` is an `ExecutableReplayRoute` with durable plugin and route identities, per-period artifact requirements, model-built artifact authorities, a mathematical preflight, and a JAX-transformable reader returning named `ActionOutput`; or `DeclaredReplay.GRID_RECOMPUTATION`; or `DeclaredReplay.UNSUPPORTED`, which solves but refuses to simulate. Leaving it unset is a build error for any solver outside the shipped set. A core that reads next-period stored values declares each access through the public `TargetValueAccess` types on `lcm.solvers`.
- An in-repository out-of-tree reference solver that imports only public modules (`lcm.solvers`, `lcm.solver_api`, `lcm.typing`) establishes the conformance contract documented in `docs/reference/custom_solvers.md`, including a solver whose core reads stored target values. `lcm.solver_api` is a facade over topical `lcm._solver_api` modules, none of which imports anything engine-private.

Diagnostics remain governed by `log_level` and per-artifact policy. The plugin and archive surfaces are explicitly versioned experimental APIs.

## Stack

Depends on #437 (`nbegm/expectation-before-inversion`).

Closes #428

https://claude.ai/code/session_01G9giPUNidKrTio57pKUxTG


src/_lcm/solution/model_authority.py
```diff
@@ -276,6 +2193,136 @@ def _policy_shape_and_node_count(
...
+            row_discrete_state_names=policy_read.row_discrete_state_names,
+            row_passive_state_names=policy_read.row_passive_state_names,
+            row_discrete_action_names=policy_read.row_discrete_action_names,
+        )
+
+    if isinstance(policy_read, NNBEGMPolicyRead):
+        if policy_read.replay_policy_is_nested:
+            # The final outer nodes are solve-generated private authority.  A
+            # serialized copy cannot authenticate itself after restoration.
+            return PersistencePolicy.NOT_PERSISTED, None
```
timmens: [P1] Persist adaptive NNBEGM replay state before calling archives complete

AdaptiveOuterMesh is a built-in solver path. This branch declares its nested replay policy NOT_PERSISTED, so a saved result cannot be simulated by a fresh compatible model. That leaves the complete-persistence requirement of #428 unmet. The generated nodes already travel in OuterPolicyBank; please represent the adaptive candidate axis as bounded solution-owned data and validate its ordering, domain, size, and alignment with every policy leaf. Alternatively, narrow the PR persistence claim and leave #428 open.

src/lcm/model.py
```diff
@@ -1518,13 +2047,15 @@ def simulate(  # noqa: C901, PLR0912
                 period_to_regime_to_V_arr,
                 period_to_regime_to_sim_policy,
                 period_to_regime_to_dissolution_flags,
+                period_to_regime_to_replay_reader,
```
timmens: [P2] Preserve the automatic solution on SimulationResult

This branch creates the SolutionResult used by automatic simulation, then projects it into engine inputs. The returned SimulationResult exposes value arrays only, so the solver diagnostics collected during this solve are inaccessible. Issue #428 requires automatic simulation and separate solve-then-simulate to expose the same diagnostics. Please retain the canonical consumed SolutionResult on SimulationResult, for example through a solution property, while sharing the owned value and artifact stores to avoid duplicate buffers.

src/_lcm/solution/model_authority.py
timmens: [P1] Persist adaptive NNBEGM replay state before calling archives complete

AdaptiveOuterMesh is a built-in solver path. This branch declares its nested replay policy NOT_PERSISTED, so a saved result cannot be simulated by a fresh compatible model. That leaves the complete-persistence requirement of #428 unmet. The generated nodes already travel in OuterPolicyBank; please represent the adaptive candidate axis as bounded solution-owned data and validate its ordering, domain, size, and alignment with every policy leaf. Alternatively, narrow the PR persistence claim and leave #428 open.

src/lcm/model.py
timmens: [P2] Preserve the automatic solution on SimulationResult

This branch creates the SolutionResult used by automatic simulation, then projects it into engine inputs. The returned SimulationResult exposes value arrays only, so the solver diagnostics collected during this solve are inaccessible. Issue #428 requires automatic simulation and separate solve-then-simulate to expose the same diagnostics. Please retain the canonical consumed SolutionResult on SimulationResult, for example through a solution property, while sharing the owned value and artifact stores to avoid duplicate buffers.

src/lcm/model.py
```diff
@@ -559,19 +745,35 @@ def _solve_from_flat_params(
...
             internal_result=internal_result,
+            regimes=self._regimes,
+            flat_params=flat_params,
         )
         if retention.retains_replay and any(
             descriptor.adaptive_outer_nodes is not None
             for descriptor in authority.replay.values()
         ):
-            self._solution_authorities[params_fingerprint] = authority
+            self._solution_authorities[params_fingerprint] = (
```
timmens: [P1] Bound adaptive authority to the lifetime of its solution

This dictionary retains one SolutionAuthority for every distinct adaptive-route parameter fingerprint. Each authority owns full-shaped NestedEGMSimPolicy template arrays, so parameter sweeps grow model-lifetime memory linearly even after their SolutionResults are dropped. Please place generated authority on the corresponding in-memory result, use a weak registry tied to result lifetime, or provide another bound that preserves replay for every live solution.

src/_lcm/solution/model_authority.py
timmens: [P1] Persist adaptive NNBEGM replay state before calling archives complete

AdaptiveOuterMesh is a built-in solver path. This branch declares its nested replay policy NOT_PERSISTED, so a saved result cannot be simulated by a fresh compatible model. That leaves the complete-persistence requirement of #428 unmet. The generated nodes already travel in OuterPolicyBank; please represent the adaptive candidate axis as bounded solution-owned data and validate its ordering, domain, size, and alignment with every policy leaf. Alternatively, narrow the PR persistence claim and leave #428 open.

src/lcm/model.py
timmens: [P2] Preserve the automatic solution on SimulationResult

This branch creates the SolutionResult used by automatic simulation, then projects it into engine inputs. The returned SimulationResult exposes value arrays only, so the solver diagnostics collected during this solve are inaccessible. Issue #428 requires automatic simulation and separate solve-then-simulate to expose the same diagnostics. Please retain the canonical consumed SolutionResult on SimulationResult, for example through a solution property, while sharing the owned value and artifact stores to avoid duplicate buffers.

src/lcm/model.py
timmens: [P1] Bound adaptive authority to the lifetime of its solution

This dictionary retains one SolutionAuthority for every distinct adaptive-route parameter fingerprint. Each authority owns full-shaped NestedEGMSimPolicy template arrays, so parameter sweeps grow model-lifetime memory linearly even after their SolutionResults are dropped. Please place generated authority on the corresponding in-memory result, use a weak registry tied to result lifetime, or provide another bound that preserves replay for every live solution.

src/_lcm/solution/fingerprint.py
```diff
@@ -0,0 +1,2465 @@
...
+                return
+            if isinstance(value, Phased):
+                if type(value) is not Phased:
+                    msg = (
+                        "Cannot durably fingerprint Phased subclass semantic value "
+                        f"{type(value).__module__}.{type(value).__qualname__}."
+                    )
+                    raise TypeError(msg)
+                self.frame(label="Phased-solve-start")
+                self.visit(value=value.solve)
```
timmens: [P1] Limit the simulate-side exemption to transition slots

This generic Phased branch drops the simulate member for every regime slot, including current-period utility and other decision primitives. In an end-to-end reproduction, two models differing only in a valid Phased.simulate utility received the same model fingerprint; the second model accepted the first archive while choosing action 1 instead of 0 and reporting value 1 instead of 0. Issue #428 and the runtime documentation exempt simulation-transition truth specifically, so please make this projection slot-aware and fingerprint simulate-side decision primitives, or explicitly broaden the compatibility contract and align parameter projection with it.

src/lcm/model.py
```diff
@@ -454,12 +608,31 @@ def __setstate__(self, state: dict[str, object]) -> None:
...
+    def _structure_fingerprint(self) -> str:
+        """Return this model's parameter-free digest, hashing it at most once.
+
+        The digest covers what the model fixes at build, so it is the same for
+        every parameter vector this instance is ever solved with. Computing it
+        walks every declared user callable, which is the dominant cost of a warm
+        solve when it is repeated per parameter vector.
+        """
+        if self._model_structure_fingerprint is None:
+            self._model_structure_fingerprint = fingerprint_model_structure(
```
timmens: [P2] Keep cached callable identity consistent with execution

This memoizes the callable-sensitive structure digest for the Model instance lifetime. With enable_jit=False, rebinding a directly referenced JAX scalar after the first solve changed the model value from 1 to 7; a fresh structure digest changed, while both SolutionResults kept the same cached model fingerprint and the earlier solution was accepted. Please invalidate or recompute the callable-dependent part before solve and simulation, or enforce a frozen dependency snapshot so executed semantics cannot diverge from the cached identity.

src/_lcm/solution/model_authority.py
timmens: [P1] Persist adaptive NNBEGM replay state before calling archives complete

AdaptiveOuterMesh is a built-in solver path. This branch declares its nested replay policy NOT_PERSISTED, so a saved result cannot be simulated by a fresh compatible model. That leaves the complete-persistence requirement of #428 unmet. The generated nodes already travel in OuterPolicyBank; please represent the adaptive candidate axis as bounded solution-owned data and validate its ordering, domain, size, and alignment with every policy leaf. Alternatively, narrow the PR persistence claim and leave #428 open.

src/lcm/model.py
timmens, line 2240: [P2] Preserve the automatic solution on SimulationResult

This branch creates the SolutionResult used by automatic simulation, then projects it into engine inputs. The returned SimulationResult exposes value arrays only, so the solver diagnostics collected during this solve are inaccessible. Issue #428 requires automatic simulation and separate solve-then-simulate to expose the same diagnostics. Please retain the canonical consumed SolutionResult on SimulationResult, for example through a solution property, while sharing the owned value and artifact stores to avoid duplicate buffers.

src/lcm/model.py
timmens: [P1] Bound adaptive authority to the lifetime of its solution

The existing model-level registry stores one SolutionAuthority for every distinct adaptive-route parameter fingerprint, and this PR snapshots richer artifact authority into each entry, including full-shaped NestedEGMSimPolicy templates. An estimation that solves new parameter vectors therefore retains those arrays after their SolutionResults become unreachable, so memory grows with the number of distinct fingerprints. Because this PR claims requirement 5 of #428, please make the corresponding result own its generated authority, keep only weak model references, or provide another lifetime bound that preserves replay for every live solution.

src/_lcm/solution/fingerprint.py
timmens, line 950: [P1] Limit the simulate-side exemption to transition slots

This generic Phased branch drops the simulate member for every regime slot, including current-period utility and other decision primitives. In an end-to-end reproduction, two models differing only in a valid Phased.simulate utility received the same model fingerprint; the second model accepted the first archive while choosing action 1 instead of 0 and reporting value 1 instead of 0. Issue #428 and the runtime documentation exempt simulation-transition truth specifically, so please make this projection slot-aware and fingerprint simulate-side decision primitives, or explicitly broaden the compatibility contract and align parameter projection with it.

src/lcm/model.py
timmens: [P2] Keep cached callable identity consistent with execution

This memoizes the callable-sensitive structure digest for the Model instance lifetime. With enable_jit=False, rebinding a directly referenced JAX scalar after the first solve changed the model value from 1 to 7; a fresh structure digest changed, while both SolutionResults kept the same cached model fingerprint and the earlier solution was accepted. Please invalidate or recompute the callable-dependent part before solve and simulation, or enforce a frozen dependency snapshot so executed semantics cannot diverge from the cached identity.

src/_lcm/persistence/solution.py
```diff
@@ -0,0 +1,2483 @@
...
+                )
+            if ref in omission_entries:
+                raise IncompatibleSolutionError(
+                    f"Artifact {ref!r} is both present and explicitly omitted."
+                )
+            present_refs.add(ref)
+            descriptor = descriptors.get(ref)
+            authority = authorities.get(ref)
+            if descriptor is None or authority is None:
+                if ref.key in _UNDESCRIBED_STANDARD_KEYS:
```
timmens: [P2] Preserve retained solver diagnostics in complete archives

This branch converts every standard SOLVER_DIAGNOSTICS entry into a NOT_PERSISTED omission. In a normal warning-level adaptive NNBEGM solve, diagnostics were present before save and the restored result had an empty diagnostics store plus NOT_PERSISTED records. Issue #428 includes solver diagnostics in the complete solution boundary, so please give the standard diagnostic payload model-verifiable authority and persist its numerical leaves, or narrow the archive guarantee explicitly.

src/_lcm/persistence/solution.py
```diff
@@ -0,0 +1,2483 @@
...
+            or prepared.leaves[0].dtype != np.dtype(schema.dtype)
+        ):
+            raise IncompatibleSolutionError(
+                f"Value at {(period, regime)!r} does not match its declared schema."
+            )
+        prepared_values.append(prepared)
+
+    descriptors = dict(metadata.artifact_descriptors)
+    if set(authorities) != set(descriptors):
+        raise IncompatibleSolutionError(
```
timmens, line 937: [P2] Allow a restored solution to use its public save method

load_solution returns an exact public SolutionResult with a save method, yet every artifact-bearing restored result has an empty private authority map and fails this equality check. A checksum-verified load followed by restored.save deterministically raises IncompatibleSolutionError before producing the copy. Please preserve enough validated archive authority to re-emit lazy entries safely, or provide a dedicated archive-copy path and make the method contract explicit for persisted results.

src/_lcm/solution/contract.py
```diff
@@ -586,6 +618,47 @@ class SolutionKernels:
     whose scope is decided by structure alone.
     """
 
+    replay_route: ExecutableReplayRoute | None = None
```
timmens: [P2] Require external solvers to declare replay semantics

`replay_route=None` is accepted for an external solver and later interpreted as the built-in grid-recomputation route. A focused custom-solver reproduction solves a tie by selecting the middle action 0.5, then separate simulation succeeds while choosing 0.0 for every subject because the engine silently replays with grid argmax. If actions feed later states, this can compound into incorrect trajectories and welfare without any compatibility error. Please require external solvers to declare an explicit public grid-recomputation route when it matches their solve semantics, or an explicit unsupported marker when exact replay is unavailable; treat an omitted route as a construction or preflight error.

src/lcm/solvers.py
```diff
@@ -51,11 +51,11 @@
 and helpers (`ContinuationSpec`, `EGMContinuationSpec`, `EGMContinuationLayout`,
 `period_to_continuation_target`, `target_period_grid`, `union_free_params`,
 `union_fixed_params`), and the artifact vocabulary (`KernelOutput`,
-`ArtifactKey`, the four keys, `ContinuationArtifact`, `ReplayMode`,
-`ReplayRoute`) — so no import from `_lcm` is needed. The surface is experimental
-until persistence, durable identity, a conformance suite, and a versioning policy
-exist; `docs/reference/custom_solvers.md` states what a solver owes the engine
-and what is still missing.
+`ArtifactKey`, descriptors and authorities, the four built-in keys,
+`ContinuationArtifact`, and the executable replay contracts) — so no import from
```
timmens, line 55: [P2] Complete the public target-value access contract

`lcm.solvers` promises that an out-of-tree solver needs no `_lcm` import, yet `CoreExecutionRequirements.target_value_accesses` accepts only the private `_TargetValueAccess`, whose fields in turn require private `ValueArtifactAddress` and `ValueConsumerAddress`. An external core that reads `next_regime_to_V_arr` therefore cannot declare that dependency through the public surface; it must leave the tuple empty or import engine internals, and the reference solver does not exercise this path. Please export stable public address/access types or provide a public helper, and add a conformance case whose core reads a next-period value.

src/lcm/model.py
```diff
@@ -692,42 +899,88 @@ def _resolve_simulate_regimes(
...
     ) -> tuple[
         PeriodToRegimeToVArr,
         PeriodToRegimeToSimulationPolicy,
         PeriodToRegimeToDissolutionFlags,
+        _PeriodToRegimeToReplayReader,
     ]:
         """Resolve one model-authoritative result into engine replay inputs."""
-        authority = self._check_solution_result_structure(
-            solution=solution, flat_params=flat_params
+        solution = self._snapshot_solution_envelope(solution=solution)
```
timmens, line 1091: [P1] Avoid repeated copies of solution buffers

On the current head, the GPU comparison reports GridSearch solve and simulation regressions of 16–28% and 12–24%, reference-chain solve regressions of 52–70%, and DC-EGM simulation rising from 270.9 ms to 785.3 ms. A focused current-head probe found every value array copied once while wrapping a no-replay GridSearch solve and three times before each simulation. In the DC-EGM case, which retains zero replay artifacts here, every omitted continuation-template leaf is copied nine times although simulation never consumes it. This violates requirement 5 of #428. Please provide one validated internal execution view that borrows owned in-memory buffers, materializes each restored entry once, and limits authority work to artifacts consumed by active replay routes, while retaining defensive copying at the public store boundary.

src/_lcm/execution/core_program.py
```diff
@@ -6,14 +6,16 @@
 or runs it deliberately dense, and refuses a kernel that publishes no graph.
 """
 
+from __future__ import annotations
```
timmens: [P2] Use the project annotation semantics

This project requires Python 3.14+ and its critical typing standard explicitly forbids `from __future__ import annotations`, because package-wide beartype relies on PEP 649 runtime evaluation. This is the only future-annotations import added by this PR. Please remove it and resolve any resulting forward references with the project’s `TYPE_CHECKING` and runtime-alias pattern.

src/lcm/solver_api.py
```diff
@@ -5,56 +5,3242 @@
...
+
+def _canonical_value_entry(*, value: object) -> _CanonicalValueEntry:
+    """Detach one eager value before any lazy result callback may run."""
+    source = value.materialize() if type(value) is _CanonicalValueEntry else value
+    private = _copy_solution_value(value=source, label="Solution value")
+    return _CanonicalValueEntry(value=private)
+
+
+@dataclass(frozen=True, kw_only=True)
+class ReplayRouteSnapshot:
```
timmens: [P2] Keep the public solver API shallow

`lcm.solver_api` now spans 4,331 lines. Major public replay contracts begin here after thousands of lines of private artifact machinery, and `SolutionResult` appears at line 4,218, so the advertised interface is difficult to inspect independently of its implementation. Please keep `lcm.solver_api` as a stable re-export facade over focused private modules, or reorder the module so its public contracts form the top-level interface. A facade preserves current imports while reducing coupling before this experimental API ships.

src/lcm/solver_api.py
```diff
@@ -5,56 +5,3242 @@
...
+            raise TypeError("ArtifactKey.schema_version must be an exact int.")
+        if self.schema_version < 1:
+            raise ValueError("ArtifactKey.schema_version must be at least 1.")
+
+
+@dataclass(frozen=True, order=True, kw_only=True)
+class SolverIdentity:
+    """Durable identity and compatibility version of an installed solver plugin."""
+
+    plugin_id: str
```
timmens: [P2] Document the new public dataclass fields

The project requires an inline PEP 257 field docstring for every frozen config or state dataclass field. A complete base-to-head AST comparison finds 74 added undocumented fields across 20 public dataclasses, starting with this identity type and continuing through artifact descriptors and authorities, replay contexts, stores, metadata, and result types. Please document each field’s meaning and invariants in place; one mechanical pass can cover the surface.

timmens (changes requested):
Requesting changes because four in-scope issues break the complete-solution contract introduced here:

1. Adaptive NNBEGM replay state is omitted from archives, so a fresh compatible model cannot reproduce the saved policy.
2. Parameter-specific adaptive authority remains owned by Model, so memory grows across unique parameter solves after their results are dropped.
3. The fingerprint excludes simulate-side decision primitives, so an incompatible archive can be accepted and produce different actions and values.
4. Internal execution repeatedly copies owned buffers, causing material solve and simulation regressions.

The nine P2 comments cover result access, archive fidelity, external-solver contracts, callable identity, API structure, typing, and documentation. All findings concern behavior or public surfaces added by this PR.

Reviewed head: 39eabc4.

src/_lcm/solution/model_authority.py
hmgaudecker: Done. The adaptive NNBEGM policy now carries its solve-generated outer nodes as the candidate axis of its own descriptor, so it is model-verifiable and persists under `ALL_PERSISTABLE_ARTIFACTS`. A consuming model admits the nodes only after checking that they are exact finite floats, strictly increasing, within the search's node budget, inside the outer state's declared domain for that period, and that every policy leaf's shape agrees with the axis lengths; only then is the rest of the descriptor compared against the model's own authority. A restored adaptive result replays in a freshly constructed model in `tests/solution/test_builtin_policy_persistence.py`, and the retention tests in `tests/solution/test_nbegm_retention_specialization.py` cover both outer searches.

src/lcm/model.py
hmgaudecker: Done. `SimulationResult.solution` returns the `SolutionResult` the simulation consumed, whether the caller passed it or the automatic solve produced it, sharing the owned value and artifact stores rather than copying them. It is cleared when the simulation result is saved and is `None` for a result read back with `load`.

src/lcm/model.py
hmgaudecker: Done. Authority is now split at the boundary no parameter vector crosses. The declared authority is bound once per model instance and grid support (a small cache keyed by the support fingerprint), while every solve-generated authority, including the full-shaped adaptive policy templates and the diagnostics descriptors, lives on the `SolutionResult` it describes and is released with it. The model holds no per-solve arrays.

src/_lcm/solution/fingerprint.py
hmgaudecker: Done. The projection is slot-aware: only the transition slots of a `Phased` declaration are exempt from simulate-side truth, and a `Phased.simulate` utility, constraint, or other decision primitive enters the structure digest. Two models differing only in such a simulate member now have different fingerprints, and the parameter projection follows the same slot rule.

src/lcm/model.py
hmgaudecker: Done, by sealing the model at build rather than recomputing per call. `Model(...)` digests the structure once and records every global and closure binding its declared callables read; `solve()` and `simulate()` refuse with `ModelSealError` naming the binding once one has been rebound, so the cached identity can never describe code the model no longer runs. A pickled model is resealed on load, and a model whose callables cannot be fingerprinted is refused at construction with `ModelInitializationError`.

src/_lcm/persistence/solution.py
hmgaudecker: Done. Each retained `SolverDiagnostics` payload is described by a model-verifiable descriptor the solve generates from the payload itself (dtype kinds, optional fields, and the diagnostics pytree registration). The archive persists its leaves, the loader attaches the template so a restored archive reads its diagnostics without a model, and a consuming model admits the descriptor only when it names exactly the published fields with the dtypes they carry. A diagnostic omission without a descriptor is refused on both save and load. The solution-format version is 2.

src/_lcm/persistence/solution.py
hmgaudecker: Done. A result restored by `load_solution` saves again through a dedicated archive-to-archive path: each payload is re-read from the archive it came from, verified against its checksum and descriptor, and written to the new archive without a model. Omission semantics are re-validated on the way, and the equality check you point at no longer applies to persisted-source results.

src/_lcm/solution/contract.py
hmgaudecker: Done. `SolutionKernels.replay_route` must now be an `ExecutableReplayRoute`, `DeclaredReplay.GRID_RECOMPUTATION`, or `DeclaredReplay.UNSUPPORTED`. A solver outside the shipped set that leaves it unset is refused at model build naming the regime and solver; `UNSUPPORTED` solves but refuses `simulate()` with the reason. `DeclaredReplay` is exported from `lcm.solvers` and the out-of-tree tests cover all three outcomes, including that grid recomputation takes the argmax over the declared action grid.

src/lcm/solvers.py
hmgaudecker: Done. `TargetValueAccess`, `ValueArtifactAddress`, `ValueArtifactKind`, `ValueConsumerAddress`, and `ValueInputChannel` are public through `lcm.solvers`, and `SolverBuildContext.solution_reachability.targets(period=..., source=...)` names the targets a core has to declare. The conformance fixture gained `TargetValueSolver`, whose core reads `next_regime_to_V_arr` through declared accesses, and `docs/reference/custom_solvers.md` documents the contract.

src/lcm/model.py
hmgaudecker: Done. A result is now consumed according to its provenance: one the same model instance solved for the same canonical parameters is read by reference through an owned view, so a simulate following a solve copies and re-validates nothing; a restored, unpickled, or foreign result is validated in full and materialized once per consuming model and parameter vector, and later simulations reuse that view. Authority work is limited to the artifacts the active replay routes consume, and the public store boundary keeps its defensive copies (`tests/solution/test_solution_consumption.py`). The benchmark comment on this PR will show the effect once the benchmark job has run.

src/_lcm/execution/core_program.py
hmgaudecker: Done. The future import is gone; the forward references resolve through the project's `TYPE_CHECKING` and runtime-alias pattern.

src/lcm/solver_api.py
hmgaudecker: Done. `lcm.solver_api` is now a re-export facade that keeps every public name and its `__all__`; the definitions live unchanged in `lcm._solver_api` (identity, descriptors, contract, authority, entries, replay, stores, result) in dependency order, and the import test checks that none of them imports anything engine-private.

src/lcm/solver_api.py
hmgaudecker: Done. Every field of the public solver-API dataclasses carries an inline PEP 257 docstring.

timmens (approved):
Thanks, the requested changes address the blocking correctness and lifecycle issues. The latest benchmark shows solve-time variation in several cases, including double-digit changes for some solve benchmarks. I consider this non-blocking given expected runtime variation across concurrent PRs, though it is worth monitoring.

PR #440, opened 2026-09-05
ENH Plan solver and simulation execution widths and device placement

### LINKED ISSUE #443: simulate() offers no way to hold the taste-shock stream fixed across two models
## What is missing

`Model.simulate` accepts `seed: int | None = None` (`src/lcm/model.py:1342`) and nothing else that controls randomness. Inside, the seed becomes a single root key —

```python
key = jax.random.key(seed=seed)          # src/_lcm/simulation/simulate.py:383
```

— from which every stream is derived, the taste-shock stream among them:

```python
taste_shock_kwargs = {"taste_shock_key": gumbel_keys["key_taste_shock"]}   # :866
...
next_states_key, next_regime_key, key = jax.random.split(key=key, num=3)   # :955
```

So `taste_shock_key` exists as an internal concept but is not reachable from the public API.

## Why a seed is not enough

The taste-shock stream shares its derivation chain with state transitions and regime transitions, and the chain is advanced per period by a `split`. Two simulations that differ in the model — a counterfactual that changes a transition, a different horizon, a different regime structure — therefore consume the chain differently and get **different taste-shock draws at the same seed**.

That makes common random numbers across arms unobtainable. A comparison between a baseline and a counterfactual then carries taste-shock noise that no seed choice removes, and the difference between the two arms is not attributable to the policy change alone.

## Concrete consequence downstream

In `lcm-zoo`, `arcidiaconoCollegeAttritionDynamics2025` implements extreme-value type-1 taste shocks and is **solve-only** because of this: no simulation parity check can be written at all, and the paper's audited contract records it under `known_engine_limitations`. Re-verified at `0.0.2.dev173+g98be11fbe`, the commit that repository pins, rather than inherited from an earlier reading.

## What would resolve it

Some way for a caller to fix the taste-shock stream independently of the rest — an optional `taste_shock_key` on `Model.simulate` that overrides the derived one is the smallest version, but the design is yours; a separate seed per stream would serve the same purpose.

The property worth having, whatever the spelling: **two simulations that differ in the model can be made to share taste-shock draws.** A useful acceptance test is that two models differing only in a policy parameter produce identical taste-shock realizations, which no seed can currently deliver.

## Not asking for

Reproducibility of a single run — that already works. This is specifically about holding one stream fixed while the model around it changes.

https://claude.ai/code/session_01QipFjBYxb5aPnmZE76cxhi

### LINKED ISSUE #450: Log the resolved simulation execution plan at debug level
## Problem

A production simulation run (ACA, `simulation_sharding="subjects"`, 3 A40s, pylcm 01d6654f) completed cleanly but its log carried no evidence that the subject-sharded route engaged: the simulation section logs transitions only, and neither `simulate.py`, `runtime.py`, `subject_devices.py` nor `chunk_admission.py` records the resolved route, subject devices, resolved axis widths or chunking at any log level. `SimulationResult` metadata does not carry the `ExecutionConfig` either. So a clean run cannot distinguish "the fix works" from "the requested sharding silently did not engage"; the only positive controls today are a rejection at the admission boundary or a forced-host-device regression test.

## Proposal

At `log_level="debug"` (and possibly `"progress"`), log once per simulate call the resolved execution plan: route (`legacy`/`subjects`), subject device ids and backend, resolved axis widths per regime, outer chunk count and admitted widths, and the budget mode. Optionally expose the same record on the result for artefact-level checks. This is diagnostic only; no numerical, RNG, ownership or admission behaviour changes.

Raised from the aca session's report on jobs 27624358/27624359 during PR #440 closure.

### LINKED ISSUE #451: Broadcast pruning drops a regime's entry law for a model-level state it does not read
## Symptom

Promoting a state from regime-level to model-level (`Model(states={..., "health": DiscreteGrid(Health)})`) makes construction fail with

```
ModelInitializationError: solution phase, period ... source 'single_f_er' -> ... target 'single_f_ret' retains a CONDITIONAL edge. The target declares state 'health', but the source does not carry 'health' and defines no entry law ...
```

even though the source regime declares exactly that entry law:

```python
# single_er regime
state_transitions["health"] = {own_ret: MarkovTransition(initial_health_draw)}
# couple_ret regime
state_transitions["health"] = {"widow_f_ret": MarkovTransition(next_health_f), "widow_m_ret": MarkovTransition(next_health_m)}
```

With `health` declared only on the retirement regimes (regime-level), the same laws are accepted as target-keyed entry laws and the model builds. Observed on the Borella marriage model (lcm-zoo) at pylcm `a675a4e2` while preparing `sharded_states=("health",)`, which requires a model-level state.

## Mechanism

`prune_broadcast_variables` (`src/_lcm/regime_building/broadcast.py`) prunes a broadcast state from a regime whose DAG never reads it, and "for states also the regime's law entry for it". An entry law such as `MarkovTransition(initial_health_draw)` does not read `health`, so after promotion `health` is pruned from `single_er` and its `state_transitions["health"]` mapping is dropped with it. `_state_handoff_errors` (`processing.py`) then finds neither the state nor the law on the source and refuses the edge.

The law was never about the source's own `health`: it is a target-keyed entry law toward regimes that keep the state. Pruning the state from the source is right; dropping the entry law toward a target that retains the state is not.

## Expected

Pruning keeps `state_transitions[name]` entries that are target-keyed mappings toward regimes which retain `name` (and only drops the source's own transition of a pruned state). Equivalently: promoting a regime-level state to model-level must not change which entry laws exist.

## Reproducer sketch

Two regimes A -> B on consecutive periods; `x` a `DiscreteGrid` read only in B's utility; A declares `state_transitions["x"] = {"B": MarkovTransition(entry_draw)}`. Declaring `x` on B only builds; declaring `x` in `Model(states=...)` fails with the message above.

## Context

Blocks discrete sharding on any state that enters some regimes through an entry law rather than being carried everywhere. Found by the Borella session; relates to the model-level-state requirement of `ExecutionConfig.sharded_states` in #440.

### LINKED ISSUE #452: Value transfer of a ProjectedRegimeValue fallback fails between regimes sharded on differently named states
## Symptom

At `39ed7ec5` (with #451 fixed and a sharded state allowed to be pruned from non-terminal regimes), sharding the Borella marriage model on the wage shock — `eps` in single regimes, `eps_f` in couple regimes (`eps_m` left unsharded because sharding both needs `n_eps**2 = 25` devices) — builds, but solve fails while resolving `couple_work`'s `ProjectedRegimeValue` fallback on the divorce edge to `single_f_work`:

```
ValueError: The source sharding is incompatible with value shape (3, 5, 3).
  _lcm/execution/value_transfer.py:908  _check_sharding_shape
  during _resolve_value_input_transfer_plan
```

The sharding being checked is `NamedSharding(mesh=Mesh('eps_f': 3), spec=P('eps_f', None, None, None, None, None))` — rank 6, `couple_work`'s own layout — against `(3, 5, 3)`, which is `single_f_work`'s value shape `(eps, wealth, ybar)`. That regime has no `eps_f` axis at all; its sharded axis is `eps`.

## Reading

The transfer planner applies the consuming regime's sharding spec to a value read from a different regime through the projected fallback, instead of the source regime's own stored layout. Same-name sharded states across regimes work (the ACA `pref_type` route); the failing case is two regimes each sharded on a differently named discrete state, linked by `ProjectedRegimeValue`/`StakeholderRoute` machinery (marriage/divorce). Reachable only once #451 and the pruning relaxation landed.

## Expected

The fallback read classifies the transfer from the source regime's stored sharding (mesh over `eps`) to the layout the consumer requires (mesh over `eps_f`, different axis name and extent alignment), via the catalogue (`CROSS_MESH_COPY`/`RESHARD`), or refuses at construction with a message naming the two regimes and their sharded axes if the route is unsupported.

## Reproducer

Borella lcm-zoo model at pylcm 39ed7ec5: promote `eps`, `eps_f`, `eps_m` to model-level states (per-regime `state_transitions` unchanged), `ExecutionConfig(devices=tuple(range(8)), sharded_states=("eps", "eps_f"))`, `.solve()`. A minimal two-regime version (A sharded on `x`, B sharded on `y`, B carries a `ProjectedRegimeValue` fallback onto A) should reproduce.

## Context

Found by the Borella session while trying solve-side sharding of the working-life wage shock; reverted. Out of scope for #440 unless HMG decides otherwise.

hmgaudecker:
Execution policy moves out of grids and solver constructors into `Model.execution_config`. Programs declare their axes, reads and output layouts; a planner uses compiler memory reports and the storage a solve retains to choose execution widths per regime and to place regimes on device submeshes. Solver API version 3 carries the execution-capability metadata this needs; version-2 archives are no longer loadable.

## What changes for users

- **Execution configuration.** `ExecutionConfig(device_memory_bytes=..., axis_widths={...}, device_memory_headroom_fraction=...)` replaces the width and batch knobs that used to live on grids and `Model.simulate(subject_batch_size=...)`. A budgeted solve or simulation is admitted only if its complete footprint — retained values, workspace, and every planned transfer's own storage on every device it touches — fits below the effective budget, which is capped at a fraction below the allocator pool so a request that would exhaust the device is refused by name instead of failing inside the compiler. Unbudgeted calls keep asynchronous dispatch and a prepared fast path for repeated small calls.
- **Sharding follows the model.** A discrete state may be sharded across devices; regimes that prune it run on one device and only a state every regime prunes is refused. Values read across regime meshes are planned from the layout they are stored on, gated edges may project a sharded state onto the referenced regime's grid, and gated edges simulate under subject sharding. Multi-device solve and simulation results match single-device references exactly on the covered models.
- **Simulation.** Decision, transition and routing programs are declared and planned like solve programs. The subject width is chosen from the budget when one is given and from a byte cap otherwise; the resolved plan (route, devices, widths, chunks, effective budget) is logged once per call and exposed on the result. `Model.simulate(taste_shock_seed=...)` gives counterfactual models a common standardized taste-shock stream (#443).
- **Solver and persistence.** Every built-in kernel runs on the public execution contract, out-of-tree solvers have a documented contract, solutions persist completely, engine functions are defined once rather than per call, and the MSS upper envelope decides its orderings from the stored operands.
- **Fixes carried in this branch:** entry laws survive broadcast pruning (#451), broadcast pruning closes both phase slices jointly, a value transfer between regimes sharded on differently named states is planned from the stored layout (#452), the resolved simulation plan is logged (#450), device-memory headroom below the allocator pool, and a gated projection reading a co-mapped sharded state no longer fails at solve.
- **Documentation** for the execution and certification layers, the sharded CI workflow, and the debugging/tuning/subject-parallel user guides; a module docstring on every engine module.

`CHANGES.md` has the full entry list.

## Verification

- CI is green at the head commit on every workflow (CPU at both precisions across sharded general, slow and solution lanes on Linux, macOS and Windows; GPU; type checks; notebooks; bibliography; benchmark).
- The complete test suite passes at both float64 and float32 on cluster CPU nodes at the head commit, with a pinned clean checkout recorded per run.
- Multi-device GPU differentials (3×A40): forward, mirror and subject-sharded models solve and simulate with zero difference against single-device references; a production-scale structural model completes solve and simulation on three A40s under the headroom cap, and the same configuration without the cap reproduced the allocator failure the cap prevents.
- Two rounds of external numerical review of the execution and admission changes closed with the accounting repairs landed; the remaining requests were for additional resource measurements, which were declined as out of scope for this PR.

## Performance status, stated plainly

Relative to this branch's own earlier state, warm solve and simulation on the reference workload improved by roughly 2×, cold calls are unchanged or better, and repeated same-shape calls compile nothing. Relative to `main`, the branch is still slower per call where the budgeted admission path runs: the GPU benchmark comment on this PR flags 21 rows, with the reduced production benchmark at 3.2–3.4× main warm across the last two runs and the small models at 1.1–1.35× warm; peak GPU memory rises on several small benchmarks because the byte-cap chooses wider subject tiles. These are the known cost of budgeted admission and of the width policy, they are accepted for this merge, and the per-call host residual is the next performance workstream. Unbudgeted calls sit at 1.1–1.3× main on the same benchmarks.

Closes #443
Closes #450
Closes #451
Closes #452


timmens:
## Review of #440

Reviewed head `2343a0d9f`, merge base `1aaa87bdf`.

Size at this head: 611 files, +137,047 / -11,502, 489 commits. 60 new modules under
`src`, 90 modified, none deleted, 183 new test files.

Measurements below were taken at `3d10d674f`, nine commits earlier. Every finding was
re-verified at `2343a0d9f`; `backward_induction.py` is byte-identical between the two and
the nine commits touch only simulation-side source, so the solve timings stand.

## What this PR does

Execution policy, meaning how a model is chopped up to fit a particular machine, moves
out of the objects that describe economics and into one `ExecutionConfig` passed to `Model`.
Twelve knobs disappear: `batch_size` and `distributed` on every grid, and ten width
fields across `DCEGM`, `NEGM`, `NBEGM` and `NNBEGM`. A new planner uses the compiler's
own memory reports to choose execution widths and to place regimes across devices.
Solver API 3 and the rewrite of simulation follow from that.

Riding along: a shared taste-shock seed for counterfactual comparison (closes #443), and
a restructuring of the repository's agent instructions.

## What was and was not reviewed

Reviewed in depth:

- The public `ExecutionConfig` surface and how it is resolved and frozen at model build.
- The solver contract additions, checking which are mandatory and which default.
- `backward_induction.py`, line-read across roughly 1,900 of its 4,828 lines: all of
  `solve()`, all of `_compile_all_functions()`, and the memory lifetime machinery. No
  defects found.
- Behaviour testing of the invariants a rewrite of this kind breaks silently.

Trusted on the strength of its tests, not read: large parts of the execution planner and
the simulation layer, and `_roll_gated_edges` / `_evaluate_edge_fold` in backward
induction. At this size a full read is not feasible in one pass, and that should be a
conscious acceptance rather than an assumption.

## Behaviour checks that passed

These are the claims where a bug would produce wrong numbers rather than an error.

| Claim | Result |
| ----- | ------ |
| Execution width must not change the answer | Bit-identical value functions at `cell` widths 4, 16, 64 against the planner's own choice, every period and regime |
| Device placement must not change the answer | Bit-identical across 1, 2 and 4 devices with a sharded state (stronger than the changelog's "a few units in the last place") |
| Subject chunking must not change simulated draws | Identical dataframes at subject widths 1, 3, 5 versus unbatched, with 13 subjects so no width divides evenly |
| `taste_shock_seed` is independent of the ordinary seed | Same taste seed with ordinary seeds 7 and 555 gives identical discrete choices; omitting it preserves ordinary seeded behaviour |

## Recommendation

Merge. The PR buys a five- to twenty-fold reduction in peak GPU memory and 30-60% faster
first runs, at a warm-execution premium. For models that previously did not fit, that is
not a trade at all. There are no users to break, and correctness held everywhere it was
tested.

Two things worth doing first, both cheap:

- Wait for the GPU benchmark currently running. The posted benchmark predates all four
  performance fixes, so it is the wrong evidence either way.
- Record the open question in G6-a while the context is fresh.

None of the comments below blocks the merge.

## Comments

### C1: the changelog tells users to keep an argument that no longer exists

- Placement: `CHANGES.md`, line 128.
- Severity: P3, non-blocking.

The subject-chunking entry ends with "`Model(n_subjects=...)` remains a prewarm hint."
That argument is gone. Anyone following this sentence gets a `TypeError`.

Confirmed by inspecting the signature at this head:

```
n_subjects in Model.__init__:      False
Model has n_subjects attribute:    False
subject_batch_size in simulate:    False
```

The removal itself is deliberate and consistent with the rest of the PR: the old
ahead-of-time compile cache is replaced by runtime executors keyed on subject shape.
Only the sentence is wrong, and it happens to sit in the migration instruction for a
breaking change.

Please either say `n_subjects` is removed and name what replaces the prewarm behaviour,
or restore the argument.

### C2: a 364 KB run receipt from one cluster job is stored in the repository

- Placement: `acceptance/aca-phase-inventory.json`.
- Severity: P3, non-blocking.

`acceptance/README.md` and `benchmarks/continuous_scaling/README.md` make clear that
these directories are committed on purpose, so an exact revision can be pulled onto
Marvin. That design is not in question here.

The narrower point is this one file. It is 363,723 bytes recording a single Slurm run,
and it hard-codes ten absolute paths from the machine it ran on, including
`/home/hmg308_hpc/aca-residency-projection-01-620813df/preflight` and
`/tmp/aca-sharding-green.I4Rtqy/green/pylcm`, together with an HPC account name. None of
those resolve for any other reader, and the file grows the repository permanently.

If the receipt has to be version-controlled, stripping the absolute paths would at least
make it readable to someone who was not on that machine. If it only needs to be
retrievable, it may belong with the job outputs rather than with the source.

### C3: plain `pytest` no longer runs, and one line fixes it

- Placement: `pyproject.toml`, near `norecursedirs` at line 436.
- Severity: P3, non-blocking.

Clone the repository and type `pytest` with no arguments and it stops before running
anything:

```
ERROR acceptance/test_aca_phase_boundaries.py                 No module named 'aca_model'
ERROR benchmarks/continuous_scaling/test_gpu_acceptance.py    No module named 'gpu_admission'
Interrupted: 2 errors during collection
```

At the merge base the same command collected 8,970 tests and exited cleanly.

Neither file can run here by design. `aca_model` is a separate package, `gpu_admission`
exists only as a `.patch`, and the GPU file requires eight physical A40 devices. Both are
cluster scaffolding that travels in the repository deliberately. The collateral damage is
that they are named `test_*.py`, so plain `pytest` finds and tries to import them.

CI and every `pixi` task escape this because each names `tests` explicitly. Only the bare
command breaks, which is what a new contributor is most likely to type.

The fix:

```toml
ini_options.testpaths = [ "tests" ]
```

`testpaths` supplies a default only when pytest is given no arguments, so it does not
restrict anything. Verified with that line added (collection counts from `3d10d674f`;
both offending files are still present at `2343a0d9f`):

| invocation | result |
| ---------- | ------ |
| `pytest` | 12,365 collected, no errors |
| `pytest acceptance/` | still reaches the file and attempts the import |
| `pytest acceptance/test_aca_phase_boundaries.py` | same |

The two explicit runs still stop on `ModuleNotFoundError: No module named 'aca_model'`,
which is this environment rather than the configuration; on a Marvin checkout where that
package is installed they run normally. The cluster workflow invokes files by explicit
path, so it is unaffected.

The line also gives the repository a definition of what "the test suite" means, which is
worth having if scaffolding is going to keep living outside `tests/`.

### C4: the changelog states one bootstrap cap where there are now two

- Placement: `CHANGES.md`, line 114. Post together with C1.
- Severity: P3, non-blocking.

The entry reads "Without a budget, streamed solve axes use their bootstrap widths, capped
at 64". Commit `478acef3c` split that rule: an axis whose block is purely temporary keeps
the cap of 64, while an axis whose tiles are concatenated into a full-size result now
bootstraps under a cap derived from `64 * 512`, clamped to `[64, 512]`.

So the unbudgeted cap is now up to 512 in one of the two cases. That same commit reports
this split as the fix for most of the warm-solve regression, which makes it the entry a
performance-sensitive reader is most likely to act on.

### G1-a: allow an execution width to be set per regime

- Placement: `src/lcm/execution.py`, the `axis_widths` field.
- Severity: P2, non-blocking.

`axis_widths` applies to the whole model, so one width covers every regime declaring that
axis. The planner it overrides is finer than that: left alone, it chooses per regime and
per period.

Two regimes with opposite shapes, `cheap` at 4,096 state cells and two actions, `heavy` at
16 state cells and 2,048 actions, showing the widths the planner actually compiled:

```
no width declared:
  cheap   4096 cells   cell width 512   ->    8 chunks
  heavy     16 cells   cell width   8   ->    2 chunks

axis_widths={'cell': 8}:
  cheap   4096 cells   cell width   8   ->  512 chunks
  heavy     16 cells   cell width   8   ->    2 chunks
```

Pinning `cell: 8` because `heavy` needs it takes `cheap` from 8 chunks to 512, a
sixty-four-fold increase in dispatches for a regime that had no memory problem.

Proposal, as discussed: accept both forms in the one field. A bare integer broadcasts to
every regime; a regime key overrides it for that regime alone. The plumbing looks
contained. `axis_widths` reaches the planner through `_CanonicalRegimeBuilder`, which is
built once and then called per regime, so it already knows which regime it is serving.

### G1-d: note in the docstring that continuous sharding is a narrow route

- Placement: `src/lcm/execution.py`, the `sharded_states` docstring.
- Severity: P3, non-blocking.

The docstring reads "States whose grid axis is spread over the regime's devices", which
describes the discrete case well. The continuous case requires, all at once: exactly one
sharded state, a model-level `LinSpacedGrid`, the same grid object retained in every
regime, `GridSearch` everywhere, singleton hard-max, and no stakeholders, gated edges,
value constraints, same-period references or taste shocks.

All of that is checked at model build and the error lists every requirement, so nothing
goes silently wrong. One sentence separating the two cases would let a reader of the
public class predict whether their model qualifies.

### G5-b: no test covers two regimes running at once with buffer donation

- Placement: `src/_lcm/solution/backward_induction.py`, the `solve()` period loop.
- Severity: P3, non-blocking. Maintainer-facing only; no user can reach this.

To stay inside a memory budget, the solver hands memory buffers back to the compiler to
reuse. Handing one over while something else still needs it would corrupt results. The
code guards this correctly, but through a convention rather than a mechanism.

`solve()` threads three mappings through each period. A release or a donation does not
edit them in place, it replaces all three. So any `SolveInputMappings` built before that
point still names buffers that have since been freed, and every use has to construct a
fresh one. The file does exactly that, eight times, several within a few lines of each
other. The repetition looks redundant and is load-bearing.

Replacing the three in-loop constructions with the `period_inputs` snapshot already in
scope, which is the tidy-up a maintainer would plausibly make, leaves the suite green:

```
tests/solution/test_donation_solve.py + test_release_log.py     30 passed
tests/execution + tests/solution (not slow)                    835 passed, 20 skipped
```

Including both files written specifically to test donation and release. The checkout was
restored and re-verified clean afterwards.

Why nothing catches it: a stale snapshot equals a fresh one until the first release or
donation *within that period*, which needs two or more dispatch units in one period, so
two regimes active at once with the earlier one donating. Almost every test model runs a
single acting regime per period, and `NBEGM` is the only shipped solver declaring
donation candidates. The production models have that shape; the fixtures do not.

Suggested fix, a test rather than a comment: one solve with two regimes active in the same
period where the first donates, asserting the published values match a non-donating twin.
That pins the invariant and covers a configuration the production models use.

The cheaper alternative is one sentence on `SolveInputMappings` saying that a release or a
donation replaces the mappings, so an instance must be built where it is used and never
held across one.

### G6-a: one measurement is missing behind the runtime premium

- Placement: PR summary comment.
- Severity: P2, non-blocking. This records a missing measurement.

Warm solve, precautionary savings with Rouwenhorst shocks, twelve periods, CPU, median of
five runs, merge base `1aaa87bdf` against `3d10d674f`:

| model size | main | PR, default widths | PR, widths pinned to full extent |
| ---------- | -------: | ----------------: | -------------------------------: |
| 200 x 200  |  66.6 ms |  81.1 ms (1.22x)  |                 80.9 ms (1.22x) |
| 600 x 600  | 355.8 ms | 466.5 ms (1.31x)  |           529.6 ms (**1.49x**) |

The pin takes effect: the planner's default compiles at `cell: 512`, and pinned it
compiles at `cell: 1400`, the full extent.

Before this PR, grids defaulted to whole-axis execution, so `main` was already running at
full width. The third column is therefore the like-for-like comparison, same model, same
execution shape, same widths, and it is the **worst** ratio of the three. That gap cannot
be chunking policy, and at identical widths it is not obviously per-dispatch overhead
either, since the number of dispatches matches `main`.

Two explanations fit, with very different costs. The surrounding bookkeeping is slower,
which the recent caching commits keep chipping away at. Or the compiled program itself
differs from `main`'s, which does not improve on its own. Nothing currently distinguishes
them, and the attribution run described in the PR summary profiled the default path rather
than this one.

Suggested experiment, a couple of hours: profile PR-at-full-width against
main-at-full-width on one model, same widths, same machine. That separates "the new
machinery costs something" from "the compiled program is different" in a single run, and
it is the only open question behind the runtime premium with no answer yet.

Caveats on these numbers: CPU only, one model family, small to medium sizes, one machine,
five repetitions. They point at the experiment rather than making a production claim. The
GPU benchmark pending at this head is the evidence that matters for production scale.

hmgaudecker:
Thanks for the review. Status of each comment, all on this branch:

**C1, C4** — `95ef0586`. The subject-chunking entry now states that `n_subjects` is removed and that runtime executors cached per subject shape replace the prewarm route. The budgeted-solve entry states both unbudgeted caps: reduced axes keep 64, tiled output axes bootstrap under a cap derived from 64 × 512 clamped to [64, 512]. The same split is in the runtime reference.

**C2** — `b657068c`. The receipt keeps its content but the nine absolute paths and the account name are placeholders now; the provenance record carries the new digest and a sanitization note. No in-repo consumer parses the file.

**C3** — `a1c453d9`. `testpaths = ["tests"]`; bare `pytest` collects 12,418 tests with no errors, explicit paths still reach the cluster files.

**G1-d** — `292088f0`. The `sharded_states` docstring separates the discrete case from the continuous route and points at the model-build check.

**G5-b** — `36d8e6a5`. New `tests/solution/test_donation_two_active_regimes.py`: two acting regimes in every period, a donating counter solver and a non-donating twin. Published values are bit-identical, and the second regime is asserted to receive the freshly donated template. Under your exact mutation (the three in-loop `SolveInputMappings` replaced by `period_inputs`) it fails at both precisions with `RuntimeError: Array has been deleted`; one detail worth knowing is that the mutation does not change published values, because the continuation roll overwrites the stale entry before anything reads it, so the value twin alone would not have caught it. `SolveInputMappings` carries the one-sentence note. The module runs in about two seconds.

**G1-a** — implemented, landing next once the certificate corridors are re-pinned. `axis_widths` accepts both forms in one field: a bare integer broadcasts, a regime-keyed mapping pins only the regimes it names. Your two-regime example reproduces as a test (the cheap regime keeps its planner width). Resolution happens where cell chunking is actually decided, in the output-layout and workspace planning of backward induction, not only in the regime builder. Per-regime widths for simulation-only axes are refused at model build because simulation plans without a regime in hand.

**G6-a** — ran the separating experiment (CPU, fp64, your model at 200×200 and 600×600, interleaved arms, forced host/device split, per-executable timing, HLO dumps). Two results:

1. The third column is not like-for-like. `main` does not run whole-axis: `core_program.py` caps the action axis at 64 (`_INITIAL_TILE_WIDTH_CAP`), only the cell axis is dense; its HLO has a trip-9 `while` over the action axis at 600×600. Pinning every axis to its extent puts the PR in a shape `main` never ran, and a dense 600-wide action product is the slowest program of all (+148 % per core). The like-for-like pin is `{cell: extent, action_product: 64}`.

2. At like-for-like widths the PR compiles the same program as `main` (same two loops, same trip counts, byte-identical fusion shapes). The remaining premium decomposes into about 9–11 ms of host bookkeeping per solve, flat in model size (2.5 % at 600×600, 10 % at 200×200), plus a compiled residual: XLA:CPU assigns `outer_dimension_partitions` of 20 parallel chunks to the six hot fusions on `main` and 8 on the PR at identical shapes (+47 % per core at 200×200, +4.8 % at 600×600). The antecedent is an income `exp` that `main` hoists into a standalone fusion and the PR evaluates inside the traced core. Restoring that structure is in progress. The large numbers in your table are width policy: default `cell: 512` is 1.48x at 600×600 against 1.21x for a dense cell axis, so the unbudgeted cell cap is being re-measured as well.

Evidence (scripts, raw samples, HLO text) will be in the closure bundle; the GPU benchmark at the current head is being rerun too.

timmens (approved):

PR #441, opened 2026-09-05
REF Define every engine function once; decide the MSS upper envelope from the stored operands

### LINKED ISSUE #439: MSS upper envelope: four decision-safety defects on main, and the repairs that were parked in July
## Summary

`src/_lcm/egm/upper_envelope/mss.py` on `main` carries four decision-safety
defects in which a structural ordering — which branch owns a query, whether a
branch switch happened at all — is decided by a single rounded comparison.
Three of them were repaired in July 2026 on the `feat/nb-egm` line and then
deliberately parked; the repairs never reached `main`, and the rebuild that was
supposed to inherit them did not happen. The fourth was found independently by
an external audit of #437 and is not covered by those repairs.

All four are reproduced below against `main`'s current `mss.py`, which is
byte-identical to the file on `nbegm/expectation-before-inversion` at
`7aedc6e03ff55e9be97c870e12fe312d7c20e989`.

## Reachability

`MSSEnvelope` is opt-in. DC-EGM's envelope defaults to `ExactEnvelope`
(`src/_lcm/solution/dcegm.py:239`), and NB-EGM/NNBEGM do not route through
`upper_envelope` at all — they use `envelope_at_query`. So none of these fires
on a default path; they affect models that select `MSSEnvelope` explicitly.
That is the same reachability argument under which they were parked in July, and
it is why this is an issue rather than a release blocker.

## The four defects

Measurements below are from a direct dissection of `refine_envelope`'s
internals — `_evaluate_envelope`, `_crossing_blocks`, and the on-envelope
test — on each witness.

### 1. A branch crossing landing exactly on a query node is discarded

`mss.py:332`

```python
within = switches & (row.grid > prev_grid) & (row.grid < this_grid)
```

Both bounds are strict, so a crossing whose abscissa coincides with either
adjacent query node is dropped. The module's own contract is that a switch is
represented by two equal-abscissa rows carrying the left and right owner's
policy; at a node-aligned crossing only the left owner survives.

Witness — two segments crossing exactly at the query node `x = 10, V = 5`:

| branch | span | values | policy |
| --- | --- | --- | --- |
| A | `[9, 10]` | `[4.875, 5.0]` | 8 |
| B | `[9.5, 10.5]` | `[4.75, 5.25]` | 2 |

```
grid    [9.0,   9.5,    10.0, 10.5]
policy  [8.0,   8.0,    8.0,  2.0 ]
value   [4.875, 4.9375, 5.0,  5.25]
```

`policy = 2` is absent at `x = 10`. Everything upstream is correct — the
crossing is computed at `grid = 10.0`, `value = 5.0`, `policy_left = 8`,
`policy_right = 2`, and the dense on-envelope check returns `True`. It is
`row.grid > prev_grid` alone, with `prev_grid == 10.0`, that sets `valid =
False`.

The symmetric case is equally broken: if the evaluator resolves the tie at the
node in favour of the incoming branch, the switch is detected at that query and
`row.grid < this_grid` discards it instead.

A downstream right-side policy read then bridges a jump that should be there.

### 2. A crossing's published value is computed far from the crossing, and rounds below both branches

`_intersect_winners` anchors each line at its own **left endpoint** (`a_x0`,
`b_x0`) and extrapolates to the intersection. When the crossing is far from
those anchors the product rounds below both branch values, and the dense
on-envelope test (`mss.py:_on_envelope_atol`, an 8·eps band) then rejects a
genuine crossing.

Witness — two branches over `[0.014666654, 0.023388933]`, values scaled by
`2**30`:

```
crossing.grid       0.01666785
crossing.value     -2.642378     <- from the far-anchored line product
crossing_envelope  -2.25         <- the dense envelope there
on_envelope         False        <- genuine crossing rejected
```

The refined output contains no crossing row at all: `n_kept = 4`, exactly the
four input nodes, while the winner demonstrably changes from branch A
(`policy 0.01`) to branch B (`policy 0.000108`) across the interval.

### 3. A 1-ULP value ordering hides an observable branch switch

`_evaluate_envelope` decides ownership from one rounded comparison. When two
branches are one float32 ULP apart at a node, the switch is never seen.

Witness — over `[23630.662109375, 25235.28515625]`, values scaled by `2**20`:

| | value at left | value at right | policy |
| --- | --- | --- | --- |
| A | `-47684.4375` | `1054401.0` | `22626.18` |
| B | `-136030.45` | `1054401.125` | `20947.01` |

B is above A at the right endpoint by **exactly 1 float32 ULP**. The evaluator
returns `winner_segment = [0, 0, 0, 0]` — A everywhere — so no switch is
detected, no crossing is emitted, and the published policy at `x = right` is
A's. The value error is 1 ULP; the **policy** error is 8 %, and the read at the
last float below the node returns the outgoing branch extrapolated across it.

### 4. A node is anchored on its stored point below a covering chord

Same root cause as 3, on the anchoring side rather than the switch side. A
stored point that lies below a covering chord by less than the certification
margin is published as the envelope.

Witness: query node at `145.0667724609375` with stored value
`-2.5820427e-04`, covered by a chord whose exact value there is
`-2.5820422e-04`. The published read is the stored point — short of the
covering chord by ≈1.3 float32 ULP. The node is anchored on its own stored
value because an endpoint tie makes it representable, rather than either
reaching the covering chord or failing loudly.

## Reproduction

A standalone reproducer for defect 1 (stdlib + JAX only, loads `mss.py`
directly, exits 0 when the defect is observed):

```
python reproduce_mss_node_crossing.py --source-root /path/to/pylcm
```

Defects 2–4 are the three tests in `tests/solution/test_round18_mss_regressions.py`
as it existed at 27a79cf22bca0da104c66e088170fe2596a81dfc. Two notes for whoever
picks this up:

- All three fail against today's `mss.py`, and I verified by dissection that
  each fails **for its own defect**, not incidentally.
- One assertion in them is stale rather than diagnostic:
  `test_an_interior_crossing_cannot_publish_a_value_below_both_branches` locates
  the crossing with `assert duplicate.size == 1`, counting duplicated abscissae.
  Today's `refine_envelope` emits one row per input candidate, and the witness
  feeds `[left, right, left, right]`, so there are two duplicated abscissae
  before any crossing exists. That locator needs rewriting against the current
  output shape; the assertion it guards is still the right one.

## Why the repairs are not in the tree

Not a quality judgement — a scope decision whose follow-through did not happen.

The lineage exists and is reachable on GitHub:

- d6ce3e786d0531e246cf2056ab1c1c865061ed62 — round-17: certified point-vs-chord coverage + node-event crossings
- 27a79cf22bca0da104c66e088170fe2596a81dfc — round-18: certified anchoring, crossing value, missed switch
- 3eaeccac849a0892d30b720c3e367fb20948c9ce — round-19: exact-zero tie + endpoint-faithful crossing value
- a6e833d95cf16502a709420a127f397787df29c7 — round-20: interior-tie floor, faithful division, deferred ownership

On 2026-07-28, d5331ba8dd3328e94ed8b9b4f3902b423c300922 removed them from the
NB-EGM line on branch-hygiene grounds, in its own words: *"Everything that is
not specifically NB-EGM/NNBEGM belongs on feat/dcegm. These files carried #400
deltas that were never split across, so they are taken wholesale from feat/dcegm
rather than reconciled here: `egm/upper_envelope/mss.py` — #400's rounds-17-20
lineage retires in favour of feat/dcegm's version."* It also deleted the five
MSS test files, `test_round18/19/20_mss_regressions.py` among them.

The #400 review reinforced that routing, on two grounds: `mss.py` fires on no
default path, and the file was *"the exact code slated for from-scratch
replacement under round-26 R2 / the MSS certification-layer rebuild, which is
owned by the #390 worker"*. Its conclusion was that the MSS findings *"should
therefore be routed to that rebuild as acceptance criteria — reproduced
counterexamples the replacement must survive — not repaired in place"*.

What then happened:

- #390 merged to `main` on 2026-08-21 carrying `feat/dcegm`'s MSS — the
  pre-rounds-17-20 version, 631 lines.
- `feat/dcegm` no longer exists, locally or on the remote.
- The certification-layer rebuild that was to inherit the counterexamples has
  not landed: `main`'s `mss.py` is byte-identical to the reverted-to version and
  still exhibits all four defects above.
- None of the four repair commits is an ancestor of `main` or of any branch;
  they survive only as unreferenced objects.

So the counterexamples were parked against a rebuild, and the parking spot went
away with the branch.

## What would close this

Either is acceptable; they differ in cost, not in outcome.

1. **Land the certification-layer rebuild** the July triage assumed, with these
   four witnesses as acceptance criteria. Defect 1 is not covered by the parked
   repairs and has to be added to the set.
2. **Port rounds 17–20 forward.** The commits do not cherry-pick: `mss.py` was
   rewritten between 2026-07-20 and today (468 insertions / 880 deletions
   between `27a79cf2^` and `main`), and `git cherry-pick -n 27a79cf2` conflicts
   in `mss.py`. The repair would have to be re-derived against the current file.

Whichever route, defect 1 needs an explicit ordering decision that the current
code does not make: which owner the node emission carries when a crossing
coincides with a query abscissa, so that the left and right records land in
weakly-ascending order without duplicating the node's own row. That is a tie
rule to be stated, not inferred.

Found while auditing #437; `mss.py` is byte-identical at that PR's base and
head, and #437 neither routes a solver newly through MSS nor uses it as an
oracle, so it is not a blocker there.

hmgaudecker:
Closes #439.

Two changes travel together here: every engine function is defined once rather than per call, and the MSS upper envelope settles its structural orderings from the stored operands.

# Every engine function is defined once

Every function the engine defines is a module-level function or a frozen dataclass with a `__call__`. No function definition runs inside another function per model build, per solve, per simulate, or per trace.

pylcm's beartype claw decorates every function definition it sees, nested ones included, and beartype memoizes each decorated function object for the life of the process. A per-call definition was therefore pinned together with everything it closed over: grids, arrays, tracers, compiled kernels, whole models. A session or an xdist worker that built many models grew with every one of them.

Measured on a second build-and-solve of one small model, counting engine functions still reachable after `gc.collect()`:

| family | before | after |
| --- | --- | --- |
| grid search | 247 | 2 |
| NEGM | 520 | 22 |
| NNBEGM, finite | 613 | 19 |
| NNBEGM, adaptive | 655 | 13 |

Every survivor is in one of the five sources named below.

- Bodies a JAX transformation calls positionally are module-level functions carrying the keyword-only exemption marker, with their bound operands supplied through `functools.partial`. Mapped axes, operand order and dtypes are unchanged.
- Callables a `dags` DAG binds by parameter name keep their parameter names and publish a signature and a name.
- The call-convention adapters and the batched product map are frozen dataclasses carrying the wrapped function's name, docstring, attributes and signature, so `inspect.signature`, `dags`, `jax.named_call`, and the JAX transformations see the function they stand in for.
- The durable model fingerprint treats a class as a closed direct reference only when it is the object a shipped pylcm module binds at its qualified name, its implementation being sealed by the separately checked pylcm version; a class that merely claims a shipped module name is refused.

`tests/test_dropped_models_release_nested_functions.py` measures the property for each shipped solver family, with a positive control so it cannot pass vacuously.

## What still defines a function per call

- `regime_building/max_Q_over_a.py` and `regime_building/collective.py`: the candidate certificate pins the reducer builders' bodies verbatim, nested definitions included, so their shape changes with that certificate or not at all.
- `solution/negm.py`, `solution/nnbegm.py` and `regime_building/processing.py`: converted together with the execution work that rewrites the same call sites.

The probe test exempts those five by path.

## A debug simulation from a loaded solution writes its snapshot

The pickled result in a simulate snapshot drops the solution the simulation consumed, alongside the value arrays the snapshot already keeps in its HDF5 file. A solution read back from an archive holds lazy value-store entries and a cache lock that cannot be pickled, so a debug simulation from `load_solution()` previously died in the snapshot writer.

## The candidate certificate seals are checked at commit time

The certificate pins every certified source by digest in a generated inventory and a hand-maintained seal map. `tests/candidate_certificate/check_seals.py` verifies both against the tree in under a second and `--fix` reseals them; the `candidate-certificate-seals` pre-commit hook runs it on any change under `src/` or to the certificate files, so seal drift no longer surfaces as the only failure of a full CI run.

# The MSS upper envelope decides its orderings from the stored operands

`MSSEnvelope` settles three structural questions — which link owns a query, whether the owning branch changed, and where the change happened. The geometry posing them is fixed; which arithmetic settles each comparison is selectable, and under the default the answers come from the stored operands rather than from a single rounded comparison.

## Ownership is certified

A provisional owner is read off the values and then challenged: the exact integer comparator in `_lcm.egm.upper_envelope.certified_sign`, which the `envelope_at_query` path already uses, decides the sign of each bracketing link's value less the standing owner's. Links certified *level* with the owner are separated right-continuously — the link reaching strictly right of the query, then the steeper one, then the earlier stored link — so a node where two branches meet belongs to the branch that owns the interval above it. A switch whose margin is one representable step is therefore published at the node the geometry puts it at.

Which links compete is decided by their stored spans alone, never by whether reading one of them succeeded numerically, and a query whose owner the comparator leaves undecided publishes `NaN` rather than falling back to a rounded comparison.

## A value is read from its own chord's endpoints

Under the certified arithmetic each link is evaluated at a query by forming the exact rational through its two stored endpoints and rounding it once to the working format, without weighted floating products. The reading is the stored value at either endpoint exactly, and elsewhere it is the chord's own value rather than a line extrapolated from one far anchor. The published value and policy at a node always name one owner.

## A crossing is built from the pieces covering its interval

A branch entered at the right node is often represented there by the piece continuing *above* that node, whose line says nothing about the interval below it. The piece covering both of the interval's abscissae is the one the crossing is solved from, so a kink sits where the branches actually meet rather than where a continuation extrapolated backwards would meet them. Exact equality at the shared endpoint connects the selected piece to the node it hands over at; a trace that is missing, ambiguous, or disconnected from that node publishes `NaN` rather than a plausible abscissa.

The gap between the two selected pieces is signed only where both are supported — two traces sharing no support cannot meet, so the owner change across a gap in the candidate cloud is a jump rather than a kink, yields no event, and leaves each node the value its own owner reads there. Where they do meet, the root is solved from the stored operands and rounded up to the first representable state the incoming branch owns. The published crossing value is the higher of the two chords at the emitted abscissa, so an inserted kink never sits below both branches.

## A crossing on a node carries both owners

A crossing whose abscissa coincides with one of the two query nodes is published. The ordering rule is stated rather than inferred: that node's own row is already one of the two records the switch needs, and the emission contributes the other.

- Crossing at the left node: that node published the outgoing owner, so the incoming owner's record is inserted after it.
- Crossing at the right node: that node publishes the incoming owner, so the outgoing owner's record is inserted before it.

Either way the kink abscissa carries exactly two rows, outgoing owner first, and no row duplicates the node's own.

## The on-envelope band is gone

Whether a crossing lies on the envelope is settled by naming the owner at the crossing abscissa and requiring it to carry the same value there as the piece the crossing was built from — a shared branch label is not provenance on its own. Nothing about the emission depends on a declared tolerance or on the working precision, so the band constant, its floor, and the test that guarded its width are removed.

## The comparison arithmetic is selectable

`MSSEnvelope(arithmetic=...)` takes `"certified"`, the default, or `"ordinary"`. The geometry above is the same either way — piece selection, node ownership and handover location are decided identically — so both settle a knot at the root of the pieces covering it. Only the comparison changes.

- `"certified"` decides on the stored operands, as described above. It needs the installed exact-affine payload for the active backend, and a regime selecting it is refused during `Model(...)` when that payload is absent rather than falling back.
- `"ordinary"` compares two readings formed in the working format. A reading is a slope and then an affine step, two rounded operations rather than one correctly rounded value, so it carries no bound in representable steps: cancellation between a chord's endpoints, or a large common level under both chords, can move a reading by many steps and reverse an ordering the correctly rounded values would separate. Candidates whose readings coincide are separated by the declared tie order. It reaches no native kernel, so it is the route available where that payload is absent; a model's values, owners and crossings under this selection are the caller's to validate at every precision, backend and transformation in use.

The selection reaches both routes the envelope is read through: the full refined row, and, in a regime solved per exogenous asset node, the bracket located at one query per node. The bracket finder builds the row under the same arithmetic the regime selected, so an ordinary selection locates its bracket in the ordinary row and requests no exact-affine entry point, and a certified selection never settles a comparison in the working format on that route either.

## Scope and cost

`MSSEnvelope` is opt-in. DC-EGM's envelope defaults to `ExactEnvelope` and the NB-EGM family does not route through `upper_envelope`, so no default path changes.

The certified comparisons run over the dense `(n_query, n_link)` block, which is what the method costs to be right. Measured on one CPU host over a synthetic candidate cloud, the ordinary arithmetic's warm per-call time is roughly seven to thirteen times below the certified one across 16/64/128 candidates at both precisions, with `n_kept` identical in every cell. Compile time moves the other way by a factor under two, since the ordinary path stages a lexicographic reduction where the certified one makes a single native call.

That measurement is one host, one witness family and one backend, and the host was not isolated from concurrent work — not a portable characterization.

The published node value is the correctly rounded value of the owning chord, not an upper bound of it; a reading can still sit one representable step below the exact envelope.

## Tests

`tests/solution/test_mss_decision_safety.py` fixes one witness per ordering. `tests/solution/test_mss_arithmetic_modes.py` pins what the two arithmetics share and where they part, including that the ordinary one reaches no exact kernel when all four native primitives are patched to raise. `tests/solution/test_mss_bracket_arithmetic.py` pins that the asset-row bracket is the pair the same arithmetic's full row holds, eagerly, under `jit` and batched, that the ordinary route completes without any exact-affine entry point while the certified one reaches it, and that an asset-row model solves under either arithmetic and matches a dense-grid brute-force oracle.

https://claude.ai/code/session_01G9giPUNidKrTio57pKUxTG


src/_lcm/utils/dispatchers.py
```diff
@@ -286,45 +286,95 @@ def _base_productmap_batched(
...
+        "FunctionWithArrayReturn",
+        _ProductMapBatched(
+            func=func,
+            product_axes=product_axes,
+            batch_sizes=MappingProxyType(dict(batch_sizes)),
+        ),
+    )
+
+
+@dataclass(frozen=True, eq=False)
```
timmens, line 299: **[P3] Preserve product-map callable metadata**

`_ProductMapBatched` has no `__name__`, so the outer `_KeywordOnlyAdapter` cannot copy one. This makes the following standard JAX instrumentation fail before executing:

```python
import jax
import jax.numpy as jnp

from _lcm.utils.dispatchers import productmap

def utility(consumption):
    return consumption

mapped = productmap(
    func=utility,
    variables=("consumption",),
    batch_sizes={"consumption": 0},
)
jax.named_call(mapped)(consumption=jnp.array([1.0, 2.0]))
```

At this head it raises `AttributeError: '_KeywordOnlyAdapter' object has no attribute '__name__'`; at the PR base it returns `[1.0, 2.0]`. Please publish a stable name on this callable and add a focused regression test. No current repository path uses `named_call` or `named_scope`, so this is a low-priority compatibility issue.

src/_lcm/solution/fingerprint.py
```diff
@@ -2047,6 +2035,22 @@ def _is_closed_terminal_reference(  # noqa: C901, PLR0911, PLR0912
...
+    Builtin and versioned numeric-library classes are sealed by their library
+    versions, and so is every class defined in a shipped pylcm module: its
+    implementation is fixed by the separately checked pylcm version, so its name
+    is its identity. Any other class binds behaviour the digest cannot see.
+    """
+    return (
+        _contains_identity(value=value, candidates=_BUILTIN_TYPE_OBJECTS)
+        or _contains_identity(value=value, candidates=_TRUSTED_DIRECT_TYPE_OBJECTS)
+        or _is_versioned_numeric_library_type(value)
+        or _is_shipped_pylcm_module_name(value.__module__)
```
timmens: **[P3] Fingerprint shipped classes by exact identity**

`_is_closed_direct_type` accepts any class whose writable `__module__` claims the `lcm` or `_lcm` namespace. Two behaviorally different classes can therefore receive the same semantic fingerprint:

```python
from _lcm.solution.fingerprint import _semantic_fingerprint

def make_utility(multiplier):
    class Scale:
        def __new__(cls, value):
            return multiplier * value

    Scale.__module__ = "_lcm.spoofed"
    Scale.__qualname__ = "Scale"

    def utility(value):
        return Scale(value)

    return utility

one = make_utility(1)
two = make_utility(2)
assert one(3) == 3 and two(3) == 6
assert _semantic_fingerprint(one) == _semantic_fingerprint(two)
```

The collision also propagates to complete model fingerprints, so a restored solution could pass compatibility checks and reuse stale values or policies. Reaching this path requires a class to claim a reserved namespace and a restored solution under the exact PyLCM version, which makes it a low-probability persistence edge case. Please compare exact class identity against captured shipped class objects, consistent with the nearby identity registries.

timmens (approved):
Approved. Two P3, non-blocking follow-ups are attached. I found no blocking issues.

src/_lcm/utils/dispatchers.py
hmgaudecker: Done. `_ProductMapBatched` publishes the `__name__` and `__qualname__` of the function it evaluates, the way `functools.wraps` would, so the keyword-only adapter copies them on and `jax.named_call` accepts the product map. `tests/test_dispatchers.py` covers both the published name and the instrumentation call.

src/_lcm/solution/fingerprint.py
hmgaudecker: Done. A class counts as a closed direct reference only when the module named by its `__module__` is a real shipped `lcm`/`_lcm` module (its spec origin resolves under the installed package roots) and walking its `__qualname__` from that module reaches the identical object. A class that merely claims a shipped module name, whether in a closure or with a spoofed qualified name, is refused as a direct class dependency, while a genuinely shipped class is still fingerprinted by identity. `tests/solution/test_solution_fingerprint.py` covers the two spoofed forms and the shipped one.

src/_lcm/egm/upper_envelope/mss.py
```diff
@@ -257,14 +332,635 @@ def refine_envelope(
...
+            a_x1=a_x1,
+            a_v0=a_v0,
+            a_v1=a_v1,
+            b_x0=b_x0,
+            b_x1=b_x1,
+            b_v0=b_v0,
+            b_v1=b_v1,
+        )
+
+    def gap(at: FloatND) -> FloatND:
```
timmens: **[P2] Hoist the MSS callbacks out of the per-trace path**

`_affine_handover` creates `gap` on every ordinary MSS trace, and `_interval_piece` creates `resolve_overlapping`. The package-wide beartype hook memoizes each decorated function for the process lifetime. In fresh processes, repeated real model build-solve-drop cycles left six additional nested functions alive per ordinary MSS model and four per certified model after `gc.collect()`. Their closures contain `DynamicJaxprTracer` objects and `_Links`. Repeated solves of one unchanged model stabilized after its first trace, while rebuilt or retraced MSS models keep accumulating survivors.

Please hoist `gap` and both `jax.lax.cond` branches to module-level helpers. The affine helper can take both lines explicitly. For `_interval_piece`, pass the traced arrays and `_Links` through the operand and select global ordinary or certified callbacks before `lax.cond`, preserving the static arithmetic choice without creating a `partial` or another per-trace callable. Please add ordinary and certified MSS cases to `test_dropped_models_release_nested_functions.py`.

timmens (changes requested):
Requesting changes for the blocking MSS lifetime regression described inline.

**[P3] Remove the obsolete MSS crossing description**

`_envelope_publishes_crossings` still says MSS suppresses a switch that lands exactly on a candidate abscissa. The new MSS implementation publishes and tests node-aligned switches. Please remove that obsolete clause while retaining the separate interior-only-branch limitation. This documentation note is non-blocking.

src/_lcm/egm/upper_envelope/mss.py
hmgaudecker: Done. The ordinary gap is a module-level function taking both lines, and the two `lax.cond` branches of the interval-piece resolver are module-level callbacks reading their traced inputs from a NamedTuple operand; the static arithmetic selects the certified or ordinary resolver before the cond instead of being closed over. `tests/test_dropped_models_release_nested_functions.py` gains an MSS family per arithmetic on the DC-EGM twin with a coarse savings grid; both failed against the per-trace definitions (8 and 12 extra survivors) and pass now.

timmens (approved):
Approved. The blocking MSS callback-lifetime regression and the non-blocking crossing-description issue are resolved. The focused lifetime, fp64 MSS, fp32 MSS, and certificate-seal checks pass, and the full CI matrix is green.

PR #453, opened 2026-09-18
ENH Public Model.validate_initial_conditions and Model.initial_conditions_feasibility

### LINKED ISSUE #446: Expose public Model initial-condition validation and per-individual feasibility without solving
## Request

Expose initial-condition validation and per-individual feasibility through public `Model` methods, without requiring a solve or simulation. This is needed by lcm-zoo's invented-initial-observation sampler; downstream code must not construct calls to private pylcm feasibility DAGs.

```python
model.validate_initial_conditions(*, initial_conditions, params) -> None
model.initial_conditions_feasibility(*, initial_conditions, params) -> boolean_array
```

Both accept the same user-facing input forms as `Model.simulate`, including a DataFrame with ages, regime names and categorical labels, plus ordinary user parameters. No internal regimes, regime-ID maps, flattened parameters or periods should be supplied by callers.

## Contract

- `validate_initial_conditions` raises `InvalidInitialConditionsError` for invalid structure, states or economic feasibility, retaining useful existing diagnostics.
- `initial_conditions_feasibility` first validates input structure, ages, active regimes and categorical codes, then returns a one-dimensional boolean mask in original individual order. `False` means no jointly admissible action combination; malformed inputs and unsupported checks raise, rather than being silently classified as infeasible.
- pylcm owns age-to-period conversion, canonical runtime scalar types, fixed/runtime parameter binding, categorical conversion, regime grouping and action dispatch.
- Use the same core implementation as simulation validation. Do not introduce a second feasibility implementation. No solving, simulation, or caller-input mutation.
- Define feasibility on the model's declared action-grid support, with batching/resource behavior explicit; do not claim continuous optimization or silently drop difficult rows.
- Check state-only constraints even when there are no action variables. Unsupported collective or age-specialized checks must fail explicitly, never return an all-true mask.

## Existing implementation to reuse

In `src/_lcm/simulation/initial_conditions.py`: `validate_initial_conditions`, `canonicalize_initial_conditions`, the structural/discrete-state checks, `_collect_feasibility_errors`, `_check_regime_feasibility`, and `_batched_feasibility_check`. Reuse the DataFrame conversion and parameter processing already used by `Model.simulate`.

`_batched_feasibility_check` already produces the per-subject mask; `_check_regime_feasibility` currently converts it into diagnostic text. Factor the mask-producing path so both public methods and simulation consume it, while leaving internal helpers private. The current early return for an empty `action_names` list also needs attention: it bypasses state-only constraints.

Source inspected at pylcm commit `e0affa16f5995a69e57e596713e61b3b0d6f01e0`; downstream assessment used the environment's pinned revision, not necessarily this checkout revision.

## Motivation and cheap acceptance tests

The lcm-zoo private adapter passed a Python integer `period` into paper functions requiring a scalar JAX `int32` Period (Attanasio, Barczyk, Cocco). Chan's treatment case exposed a Python-int fixed-parameter/JAX-array mismatch. These are downstream-adapter findings, not a claim that the existing public simulation route has the same defects.

Add tiny unsolved fixtures covering:

- DataFrame and mapping inputs; multiple regimes and nonconsecutive model ages; stable row order.
- Period-annotated feasibility ancestors under runtime type checking, and fixed versus runtime scalar parameters.
- A mask containing both feasible and infeasible people, agreeing with an independent exhaustive scalar action oracle.
- Joint feasibility, including constraints each individually satisfiable but jointly impossible.
- No-action state constraints; runtime grids and carried states.
- Malformed ages, labels, shapes and state schemas raise; unsupported checks never count as valid.
- Validator success iff the mask is all true for structurally valid supported inputs; informative rejection otherwise.
- Solve/simulate are not called; inputs unchanged; simulation uses this same core.

## Downstream sequencing

Related: https://github.com/OpenSourceEconomics/lcm-zoo/pull/162

The maintainer intends to merge lcm-zoo #162 before this feature exists. Its shared sampler will deliberately raise `NotImplementedError`, with the private adapter removed. **The five paper PRs must not merge until this upstream feature lands, is adopted through the public interface, and is verified on their paper cases.** No private fallback or paper-local scalar workaround should bridge the gap.

Grid-based proposal generation stays in lcm-zoo; this request is for pylcm-owned validation/feasibility, not a new population generator.

hmgaudecker:
Closes #446.

## What

Two keyword-only public methods on `Model` that check a proposed population without solving or simulating:

- `Model.validate_initial_conditions(*, initial_conditions, params) -> None` raises `InvalidInitialConditionsError` with the same diagnostics `simulate` would produce, or returns.
- `Model.initial_conditions_feasibility(*, initial_conditions, params) -> Bool1D` returns a one-dimensional boolean mask in caller order, `True` exactly for the subjects that admit at least one jointly feasible action combination on the declared action grids.

Both accept the `simulate` input forms (a mapping of arrays or a DataFrame with regime names, ages and categorical labels) and ordinary user parameters. Malformed inputs raise from both. The checks run eagerly on one device, without device-memory admission or subject padding, and take no `log_level`. lcm-zoo's initial-observation sampler is the first consumer.

## Behaviour changes in `simulate`

- Subjects starting in a regime **without actions** are now checked against that regime's state-only constraints. Previously such rows were waved through unchecked; populations that simulated from a violating row now raise `InvalidInitialConditionsError`.
- A constraint that depends on an **age-specialized function** while subjects start away from the regime's representative age now raises `UnsupportedOperationError` instead of `InvalidInitialConditionsError`, and `simulate` never downgrades it to a warning.

## How

- One mask-producing core (`initial_conditions_feasibility_mask`, `_regime_feasibility_mask`) shared by `simulate`'s preflight and the two public methods; `_check_regime_feasibility` is now a message wrapper that explains from the mask instead of recomputing it, so the certified `_gather_feasibility_inputs` program counts are unchanged.
- Every initial-condition test moved into `tests/simulation/initial_conditions/`, chunked by validation stage, with a pure-Python scalar oracle (`_oracle.py`) that enumerates the action grid for each subject and calls the user's constraint functions with scalars.
- The candidate certificate is re-anchored with `repin_corridors.py` (five callables and the module surface of `initial_conditions.py`, the module surface of `model.py`) and resealed; `verify.py` and `verify.py --self-test` exit 0. The regime-slot read ledger rows for the feasibility producer are retargeted.
- The frozen CI workload manifest is re-keyed for the moved files (renamed modules keep their observed weights; split and new modules are listed as unweighted); `generate_ci_workloads.py --check` passes and the manifest is phantom-free and complete at every commit.
- Docs: API reference, user guide section "Checking initial conditions without simulating", agent guide, `CHANGES.md`, and a `CONTEXT.md` glossary establishing **subject** as the canonical term for one simulated unit.

## Coverage instrumentation (timmens' note on #440)

`codecov/patch` read 21.53% on #440 because 34 test modules exercise the execution and sharding code in child Python processes and nothing measured those children. `pyproject.toml` now sets `[tool.coverage.run] patch = ["subprocess"]` and `sigterm = true`: coverage hands each child the parent's configuration through the environment, the `.pth` hook it installs starts measurement at interpreter startup, and pytest-cov combines the children's data with the parent's. Measured on two spawning modules with `--cov=src`: 14523 covered source lines before, 19660 after, same 40 tests. Commit `2b49c30e`.

## CI status

Every job on `main` and on every PR currently fails in the setup step "Probe or reinstall the installed native payload" since jax 0.11.2 reached PyPI on 2026-09-17: the isolated build of the native kernel picks up the new jax while the lock holds 0.11.1. #454 fixes that; this PR's CI can only go green once #454 is in `main` and this branch is rebased or rerun.

## Verification

Tree identical to the last Marvin-verified head (`65fecdc6` on the pre-squash #440 tip, which is byte-identical to `main`).

Marvin, `intelsr_short`, every Linux job of `cpu.yml` replayed per precision (general shards at 16 workers, slow-solution shards at 8, everything else as in CI):

| leg | steps | tests | failures | errors | skipped |
| --- | --- | --- | --- | --- | --- |
| fp64 | 36 | 13155 | 0 | 0 | 314 |
| fp32 | 34 | 12768 | 0 | 0 | 586 |

Locally (Python 3.14.7, `tests-cpu`): `prek run --all-files` exit 0; certificate, re-pin and ledger batteries 452 passed; `tests/ci` plus the initial-conditions and related batteries 695 passed at fp64 and 239 at fp32.

https://claude.ai/code/session_01WVwaHeu3JsLpBjn3XbE6zu


src/lcm/model.py
```diff
@@ -2574,6 +2580,113 @@ def simulate(  # noqa: C901, PLR0912, PLR0915
             )
         return result
 
+    @beartype(conf=PARAMS_CONF)
+    def validate_initial_conditions(
```
timmens, line 2584: **Blocking.** A household population can pass this check and then be rejected by `simulate`.

In models with collective (household) regimes, every subject starting in a household has to declare which member it is, because that decides where its row goes when the household dissolves. That check, `_initial_own_stakeholder`, lives inside `simulate.py` and runs downstream of `validate_simulation_inputs`. Both new methods stop before it.

Verified against this head:

```
model.validate_initial_conditions(...)     -> returned, population reported VALID
model.initial_conditions_feasibility(...)  -> [True True True]
model.simulate(...)  on the identical input:
  InvalidInitialConditionsError: Subjects starting in the collective regime(s)
  ['married'] need an `own_stakeholder` entry in `initial_conditions`, naming the
  role each one occupies (one of {'f': 0, 'm': 1}). ...
```

Reproducer:

```python
model = Model(
    regimes=_make_dissolution_regimes(),
    ages=AgeGrid(start=0, stop=3, step="Y"),
    regime_id_class=DissolutionRegimeId,
)
initial_conditions = {
    "wage": jnp.asarray((1.0, 2.0, 3.0)),
    "age": jnp.zeros(3),
    "regime_id": jnp.full(3, model.regime_names_to_ids["married"], dtype=jnp.int32),
}
model.validate_initial_conditions(
    initial_conditions=initial_conditions, params=_DISSOLUTION_PARAMS
)  # returns
model.simulate(
    params=_DISSOLUTION_PARAMS, initial_conditions=initial_conditions,
    log_level="off", seed=0,
)  # InvalidInitialConditionsError
```

(`_make_dissolution_regimes`, `DissolutionRegimeId` and `_DISSOLUTION_PARAMS` from `tests/regime_building/test_collective_regime_simulate.py`.)

This is the false negative #446 asked to rule out, and lcm-zoo's sampler is exactly the caller that would hit it. Two honest ways forward: move the role check into the shared preflight so both paths see it, or narrow the promise this docstring and the reference page make, which currently reads "the same diagnostics `simulate(log_level=\"debug\")` reports". Right now the documentation describes the first and the code does the second.

Whichever you pick, the missing test noted in the review body is what keeps it true.

tests/candidate_certificate/direct_flow.py
```diff
@@ -3295,20 +3295,20 @@ def _simulation_dispatch_corridor_errors(*, tree: ast.Module, source: str) -> li
...
             "_discrete_initial_specs": "6e837939569bad3b27093ab4a14ce38ad16756f86910f37127432d07291195a6",
             "_pack_initial_summary": "504e89911be3016190066e13c8acebdb890328a4a7d40c4f9e8afd6a6f0b748a",
             "_read_initial_cohorts": "c2de600ad5640aabae10671db6123ec1dcd7dac80b9bf50a8a21e022d30838ad",
-            "validate_initial_conditions": "706a297af1125941763bf594511ee343f31dc8a725d38aa1614048c232a8d9c8",
-            "_collect_feasibility_errors": "ccb25705a003d95b1cd10f406df14ef43a809508c0f0028b6e1a13ddf7022b02",
+            "validate_initial_conditions": "72d30bb8911db80e74da07e9d2d1daf1241f76addee51cca49acea63c14adcae",
+            "_collect_feasibility_errors": "44069ec69fdb5195f039c09ee4354c8667a217666ff0418cea4e8753b1f93c9f",
             "_age_specialized_feasibility_message": "1eefe701cada0b668ab2eab2e404120b47d2c65ff390afeb1e726ae420fde174",
-            "_check_regime_feasibility": "697f56ea2750627de988226022db4f519209a71b434bd9bf780dcab3682d5dff",
+            "_check_regime_feasibility": "6d08d0a147ab6ec009f5be36d224fb436537f8ba426faac3f3c772e4da7e7e0a",
```
timmens, line 3308: **Blocking.** This fingerprint, and the two matching ones at lines 6320 and 6738, now guard a function that no longer contains what they were written to guard.

Three corridor contracts pin `_check_regime_feasibility`:

- `_SIMULATION_ADAPTER_CONTRACTS` here, memory-profiled host transport
- `_UNIFORM_PROCESS_CONTRACTS` at 6320, `process_grid_resolver` admission
- `_ACTION_GRID_CONTRACTS` at 6738, `PreflightActionGrids.resolve` admission

Every one of those concerns lived in that function's body. This PR moved the `action_grid_resolver.resolve(...)` call, the `state_action_space(process_grid_resolver=...)` call and the `run_simulation_operation(memory=...)` transport into the new `_regime_feasibility_mask`. What remains under the pin is a function that formats an error message.

`_regime_feasibility_mask` is fingerprinted in none of the three. Because the checks are name-keyed, `_exact_callable_errors` plus a module-surface hash, the surface hash noticed that names changed and was resealed, while the guarded behaviour quietly left the guarded function. The certificate reports green covering strictly less than it did on `main`.

Fix: add `_regime_feasibility_mask` to the three dicts and reseal, and consider whether `_check_regime_feasibility` still belongs in any of them now that it holds no transport.

Worth a separate thought, out of scope here: name-keyed pinning drops coverage silently on any extract-function refactor. Pinning by call site, "whatever function contains `action_grid_resolver.resolve` must be pinned", would have failed loudly instead.

tests/simulation/initial_conditions/test_public_api.py
```diff
@@ -0,0 +1,546 @@
...
+                    "age": [0.0],
+                    "wealth": [10.0],
+                    "health": ["immortal"],
+                }
+            ),
+            params={"discount_factor": 0.95},
+        )
+
+
+def test_runtime_supplied_grid_points_are_used() -> None:
```
timmens, line 306: The docstring overclaims. Runtime-supplied grid points do not bound the feasible set here.

The mask `[False, True]` falls out of the consumption action grid's minimum of 0.5 against the raw wealth values 0.3 and 5.0, under `consumption <= wealth`. The wealth grid never enters the verdict, because feasibility reads the supplied wealth value directly. The same input with any other wealth grid gives the same mask.

The test does guard something real, that runtime params reach `state_action_space` so an `IrregSpacedGrid` declared with runtime points can be built at all, which is what the implementation comment in `_regime_feasibility_mask` warns about. So the fix is the docstring rather than the test.

tests/simulation/initial_conditions/test_feasibility_mask.py
```diff
@@ -0,0 +1,348 @@
...
+                "regime_id": jnp.array([_dead, _dead]),
+            },
+            regimes=model._regimes,
+            regime_names_to_ids=model.regime_names_to_ids,
+            flat_params=flat_params,
+            ages=model.ages,
+        )
+
+
+def test_action_free_regime_state_only_constraint_accepts_satisfying_subjects() -> None:
```
timmens, line 228: This covers the PR's headline behaviour change and asserts only that nothing was raised.

Per the repo's own standard on concrete-value assertions, worth calling `initial_conditions_feasibility` on the same population and asserting an explicit all-`True` mask. As written, a regression that made action-free regimes silently feasible again would not be caught here; only the sibling "rejects" test would fail.

timmens (changes requested):
## Request changes

Two blocking items, both small to act on. Everything else is requested but not blocking.

Reviewed at `238c70a978f47e771c2c53e75735aabd6904cfa6` against merge base `2160f23b1`.

Summary of what I checked and what held up:

- `test_public_api.py` and `test_feasibility_mask.py` at this head: **45 passed**.
- `generate_ci_workloads.py --check` exits 0. The manifest re-key is clean: no phantom entries, no omissions, the nine renamed modules carried their exact recorded weights, the six new modules are unweighted as the description says.
- The test relocation is faithful. All 37 moved tests are byte-identical apart from import paths, the `_make_foo` to `make_foo` rename and one cosmetic docstring rewording. No assertion loosened, nothing dropped or skipped.
- Index ordering between `subject_indices` and `idx_arr` is consistent in all three construction branches of `_regime_feasibility_mask`, so the public path's scatter is correct. The action-free path activates machinery that already existed on `main` behind an unreachable early return. Age-specialized detection is equivalent with and without cohort metadata, so dropping the serial retry loses nothing. Public-path parity with `simulate` holds: same DataFrame conversion, same canonicalization, `process_grid_resolver` `None` in both without entry-allocation accounting.
- Docs, including the new `CONTEXT.md`, read in full against the code and accurate, with the one exception in the first inline comment.

Two further requests that have no natural inline anchor:

### Add the test that would have caught the first comment

`test_validate_raises_exactly_when_mask_has_a_false` covers the all-feasible equivalence. Nothing covers whether the new methods and `simulate` agree on which populations are acceptable. One case where `simulate` rejects a population and the validator is asserted to reject it too would have caught the collective gap below, and would keep the two from drifting apart as checks are added.

### Both methods are locked to a single eager pass, by documentation

The docstrings state that the checks run "eagerly on the default device, without device-memory admission or subject padding". Sensible today, and it is now part of the published contract, so relaxing it later becomes a contract change rather than an implementation detail. Is that deliberate, given that the sampler may propose very large populations?

### Bookkeeping

- `tests/regime_slot_read_ledger.csv` refreshes line numbers for the two rows this PR moves while neighbouring rows still cite a much older layout (`_per_constraint_feasibility` at 1026, actually near 1960). Either refresh the initial-conditions block as a whole or drop line numbers from the justification text.
- `test_dataframe_conversion.py` imports two private helpers back out of `tests/test_pandas_utils.py`, so the new directory is not quite self-contained.
- Pre-existing and not caused by this PR, but the file is being touched anyway: `test_feasibility_mask.py:114` `test_irreg_spaced_grid_with_runtime_points` omits `age` and uses a bare `pytest.raises`, so it is caught by the structural "age must be provided" check and never reaches the runtime-grid path it is named for. I confirmed this: without `age` the error is `'age' must be provided in initial_states ...`; with `age` supplied it becomes `All actions are infeasible for 1 subject(s) in regime 'working_life'.` One-line fix.

hmgaudecker:
Thanks, all nine points addressed at `e641c783`.

**Blocking**

- **Collective start without `own_stakeholder`.** Took the first route: the role check now lives in the shared preflight. `_initial_own_stakeholder` and `_regimes_reachable_from` moved verbatim from `simulate.py` into `initial_conditions.py` (their callable digests are unchanged, only the owner moved); `validate_initial_conditions` calls it after the feasibility diagnostics, exactly where `simulate` asks, and `initial_conditions_feasibility_mask` treats it as a structural precondition. `simulate` keeps its own call for the role array, so its body and `validate_simulation_inputs` are byte-identical. Your reproducer now raises the same message from both methods.
- **Certificate.** `_regime_feasibility_mask` is pinned in all three corridors. `_check_regime_feasibility` stays pinned: it remains the sole route by which the preflight's `process_grid_resolver`/`action_grid_resolver` reach the producer, and it still carries the profiled `_per_constraint_feasibility` transport (`memory=`). Re-pinned with `repin_corridors.py`, resealed; `verify.py` and `--self-test` exit 0. Agreed on call-site pinning being the robust fix; happy to open an issue for it.

**Requested**

- **Parity test.** `test_collective_start_without_roles_is_rejected_like_simulate` (parametrized over both methods) runs `simulate` and the method on your population and asserts both raise with identical messages. It was red before the move and is green after.
- **Eager single pass.** Deliberate for now, but you are right that it should not be a promise. The docstrings, reference page, agent guide and `CHANGES.md` now say the verdict does not depend on how the check is executed and that it *at present* runs in one eager pass without admission or padding, so admitting large populations later is an implementation change.
- **Runtime-grid docstring** (`test_public_api.py`) rewritten as you describe: the test guards that runtime points reach `state_action_space`; the verdict comes from the consumption grid's minimum.
- **Action-free "accepts" test** now also asserts the explicit `[True, True]` mask.
- **Ledger** line numbers of the whole `initial_conditions.py` block refreshed (plus the import line).
- **Self-contained directory.** The heterogeneous-health fixture moved into `tests/simulation/initial_conditions/_models.py` (`HetRegimeId`, `make_heterogeneous_health_model`); `tests/test_pandas_utils.py` imports it from there. The one remaining cross-import is the dissolution model for the new parity test, which already exists twice in `tests/regime_building/`; a third copy seemed worse than the import.
- **`test_irreg_spaced_grid_with_runtime_points`** supplies `age` and matches on `infeasible for 1 subject`, so it reaches the runtime-grid path it is named for.

**Evidence** (Python 3.14.7, `tests-cpu`): `prek run --all-files` exit 0; `generate_ci_workloads.py --check` exit 0 (no test files added or moved); fp64 battery (initial conditions, pandas utils, collective simulate, guards, age-specialized solve, admission, certificate, ledger, `tests/ci`): 1141 passed, 16 skipped, 0 failures; fp32 battery (initial conditions, collective simulate, pandas utils): 287 passed, 0 failures. CI at the previous head `238c70a9` was 53 passes with only the GPU benchmark comment pending; CI for the new head is running.

src/lcm/model.py
hmgaudecker: Fixed in `358b570c`, taking the first route. `_initial_own_stakeholder` (and its helper `_regimes_reachable_from`) moved verbatim from `simulate.py` into `initial_conditions.py`. `validate_initial_conditions` calls it right after the feasibility diagnostics, which is where `simulate` asks, so the order of messages matches; `initial_conditions_feasibility_mask` treats it as a structural precondition and raises before computing anything. `simulate` keeps its own call because it needs the role array, so its body and `validate_simulation_inputs` are unchanged. Your reproducer now raises the same `InvalidInitialConditionsError` text from all three entry points, and `test_collective_start_without_roles_is_rejected_like_simulate` (parametrized over both methods) pins that: red before the move, green after. Docstrings, reference page, user guide and `CHANGES.md` now name the role declaration among the checks.

tests/candidate_certificate/direct_flow.py
hmgaudecker: Fixed in `358b570c`: `_regime_feasibility_mask` is now pinned in `_SIMULATION_ADAPTER_CONTRACTS`, `_UNIFORM_PROCESS_CONTRACTS` and `_ACTION_GRID_CONTRACTS`. I kept `_check_regime_feasibility` in all three on purpose: it is still the only route by which the preflight's `process_grid_resolver` and `action_grid_resolver` reach the producer, and it still owns the profiled `_per_constraint_feasibility` transport (`memory=`), so dropping it would again shrink what the certificate covers. The moved `_initial_own_stakeholder` is re-pinned under its new owner in `_COMBINED_INPUT_CONTRACTS` with an unchanged digest. Re-pinned with `repin_corridors.py`, resealed; `verify.py` and `verify.py --self-test` exit 0. Agreed that name-keyed pinning fails silently on extract-function refactors; I would open a follow-up issue for call-site pinning if you want one.

tests/simulation/initial_conditions/test_public_api.py
hmgaudecker: Docstring fixed in `358b570c`; the test is unchanged. It now says the test guards that runtime-supplied points reach `state_action_space` so an `IrregSpacedGrid` declared with runtime points can be built at all, and that the verdict follows from the consumption grid's minimum against the supplied wealth values, not from the wealth grid.

tests/simulation/initial_conditions/test_feasibility_mask.py
hmgaudecker: Done in `e641c783`: the test now also calls `initial_conditions_feasibility_mask` on the same population and asserts `mask.tolist() == [True, True]`, so a regression that silently marks action-free regimes feasible again would fail here as well as in the sibling "rejects" test.

timmens (approved):

PR #456, opened 2026-09-19
PERF Add solve diagnostics, bounded planning and faithful replay

hmgaudecker:
This change makes expensive solve phases and individual compiled cores measurable, bounds optional workspace search, and preserves the transfer context needed to replay a captured period on its recorded device layout.

- Add public-solve phase records and per-wave compilation counters.
- Compile donation-free fallbacks only after the donating primary is admitted, retaining conservative admission.
- Add optional bounded width search; exhaustive search remains the default.
- Capture and validate device placements and transfer context during recorded-layout replay.
- Preserve declared lottery-axis order across Python processes.
- Reuse model-owned NB-EGM budget-affinity and continuation-constancy derivative programs. Every scheduled validation receives current parameters; successful verdicts are never cached. Programs and underlying model functions are released with their owning check, and failed abstract specializations are cleared.

The general GridSearch action-width-1 default was tested and reverted. Its clean benchmark slowed Precautionary solve by about 5.5× and Iskhakov GridSearch by about 10.9× compared with the preceding commit; reduced ACA also slowed. General solver defaults therefore retain planned widths. Explicit width 1 remains available for configurations where measurements support it.

For the complete ACA model with preference-state sharding across three A40s, a paired experiment reduced cold solve time from 125.9 to 63.9 minutes and warm time from 88.2 to 59.6 minutes. All 182 output-array hashes and schemas match across widths, and both arms match their own cold/warm values. This does not establish universal gains: a separately measured assets-sharded core was slightly slower at width 1. The mechanism behind the reversal remains an experimental question.

The eight-A40 assets-sharded full solve took 35.6 minutes cold and 23.0 minutes warm. All 182 arrays match the three-GPU default exactly after named-axis alignment. Both device count and shard axis change in this comparison; eight-GPU width 1 was not measured.

Validation includes 51 focused NB-EGM tests per precision, the original model-lifetime regression, unchanged planner/replay regressions, strict four-device numerical controls, and 406 certificate mutation controls. All final CI workflows and the benchmark pass at fd2cf21; automatic benchmarks are active. The final benchmark restores the general-model timing pattern after rollback.

Native ACA validation on the corrected source passed in fp32 and fp64: 72 scheduled check calls across initial, repeated, changed and restored parameter draws, totaling 576 accepted calls. Fingerprints verify parameter changes/restoration, cache-size vectors remain stable, and warm draws add no lowering or backend compilation. Warm validation still takes roughly 17–20 minutes per draw; the remaining tracing and runtime need attribution. Accepted calls do not imply 72 distinct successful differentiations.

The replay-context repair passed external closure review. A separate original ACA placement failure reproduces identically on baseline and measured source 0cbcc088; it is inherited and its matched evidence is retained. Replay evidence does not establish full-production memory admission, and cross-width parity does not establish absolute fp32 economic accuracy. Production M3 captures and control measurements remain pending.


src/lcm/model.py
```diff
@@ -893,24 +947,32 @@ def solve(
...
+                retention=retention,
+                max_compilation_workers=max_compilation_workers,
+                log_path=log_path,
+                log_keep_n_latest=log_keep_n_latest,
+                process_grid_resolver=None,
+                call_id=call_id,
+            )
+            solved = result.values  # noqa: PD011
+            if isinstance(solved, ValueStore):
+                solved._block_until_ready()  # noqa: SLF001
```
timmens, line 974: What about accounting for this readiness wait in a named phase? Since device work is asynchronous, computation launched during backward induction may only complete here. In that case, the phase records would report the wait as residual and understate `backward_induction`. Would it make sense to include this wait in that phase or add an explicit readiness phase?

src/_lcm/solution/period_capture.py
```diff
@@ -88,5 +321,109 @@ def capture_kernel_inputs(
...
+    period: int,
+    kernel_kwargs: dict[str, Any],
+    compiled_cores: Mapping[str, PlannedCore],
+) -> PeriodLayouts:
+    """Collect the layout descriptors of one regime-period."""
+    kernel = regime.solution.period_kernels[period]
+    kernel_type = type(kernel)
+    return PeriodLayouts(
+        route=f"{kernel_type.__module__}.{kernel_type.__qualname__}",
+        device_ids=tuple(int(device.id) for device in jax.devices()),
```
timmens, line 346: What about recording only the devices whose placements need substitution? A model configured to use two devices on a four-device host currently records all four visible devices, so replay with the same two solver devices is rejected. Restricting this inventory to the effective layout would let captures move between hosts with the same solver topology without requiring unrelated visible devices.

timmens (commented):
Two non-blocking discussion questions from the review.

src/lcm/model.py
hmgaudecker: Good point, done in 2bfec91f. The wait now runs inside its own `result_readiness` phase, so the phases of a public solve account for it by name.

In practice the unaccounted time was already small. Every log level that emits phase records also runs the per-period NaN/Inf check, and that check waits on each period's V before the next period is dispatched. On a production ACA warm solve, the named phases summed to 3590.566 s against 3590.607 s for the whole call, so about 0.04 s was left over. But that relied on the check happening to synchronise; with the phase in place, the phases account for the whole call by construction.

src/_lcm/solution/period_capture.py
hmgaudecker: Agreed, done in 2bfec91f. A capture now records only the devices its committed placements use, kept in `jax.devices()` order, so a model solved on two of four devices replays on any two devices.

Uncommitted leaves are left out. Replay leaves them where the backend puts them and never maps them to another device, so their default device isn't part of the layout. Without that exclusion, a model on devices (2, 3) recorded (0, 2, 3).

The new test solves on devices (2, 3) of four forced host devices, checks that the capture records exactly (2, 3), and replays on (0, 1). The existing layout tests now replay on as many devices as the capture recorded.

timmens (approved):

PR #457, opened 2026-09-22
FIX Make NB-EGM candidate values invariant to the cell tile width

### LINKED ISSUE #445: NB-EGM float32 cell tiling can switch a conditional policy at an envelope crossing
Changing NB-EGM's inner cell width can switch a conditional consumption policy in float32 near an envelope crossing, even when the corresponding value changes by one float32 spacing. This predates the execution-configuration migration in #440: the old `NBEGM(cell_block_size=...)` option at `ea95639bbac190f322f0cdef473208288a20f2ce` reproduces the identical result.

The CPU witness uses `tests.test_models.n_nbegm_toy.build_model`, the `n_nbegm` variant, two periods, three illiquid-state points from 0 to 20, discount factor 0.95, and the existing adaptive `_MESH` from `tests.simulation.test_nnbegm_split_workflow_parity`. Set only `max_refinement_rounds=16`; keep its `value_atol=value_rtol=1e-4` and `fail_closed=True`. Compare cell widths 3 and 1 with the outer-candidate width held at 2. The regression is preserved in #440's `tests/solution/test_nbegm_nested_axes.py`.

At the conditional adjuster bank coordinate `(65, 1, 10)`, the generated outer node is `11.071428298950195`, illiquid state is 10, and liquid grid index is 10:

| Output | Cell width 3 | Cell width 1 |
| --- | ---: | ---: |
| Consumption policy | 4.506383895874023 | 4.573979854583740 |
| Value | -4.809202671051025 | -4.809203147888184 |
| Marginal utility | 0.04924290254712105 | 0.04780387505888939 |

The consumption difference is about 0.0676. All 139 generated outer nodes and all finite masks remain exactly equal. Final solved values differ by at most `3.814697265625e-6` in this witness. The existing 64-ULP full-policy bound fails at float32 and passes at float64 on the tested CPU.

Controls separate this from an outer-loop ordering or migration defect:

- Repeating the same configuration in one process is bit-exact across solved values and all replay leaves.
- Changing only the outer-candidate width from 2 to 1 is also bit-exact.
- Changing only the inner cell width reproduces the policy switch.
- Matching-width pre-migration and migrated results are bit-exact across all those arrays.
- Calling the inner core at the fixed outer node reproduces the switch without adaptive host scheduling.
- Observing the actual stored candidate records through the existing barrier-bound production observer preserves the switch. The exact rational owner/readout certificates pass separately for each width: width 3 selects corner link 1160; width 1 selects interior link 44. The certified reduction selects correctly from each width's records, which differ slightly before reduction.

This evidence establishes sensitivity to candidate-record rounding at a crossing. It does not yet identify the upstream arithmetic transformation that perturbs those records or establish a production-model impact. In particular, it does not support changing the exact envelope comparator or relaxing a policy tolerance.

Keep the focused policy witness and its existing bound as a narrow known failure at float32, alongside passing exact node/mask and value checks. A follow-up should isolate the upstream width-dependent arithmetic and assess the selected policy and marginal channels, using the fixed-node witness and exact owner certificates to distinguish candidate construction from envelope selection. The #440 transport migration can be assessed against its matching-width legacy control independently.

hmgaudecker:
Closes #445.

## Problem

XLA's CPU backend contracts `utility + discount_factor * cont_value` into a fused multiply-add inside the vmapped NB-EGM cell program but not inside the `lax.map` batch-1 program. The same node therefore produced candidate values differing by one ULP depending on the cell tile width, so envelope ownership could flip at near-ties.

## Fix

`_discounted_continuation` wraps the discounted product in `jax.lax.optimization_barrier` at the two candidate-value sites in `nbegm_step.py`, which pins the rounding of the product before the addition regardless of batching.

New test `test_inner_core_records_are_cell_width_invariant_at_a_fixed_node` asserts bitwise-equal published outputs across cell widths; the xfail in `tests/solution/test_nbegm_nested_axes.py` is removed.

## Scope and caveat

Only the two sites that feed the envelope comparator are barriered. Sibling `discount_factor * cont_value` expressions elsewhere in the EGM family are left as they are: none has a demonstrated width-dependent failure, and barriering them uniformly would cost unmeasured compile time and fusion opportunities. The new test checks published outputs, not every pre-envelope record; the docstring states that limit.

## Verification

- fp32 nested-axes lane: 7 passed, 0 failed
- full CPU suite on Marvin (`-n 32`): 5261 passed, 0 failed, 7 skipped
- `prek run --all-files` clean

https://claude.ai/code/session_013LSspB3nyDJKhGrGZCvjCz


timmens (approved):
Reviewed the numerical rounding boundary and its regression coverage. The focused float32 witness passes, the full CI matrix is green, and I found no blocking or non-blocking issues.

PR #459, opened 2026-09-22
PERF Cell-tile frontier: ceiling, key gate, parameter-free program identity, padded maps

hmgaudecker:
Five commits on top of #456 (`perf/layout-faithful-capture`, fd2cf21b): the opt-in cell ceiling that closed the `pylcm-general-performance` Pro chain at round 1, plus four follow-ups now bundled for a new round-0 review.

## What changes

- **Opt-in ceiling on planned cell tile widths (8b1fe1ff):** `ExecutionConfig(axis_width_ceilings={"cell": N})` caps the planner's cell tile; no default changes.
- **Compatibility-key gate (5cb5a889):** 15 one-fact mutations of the lowering key (shape, dtype, axis order, device count, donation set, static option, function body, compiler option, x64, compatible parameter value, budget re-admission, fresh-process reproducibility, no model retention).
- **Binding collective ceiling tests (3897f89d):** the ceiling binds on a two-stakeholder model; widths, reservation and refusals move as specified. One strict `xfail` records a repeatable 1-ULP difference on collective GridSearch under the ceiling (FMA contraction of `u + beta * V'`, CPU backend), accepted as a documented caveat.
- **Program identity without solve-time parameter values (998b3130):** lowering keys derive from model structure, fixed parameters and grid support; the solution-artifact fingerprint still includes parameters (record version 7).
- **Padded mapped axes (480ae240):** `map_over_leading_axis` pads a mapped axis to a multiple of the batch width when the added rows are at most 1 %, so `jax.lax.map` no longer traces its body a second time for a remainder batch.

## Receipts for the padding

| measurement | base 8b1fe1ff | 480ae240 |
|---|---|---|
| CPU wave-0 frontend wall, 37 NB-EGM programs, reduced grid | 1948.6 s | 1148.4 s |
| GPU cold wave-0 `compilation_waves` wall, production grid, 3×A40 | 3125.2 s | 1679.8 s |
| GPU cold Σ compile, same | 7402.9 s | 4062.5 s |
| Full ACA GridSearch solve, 182 value arrays | — | 182/182 bitwise equal |
| Full ACA cold / warm solve | 4044.6 / 3629.2 s | 4066.5 / 3644.6 s |

Seed widths, evaluation decisions and the 41 refusals are identical on all 182 cores in the GPU pair. One run per arm, different nodes.

## Tests

`tests/solution/test_compilation_key_mutations.py`, `tests/solution/test_collective_cell_ceiling.py`, `tests/test_dispatchers.py`, `tests/solution/test_compilation_identity.py`: 87 passed, 1 xfailed (CPU, local).

Stacked on #456; independent of #457.

https://claude.ai/code/session_013LSspB3nyDJKhGrGZCvjCz


src/lcm/model.py
```diff
@@ -680,6 +681,14 @@ def __init__(
                 _simulation_programs(regimes=self._regimes),
             ),
         )
+        fail_if_axis_widths_name_undeclared_axes(
+            axis_widths=execution_config.axis_width_ceilings,
```
timmens: This validation accepts ceilings for axes declared by either phase, so `axis_width_ceilings={"subject": 1}` passes model construction. Simulation planning never forwards `execution.axis_width_ceilings` to `workspace_width_candidates` or `plan_workspace`, causing the ceiling to be silently ignored. For a light unbudgeted simulation with at least 262,144 subjects, the runtime can select a subject width of 262,144 despite the declared ceiling of 1. Please thread the ceilings through simulation candidate enumeration and workspace planning, or restrict the field and validation to solve axes.

timmens (changes requested):
One blocking issue: axis-width ceilings are accepted for simulation axes but are not applied during simulation planning.

src/lcm/model.py
hmgaudecker: Agreed, the ceiling never reached simulation planning. Fixed in 059e5435 by threading `axis_width_ceilings` into all three simulate planners:

- chunk-profile candidate enumeration (`_ChunkProfiler.__call__`);
- the independent anchor widths (`_independent_anchor_widths`);
- dispatch planning (`SimulationRuntime._prepare_materialized`). This covers your unbudgeted case: the derived subject tile is a fixed width, and `plan_workspace` narrows fixed widths to the ceiling.

New tests in `tests/solution/test_cell_frontier_ceiling.py`, each run with no budget and with a budget:
- every simulate dispatch tiles `subject` at the ceiling;
- a moving control: without the ceiling, the same simulation tiles wider;
- the simulated panel is exactly equal with and without the ceiling.

